PR #1, opened 2019-11-18
Initial scaffolding

losvedir:
Asana: [basic code scaffolding](https://app.asana.com/0/584764604969369/1145845426232957).

Inspired by the Skate repo. They are in the process of removing babel (open PR), so I just went ahead and did that right from the get-go.

One thing I didn't set up is frontend tests. There is no code to test, but I could create a dummy module and a dummy test, just to establish the framework. Skate uses react and has a lot of tests tailored to that, but I'm not sure if we'll be using react or not, so I didn't want to go too far down that direction. But I could create a simple code-only jest framework like they have.


losvedir:
It looks like Skate's "remove babel" PR is more speculative than I realized, and they're still not certain it will work with full browser support, etc. However, since this app is just internal to us for now, I think it's probably still a good thing to try using exclusively TypeScript and just the TS compiler (which can target ES3, ES5, and more) and seeing how it goes. We can always add babel later if we need to.

.tool-versions
```diff
@@ -0,0 +1,3 @@
+elixir v1.9.1-otp-22
```
lemald: It was Sentry that didn't work with Elixir 1.9 yet, correct? Are we okay with not being able to use it yet in Arrow?

assets/tsconfig.json
```diff
@@ -0,0 +1,62 @@
...
+
+    /* Strict Type-Checking Options */
+    "strict": true,                        /* Enable all strict type-checking options. */
+    "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
+    // "strictNullChecks": true,              /* Enable strict null checks. */
+    // "strictFunctionTypes": true,           /* Enable strict checking of function types. */
+    // "strictBindCallApply": true,           /* Enable strict 'bind', 'call', and 'apply' methods on functions. */
+    // "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
+    "noImplicitThis": true,                /* Raise error on 'this' expressions with an implied 'any' type. */
+    // "alwaysStrict": true,                  /* Parse in strict mode and emit "use strict" for each source file. */
```
lemald: I don't know much about TypeScript, but are we sure we want to start out with a lot of these strictness options disabled? It seems like the sort of thing where it's easy to go from strict to not strict if we decide later, but harder to go in the other direction. Plus my gut instinct is that stricter typing is always better, but you already know that. 😃

losvedir: Ha, well you know I'm a big fan of stricter types, too! I just pulled this over verbatim from Skate, for some consistency, but that's a good point about it being easier to go the other direction. I'll enable all the strict checks for now and we can disable as necessary.

README.md
```diff
@@ -9,3 +9,11 @@ To start your Phoenix server:
   * Start Phoenix endpoint with `mix phx.server`
 
 Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
+
+Useful when developing:
+
+  * Test elixir code with `mix test`
```
paulswartz: Lint with `mix credo` and `mix dialyzer`

losvedir: Ah, you're right! Forgot to set those up, too.

PR #2, opened 2019-11-25
feat: basic CI setup

handorff:
- elixir: mix format, test, credo, dialyzer
- typescript: npm test and npm run check
- install versions of erlang, elixir, node matching `.tool-versions`
- also install latest npm along with the new node version; the preinstalled one doesn't have `npm ci`

### DIFF
--- a/semaphore/credo.sh
+++ b/semaphore/credo.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -ex
+
+mix do deps.get, credo --strict
--- a/semaphore/credo.sh
+++ b/semaphore/credo.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -ex
 
-mix do deps.get, credo --strict
+mix credo --strict
--- a/semaphore/dialyzer.sh
+++ b/semaphore/dialyzer.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+export MIX_ENV=dev
+mix compile --force
+find $SEMAPHORE_CACHE_DIR -name "dialyxir_*_deps-$MIX_ENV.plt*" | xargs -I{} cp '{}' _build/$MIX_ENV
+export ERL_CRASH_DUMP=/dev/null
+mix dialyzer --plt
+cp _build/$MIX_ENV/*_deps-$MIX_ENV.plt* $SEMAPHORE_CACHE_DIR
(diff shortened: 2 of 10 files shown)


semaphore/credo.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -ex
+
+mix do deps.get, credo --strict
```
losvedir: `setup` should have gotten the deps already, right?

semaphore/setup.sh
```diff
@@ -0,0 +1,30 @@
...
+if ! kiex use $ELIXIR_VERSION; then
+    kiex install $ELIXIR_VERSION
+    kiex use $ELIXIR_VERSION
+fi
+
+mix local.hex --force
+mix local.rebar --force
+mix deps.get
+
+pushd assets && npm ci && popd
```
losvedir: `npm` has a flag `--prefix` that lets you run commands as if it were in a subdirectory, without actually going there: `npm --prefix assets ci`.

I think it's a little better because if the `npm` command fails for some reason, then the script is left in the higher directory rather than failing to `popd`.

semaphore/test.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+
+mix coveralls.json &&
+pushd assets && npm test && popd &&
+bash <(curl -s https://codecov.io/bash)
```
losvedir: Have you been able to confirm this working? Like, are the codecov status checks appearing in the GitHub PR? I checked now, but it's still waiting for semaphore to finish.

For light-rail-speeds recently I had to do two things:

1. Add a `codecov.yml` (`.codecov.yml` is also supported, and I think we're inconsistent as an organization which we use: API even has both!)
2. Enable the GitHub repo to have a codecov webhook.

semaphore/dialyzer.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+export MIX_ENV=dev
+mix compile --force
```
lemald: Do we also want to use the `--warnings-as-errors` flag?

semaphore/test.sh
lemald: All of the other repos I'm familiar with specify a token with the `-t` argument here, do we need to do that for this one as well?

PR #3, opened 2019-12-05
Deploy to AWS

losvedir:
Asana: [basic dev infrastructure & deployment setup](https://app.asana.com/0/584764604969369/1145845426232959).

Should be reviewable commit-by-commit. The commits are:

* Move the `secret_key_base` into an ENV var in prod. In many of our other apps this is in Semaphore, since that's where the release is built. However, upon looking into it, this can actually be configured at runtime, by adding an `init` callback to the `Arrow.Endpoint` module, and therefore live in AWS. Unfortunately, for now, it's an EB environment variable, but it would be good to get it into the Secrets Manager. Still, I figured it's a step in the right direction compared to having it in Semaphore.

* Similarly, update the DB config to accept an ENV var at runtime. This is how our other apps work, and I found it useful to follow to ensure I had the roles and security groups set up. However, the approach is superseded by a later commit.

* Add the Dockerfile and script to build a release and upload the docker image to our ECR repository.

* Add a script to deploy the app to EB. It works by using a `Dockerrun.aws.json` file which specifies the image tag, which EB then pulls.

* Switch to using IAM authentication for the RDS instance.


config/prod.exs
```diff
@@ -10,12 +10,18 @@ use Mix.Config
 # which you should run after static files are built and
 # before starting your production server.
 config :arrow, ArrowWeb.Endpoint,
+  http: [:inet6, port: String.to_integer(System.get_env("PORT") || "4000")],
```
paulswartz: I thought configuration in `config/` is set at build time, not run time?

lib/arrow/repo.ex
```diff
@@ -2,4 +2,32 @@ defmodule Arrow.Repo do
   use Ecto.Repo,
     otp_app: :arrow,
     adapter: Ecto.Adapters.Postgres
+
+  def init(_context, config) do
```
paulswartz: Do we need to call `super` at all?

semaphore/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+
+githash=$(git rev-parse --short HEAD)
+gitmsg=$(git log -1 --pretty=%s)
+
+# Create Dockerrun file pointing to ECR image, and zip it up
+mkdir $githash
+sed "s/GITHASH/$githash/g" ./semaphore/Dockerrun.aws.json > $githash/Dockerrun.aws.json
+pushd $githash
+zip -r ../$githash.zip .
+popd
```
paulswartz: can we also delete `rm -r $githash` after zipping?

config/prod.exs
losvedir: Oh, true. I just brought that over from `prod.secret.exs` when I deleted that file, without looking at it too much. I'll just hardcode the value since you can't just change the ENV var without _also_ changing the Dockerfile and Dockerrun.aws.json files anyway.

Looks like further down in the file there's the same `System.get_env` issue for the Ecto `pool_size` option. Have we ever changed that in our apps? I'd be inclined to just set it to `10` directly, same as this config item, but it's also possible to set this one in the `Repo.init/2` callback.

paulswartz: Yeah, I agree w/ the hardcoding: I don't think we've ever overridden it in a Docker app.
For pool_size, that seems like something we might want to change from the configuration, so `init/2` is probably good for that.

lib/arrow/repo.ex
losvedir: Nope, looks like it's not `defoverridable`. I tried, but it crashed.

semaphore/deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+
+rm -r $githash
+
+# Put zip on S3 and use to create application version
+aws s3 cp $githash.zip s3://$S3_BUCKET_NAME/$APP/$githash.zip
```
paulswartz: Should also `rm $githash.zip` afterwards.

PR #4, opened 2019-12-20
feat: AWS Cognito integration

lemald:
Ticket: [AWS Cognito integration](https://app.asana.com/0/584764604969369/1145845426232983/f)

Should be reviewable commit-by-commit in a clean way. A lot of this is carried over fairly directly from Signs UI, so it may be instructive to look there as well.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -21,6 +21,11 @@ config :arrow, :redirect_http?, true
 
 config :arrow, ArrowWeb.AuthManager, issuer: "arrow"
 
+config :ueberauth, Ueberauth,
+  providers: [
+    cognito: {Arrow.Ueberauth.Strategy.Fake, []}
+  ]
+
 # Configures Elixir's Logger
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
--- a/config/config.exs
+++ b/config/config.exs
@@ -19,6 +19,8 @@ config :arrow, ArrowWeb.Endpoint,
 
 config :arrow, :redirect_http?, true
 
+config :arrow, ArrowWeb.AuthManager, issuer: "arrow"
+
 # Configures Elixir's Logger
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
--- a/config/config.exs
+++ b/config/config.exs
@@ -25,8 +25,15 @@ config :arrow, ArrowWeb.AuthManager, issuer: "arrow"
 
 config :ueberauth, Ueberauth,
   providers: [
-    cognito: {Arrow.Ueberauth.Strategy.Fake, []}
+    cognito: {Ueberauth.Strategy.Cognito, []}
   ]
+
+config :ueberauth, Ueberauth.Strategy.Cognito,
+  auth_domain: {System, :get_env, ["COGNITO_DOMAIN"]},
+  client_id: {System, :get_env, ["COGNITO_CLIENT_ID"]},
+  client_secret: {System, :get_env, ["COGNITO_CLIENT_SECRET"]},
+  user_pool_id: {System, :get_env, ["COGNITO_USER_POOL_ID"]},
+  aws_region: {System, :get_env, ["COGNITO_AWS_REGION"]}
 
 # Configures Elixir's Logger
 config :logger, :console,
--- a/config/config.exs
+++ b/config/config.exs
@@ -17,7 +17,9 @@ config :arrow, ArrowWeb.Endpoint,
   render_errors: [view: ArrowWeb.ErrorView, accepts: ~w(html json)],
   pubsub: [name: Arrow.PubSub, adapter: Phoenix.PubSub.PG2]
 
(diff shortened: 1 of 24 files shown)


config/config.exs
```diff
@@ -17,6 +17,10 @@ config :arrow, ArrowWeb.Endpoint,
   render_errors: [view: ArrowWeb.ErrorView, accepts: ~w(html json)],
   pubsub: [name: Arrow.PubSub, adapter: Phoenix.PubSub.PG2]
 
+config :arrow, ArrowWeb.AuthManager,
+  issuer: "arrow",
+  secret_key: "test key"
```
losvedir: In general I'm a little leery of providing insecure defaults and then overriding them in prod, because it seems possible to accidentally change the prod somehow and inadvertently fall back to the default. Can we remove the `:secret_key` from the global `config.exs` and put it in `dev.exs` and `test.exs` explicitly?

lib/arrow_web/router.ex
```diff
@@ -13,8 +13,16 @@ defmodule ArrowWeb.Router do
...
+    plug(ArrowWeb.AuthManager.Pipeline)
+  end
+
+  pipeline :ensure_auth do
+    plug(Guardian.Plug.EnsureAuthenticated)
+  end
+
   scope "/", ArrowWeb do
-    pipe_through :browser
+    pipe_through [:browser, :auth, :ensure_auth]
```
losvedir: I just realized that this will actually interfere with the health check requests, so this PR is actually blocked by those other two tickets.

lemald: Oh, good point. Well I can pick that one up in a moment.

config/config.exs
lemald: Fixed.

losvedir: I realize this is what signs ui did. We're just iterating across our apps to better and better practices. :)

lib/arrow_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,29 @@
...
+      ttl: {expiration - current_time, :seconds}
+    )
+    |> redirect(to: ArrowWeb.Router.Helpers.page_path(conn, :index))
+  end
+
+  def callback(
+        %{assigns: %{ueberauth_failure: %Ueberauth.Failure{}}} = conn,
+        _params
+      ) do
+    send_resp(conn, 403, "unauthenticated")
```
losvedir: Technically, a 403 would be "unauthorized". A 401 is "unauthenticated".

I'm trying to remember how this failure callback works. Do we fall into it both if the username/password flow fails (401), _and_ if the group is not the admin group (403)?

lemald: If the group isn't the admin group it goes to a different controller, so yeah, this should be a 401 here I guess.

lib/arrow_web/ensure_arrow_group.ex
```diff
@@ -0,0 +1,22 @@
+defmodule ArrowWeb.EnsureArrowGroup do
+  import Plug.Conn
+
+  @arrow_group "arrow-admin"
```
losvedir: I see this literal a few times in the PR. What if we moved it to `config.exs` so that all the places in here can access it via `Application.get_env/2`?

config/prod.exs
```diff
@@ -17,6 +17,18 @@ config :arrow, ArrowWeb.Endpoint,
 
 config :arrow, ArrowWeb.AuthManager, secret_key: {System, :get_env, ["ARROW_AUTH_SECRET"]}
 
+config :ueberauth, Ueberauth,
```
losvedir: In the secure-by-default vein, can you also move the fake strategy out of `config.exs` into `test.exs` and `dev.exs` separately?

lib/arrow_web/controllers/auth_controller.ex
lemald: This should be fixed.

config/prod.exs
lemald: Done

lib/arrow_web/ensure_arrow_group.ex
lemald: Done

config/config.exs
```diff
@@ -17,7 +17,23 @@ config :arrow, ArrowWeb.Endpoint,
   render_errors: [view: ArrowWeb.ErrorView, accepts: ~w(html json)],
   pubsub: [name: Arrow.PubSub, adapter: Phoenix.PubSub.PG2]
 
-config :arrow, :redirect_http?, true
+config :arrow,
+  redirect_http?: true,
+  arrow_group: "arrow-admin"
```
losvedir: Nitpick: Maybe `cognito_group` or something to that effect is a little clearer?

lemald: Yeah that's probably better, just updated it.

PR #6, opened 2019-12-20
feat: redirect http to https

lemald:
Ticket: [🏹 Redirect prod requests to HTTPS](https://app.asana.com/0/584764604969369/1154773501302867/f)

This is basically the same way Signs UI does it, but I went with the default-to-secure approach here and explicitly overrode it to _not_ redirect to HTTPs in dev and test. Apparently Codecov thinks this is fine and the equivalent code in Signs UI has no tests I can find, but it would be nice to test it somehow, I'm just not sure how...

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -17,6 +17,8 @@ config :arrow, ArrowWeb.Endpoint,
   render_errors: [view: ArrowWeb.ErrorView, accepts: ~w(html json)],
   pubsub: [name: Arrow.PubSub, adapter: Phoenix.PubSub.PG2]
 
+config :arrow, :redirect_http?, true
+
 # Configures Elixir's Logger
 config :logger, :console,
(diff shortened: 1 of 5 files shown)


losvedir:
Testing would be good.... but yeah, I'm not entirely sure how either. I _think_ router pipelines are evaluated at build time for speed, so I don't believe you can override `:redirect_http?` just for a single test. (Might be worth checking, though. If you can, then you should be able to set it to `true` for the test, do a controller test, and assert a 301 response, and then try again with `x_forwarded_proto` header set and assert 200.)

I guess one idea might be to actually leave it set to `true` and update the `ConnCase` helper to always include the header on tests, akin to how you build the authenticated conn in that other PR. And then have one test where you delete that header and assert a redirect.

lemald:
@losvedir okay well as expected the first, simpler approach didn't work, but the second one did!

PR #14, opened 2019-12-23
feat: initial model for Adjustment & Disruption

paulswartz:


priv/repo/migrations/20191223182231_create_disruption_adjustments.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Arrow.Repo.Migrations.CreateDisruptionAdjustments do
+  use Ecto.Migration
+
+  def change do
+    create table(:disruption_adjustments) do
+      add :disruption_id, references(:disruptions, on_delete: :nothing)
```
losvedir: Hmm, I think we might want `on_delete: :delete_all` for the `disruption_id` since it seems plausible to me that we might have some UI ability to just delete a disruption and we'd want to delete all the join records here with it.

I can see `on_delete: :nothing` for the `adjustment_id`, though, because it probably is correct to raise an exception if we try to delete an adjustment in use.

priv/repo/migrations/20191223181443_create_adjustments.exs
```diff
@@ -0,0 +1,15 @@
...
+  def change do
+    create table(:adjustments) do
+      add :source, :string
+      add :source_label, :string
+      add :route_id, :string
+
+      timestamps(type: :timestamptz)
+    end
+
+    create unique_index(:adjustments, [:source, :source_label], name: :source_source_label)
```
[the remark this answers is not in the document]
losvedir: For that matter, do we not want it unique on just the `source_label`? I guess we haven't worked out exactly how Arrow-sourced adjustments are going to work. I was sort of envisioning if we create one from Arrow then the source is `arrow`. But when the app starts up it reads the `shuttles.json` file and updates the source from `arrow` -> `gtfs_creator` for any records we've since added it to gtfs_creator.

If we allow the same `source_label` for both sources, I think that means we have to do a little more juggling. When we add the `source_label` to gtfs_creator we'll have to have Arrow go in and update the foreign keys to the new adjustment. And then maybe delete the now-orphaned arrow-sourced one?

priv/repo/seeds.exs
```diff
@@ -9,3 +9,24 @@
...
+    %{
+      inserted_at: now,
+      updated_at: now,
+      source: "gtfs_creator",
+      source_label: j["id"],
+      route_id: j["attributes"]["route_id"]
+    }
+  end)
+
+Repo.insert_all(Adjustment, adjustments)
```
losvedir: I think at grooming we had talked about maybe populating with some sample disruptions as well? So the home page can look like the mocks more or less.

priv/repo/shuttles.json
```diff
@@ -0,0 +1 @@
...es": {"route_id": "CR-Franklin"}}, {"id": "ForgeParkSouthStation", "attributes": {"route_id": "CR-Franklin"}}, {"id": "GreenBGovernmentCenterNorthStation", "attributes": {"route_id": "Green-B"}}, {"id": "GreenCGovernmentCenterNorthStation", "attributes": {"route_id": "Green-C"}}, {"id": "GreenDGovernmentCenterNorthStation", "attributes": {"route_id": "Green-D"}}, {"id": "GreenEGovernmentCenterNorthStation", "attributes": {"route_id": "Green-E"}}, {"id": "SL1NoTransitway", "attributes": {"route_id": "741"}}, {"id": "SL2NoTransitwayAM", "attributes": {"route_id": "742"}}, {"id": "SL2NoTransitwayPM", "attributes": {"route_id": "742"}}, {"id": "SL2NoTransitwayPMNoBlackFalcon", "attributes": {"route_id": "742"}}, {"id": "SL3NoTransitway", "attributes": {"route_id": "743"}}, {"id": "SLWNoTransitway", "attributes": {"route_id": "746"}}, {"id": "HaymarketSullivanTufts", "attributes": {"route_id": "Orange"}}, {"id": "RugglesSullivan", "attributes": {"route_id": "Orange"}}, {"id": "MattapanBusReplacement", "attributes": {"route_id": "Mattapan"}}, {"id": "LowellBusReplacement", "attributes": {"route_id": "CR-Lowell"}}, {"id": "FranklinViaFairmount", "attributes": {"route_id": "CR-Franklin"}}]
```
losvedir: Maybe we should call this `adjustments.json`?

priv/repo/seeds.exs
paulswartz: That's right, but I was holding off on that until the Disruption model is actually detailed enough to support the samples we had in the mockup.

priv/repo/shuttles.json
paulswartz: I thought about that, but it's the shuttle data from gtfs_creator. The seeds.exs then turns them into Adjustments.

PR #23, opened 2019-12-27
feat: run migrations at startup (prod only)

paulswartz:
We run a transient GenServer to handle the migrations, and then stop when the
migrations complete. If there's an error, the supervision tree will restart
the migration task.

### DIFF
--- a/lib/arrow/repo/migrator.ex
+++ b/lib/arrow/repo/migrator.ex
@@ -1,0 +1,43 @@
+defmodule Arrow.Repo.Migrator do
+  @moduledoc """
+  GenServer which runs on startup to run Ecto migrations, then terminates.
+  """
+  use GenServer, restart: :transient
+  require Logger
+
+  @opts [module: Ecto.Migrator]
+
+  def start_link(opts) do
+    opts = Keyword.merge(@opts, opts)
+    GenServer.start_link(__MODULE__, opts)
+  end
+
+  @impl GenServer
+  def init(opts) do
+    {:ok, opts, {:continue, :migrate}}
+  end
+
+  @impl GenServer
+  def handle_continue(:migrate, opts) do
+    migrate!(opts[:module])
+    {:stop, :normal, opts}
+  end
+
+  defp migrate!(module) do
+    for repo <- repos() do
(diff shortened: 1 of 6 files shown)


lib/arrow/repo/migrator.ex
```diff
@@ -0,0 +1,38 @@
...
+  @opts [module: Ecto.Migrator]
+
+  def start_link(opts) do
+    opts = Keyword.merge(@opts, opts)
+    GenServer.start_link(__MODULE__, opts)
+  end
+
+  @impl GenServer
+  def init(opts) do
+    {:ok, opts, {:continue, :migrate}}
```
losvedir: Wow! TIL about `:continue`.

lib/arrow/repo/migrator.ex
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  @impl GenServer
+  def handle_continue(:migrate, opts) do
+    migrate!(opts[:module])
```
losvedir: Might be useful to log when the migrations successfully complete, too? Nice way to easily verify that the migrations ran in prod, and also how long they took.

lib/arrow/repo/migrator.ex
paulswartz: :+1: it should address the 5s startup issue.

lib/arrow/repo/migrator.ex
paulswartz: Can do!

paulswartz: done in latest commit

paulswartz:
Test deploy looks good:
```
17:29:03.844 [info] Migration finished repo=Elixir.Arrow.Repo time=1000
17:29:03.818 [info] == Running 20191223182231 Arrow.Repo.Migrations.CreateDisruptionAdjustments.change/0 forward
17:29:03.771 [info] == Running 20191223181443 Arrow.Repo.Migrations.CreateAdjustments.change/0 forward
17:29:03.736 [info] == Running 20191223181419 Arrow.Repo.Migrations.CreateDisruptions.change/0 forward
17:29:02.843 [info] Migrating repo=Elixir.Arrow.Repo
```

PR #25, opened 2019-12-27
feat(Disruption): date exceptions and trip short names

paulswartz:
These are easier than the recurrences, so starting with them.

### DIFF
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -9,10 +9,14 @@ defmodule Arrow.Disruption do
   use Ecto.Schema
   import Ecto.Changeset
 
+  alias Arrow.Disruption.{Exception, TripShortName}
+
   schema "disruptions" do
     field :end_date, :date
     field :start_date, :date
 
+    has_many :exceptions, Exception
+    has_many :trip_short_names, TripShortName
     many_to_many :adjustments, Arrow.Adjustment, join_through: "disruption_adjustments"
 
     timestamps(type: :utc_datetime)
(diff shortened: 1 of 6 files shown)


priv/repo/migrations/20191223181711_create_disruption_exceptions.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Arrow.Repo.Migrations.CreateDisruptionExceptions do
+  use Ecto.Migration
+
+  def change do
+    create table(:disruption_exceptions) do
+      add :excluded_date, :date
```
losvedir: Maybe `null: false` here? The Ecto validation implies it's required, and I can't think of a reason we'd want a disruption in the DB without a date.

priv/repo/migrations/20191223181837_create_disruption_trip_short_names.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Arrow.Repo.Migrations.CreateDisruptionTripShortNames do
+  use Ecto.Migration
+
+  def change do
+    create table(:disruption_trip_short_names) do
+      add :trip_short_name, :string
```
losvedir: I _think_ this could also be `null: false`, but I'm less certain of how exactly we use trip short names, so I could be wrong about that.

priv/repo/migrations/20191223181711_create_disruption_exceptions.exs
paulswartz: yep, fixed!

priv/repo/migrations/20191223181837_create_disruption_trip_short_names.exs
paulswartz: yep, fixed!

lib/arrow/disruption.ex
```diff
@@ -9,20 +9,34 @@ defmodule Arrow.Disruption do
...
+    trip_short_names =
+      for name <- attrs[:trip_short_names] || [],
+          do: TripShortName.changeset(%TripShortName{}, %{trip_short_name: name})
+
     disruption
     |> cast(attrs, [:start_date, :end_date])
     |> validate_required([:start_date, :end_date])
-    |> put_assoc(:adjustments, attrs.adjustments)
+    |> put_assoc(:adjustments, attrs[:adjustments] || [])
+    |> put_assoc(:exceptions, exceptions)
```
losvedir: I'm not super familiar with Ecto associations, but the [docs](https://hexdocs.pm/ecto/Ecto.Changeset.html#put_assoc/4-example-adding-tags-to-a-post) seem to imply that `put_assoc` is best used when working with _all_ of the children of a record. From the UI, at least, I can see us adding just a single Exception at a time, so it might not be the best approach here. (It feels more akin to the linked "adding a comment to a post" example.) That said, I'm happy to leave this as-is for now, and we'll just change how we work with the associations in the app later, if need be.

paulswartz: Yeah, I'm definitely not confident that this is the best long-term (or even short/medium term) way that we want to be handling this. Maybe we'll want something like `Disruption.add_exception/2` which adds a single one, but doesn't modify the disruption otherwise? Seems like something we'll address as we start adding the interactions with the UI.

PR #28, opened 2019-12-30
Days of week / recurrences

paulswartz:


priv/repo/seeds.exs
```diff
@@ -30,3 +31,88 @@ adjustments =
...
+      },
+      %{
+        start_date: ~D[2019-12-11],
+        end_date: ~D[2020-03-29],
+        adjustments: [Repo.get_by!(Adjustment, source_label: "ForgeParkSouthStation")],
+        excluded_dates: [~D[2019-12-28], ~D[2019-12-29], ~D[2020-01-04], ~D[2020-01-05]],
+        trip_short_names: ["1716", "1718", "1719", "2716", "2718", "2719"],
+        days_of_week: [
+          %{saturday: true, sunday: true}
+        ]
```
paulswartz: This is one example where the current model isn't ideal.
- the start/end/excluded dates/days_of_week are all the same
- but the different adjustments apply to different `trip_short_names`, so we need two Disruptions.

lib/arrow/disruption.ex
```diff
@@ -9,12 +9,13 @@ defmodule Arrow.Disruption do
...
   import Ecto.Changeset
 
-  alias Arrow.Disruption.{Exception, TripShortName}
+  alias Arrow.Disruption.{DayOfWeek, Exception, TripShortName}
 
   schema "disruptions" do
     field :end_date, :date
     field :start_date, :date
 
+    has_many :days_of_week, DayOfWeek
```
losvedir: I think we had been talking about having just one recurrence per disruption. Should this be a `has_one`, with a unique constraint on the `disruption_id` of the other table?

OTOH, we might have ended up at that idea because we were at one point talking about just having a "recurrence string" field on the disruption. This data model naturally supports multiple recurrences per disruption, so maybe we _should_ allow that. We'd have to update our UI mockups a little bit, though, if we want to support adding that now.

paulswartz: This lines up with the [Figma mockups](https://www.figma.com/file/X3WXdYxUkWE9UjtOWGdF6v/arrow?node-id=0%3A1), in particular "create-1" where Friday and Saturday/Sunday have different start times.

losvedir: For that one I believe the plan was a special case in the UI when Friday, Saturday, and Sunday were selected, and that the underlying recurrence would be just Friday at that start time, with a duration of, say, 52 hours in seconds. Note that the Sat/Sun select boxes are disabled because when Fri/Sat/Sun is selected, start of service to end of service on Sat/Sun is implied.

Which reminds me: if we are going that way we need a duration field on this model.

All that said, as I type it out now it sounds a bit convoluted and I think was a response to trying to work in tandem with a recurrence string. But if this model more naturally supports that UI by having two `disruption_days_of_weeks`, one for Friday, and one for Sat/Sun, I like that.

I think the only reservation I have is if we think we'll ever need a more complex recurrence that is only supportable by having the embedded recurrence string (e.g. "first Sunday of the month" or something like that").

paulswartz: I agree that this is a bit of a change from what we'd discussed earlier. I'm happy to leave this open/blocked and wait to discuss it when the team is back in the office.

losvedir: Well, I just wanted to flag it as different is all. Now that I see it, I actually do prefer your approach here. I'll approve the PR, but yeah maybe we should run it by Paul H, at least, who I think proposed the recurrence + duration idea.

handorff: I'm OK with what's here. I agree that we wouldn't use more complex recurrence rules.

One advantage of using an ical recurrence rule instead would be so that we can use existing libraries which have support for that format. But since we're using such a restricted subset of recurrence rules, I'm could be willing to believe that the extra work we'd have to do in order to encode user submissions as ical rules would outweigh the benefit of being able to use existing ical support.

A couple of examples in case this changes anyone's mind:
- https://hexdocs.pm/cocktail/Cocktail.Schedule.html can give a list of occurrences given a rule, as well as making a text description. However, we'd still have to add our own support for excluding certain dates.
- https://fullcalendar.io/docs/rrule-plugin can display calendar events given an ical string and duration.

priv/repo/seeds.exs
handorff: If we want to, we can associate trip short names with a (disruption_id, adjustment_id), in which case we'd only need one disruption.

lib/arrow/disruption.ex
paulswartz: It should be possible to create rrules from this? start_time+end_time give you the duration, and the days give you a recurrence.

PR #30, opened 2020-01-06
feat: page header (with React)

handorff:
- Set up react, react-dom
- Set up bootstrap
- Basic setup for multiple pages / create an edit page
- Create header


assets/css/app.scss
```diff
@@ -1,3 +1,21 @@
...
+  font-style: normal;
+  font-weight: bold;
+  font-size: 25px;
+  line-height: 31px;
+  display: flex;
+  align-items: center;
+}
+
+.header-arrow-text {
+  padding-left: 8px;
```
paulswartz: Can we use `rem` instead of pixels?

assets/src/edit.tsx
```diff
@@ -5,15 +5,32 @@ declare function require(name: string): string
 // tslint:disable-next-line
 require("../css/app.scss")
 
+import "bootstrap/dist/css/bootstrap.min.css"
```
paulswartz: Should this go into `app.scss` instead, rather that putting it in each file?

losvedir:
Is the idea then that the whole page will be one react app? I think I would have thought the header, at least, would be part of the HTML template, unless it responds in some way to what's going on in the page.

Beyond that, I think the implication of going all react is we're not going to be using a lot of the Phoenix form helpers, changesets, etc. Will the react app post via JSON so the backend is essentially just an API (I think this is how Skate works), or will it manipulate the form on the page and submit to Phoenix HTML controllers?

handorff:
@losvedir: we were imagining that at least home and create/edit would be different pages (so it's not just a full single-page app). The create-* and detail-* layouts in Figma would be a part of the edit page which we're currently working on.

"The backend is essentially just an API" does match what we had in mind, but speaking for myself, having never used Phoenix I don't have a sense of what we'd be giving up by doing that. I'm certainly not opposed to switching to using react more on an element-by-element basis if that makes more sense for what we're trying to build.

losvedir:
Okay, I like the home page and the create/edit page being different. That said, how do you plan to do it with this structure? With `<App>` being the top of the react tree, which is parent to both of those pages, what will those controllers actually render on the Phoenix side? Some sort of JSON configuration combined with a router of some sort at the `App` level?

Perhaps the header and frame should just be part of the phoenix app template, and what each of the "pages" renders is the actual entry point to its own react tree. So you'd have a `<HomeList>` component that the `PageController.home/1` renders, and a `<Disruption>` component that the `PageController.disruption/1` renders. (Here, since you were talking about the backend being a JSON API, I'm envisioning the controllers that load the JS just being simple `PageController` templates you might use for a static page.

Changesets and Phoenix form helpers are pretty neat, especially when dealing with validations and errors. Right now we have our model configured with the validations (uniqueness, presence, etc) and if you try to create a record against a validation then it will fail. If you then use that failing changeset in a form helper it will prefill your previously entered values, and also highlight the failed ones.

All that said, this sort of helper has often been fiddly for me (I have more experience in Rails, but have also built some forms in Phoenix that have been similar), and don't seem to play well when you want to introduce fancy JS interactions. So I'm okay with foregoing it altogether and doing the JS frontend / JSON API backend thing.

losvedir:
Actually, thinking about it more, I don't think I like the `PageController.home/1` and `PageController.disruption/1` approach. Probably better to do it the "REST" way, with `DisruptionController.index/1` and `DisruptionController.[new|edit]/1`, etc. But still, those pages would render the react tree in the same way I was describing.

lemald:
@losvedir for what it's worth, I was looking at [this StackOverflow question](https://stackoverflow.com/questions/31933359/using-react-in-a-multi-page-app) and this approach sounded less hack-y to me than having one file and then basically determining which page to render in the react. In any event it looks like it shouldn't be hard to switch between approaches if we want to.

losvedir:
Yeah, that's the approach I've used before, for the reasons mentioned in that post. I don't find it so bad, but since this is just internal and we have fine bandwidth I won't make a fuss about it.

losvedir:
It's also useful if you want to render _two_ independent react component trees on the page, but I guess that doesn't apply to us, since the whole page is a react app.

PR #50, opened 2020-01-24
feat: Start of Create / Edit View

lemald:
Asana ticket: [🏹 create/preview view](https://app.asana.com/0/584764604969369/1148734524638917/f)

Not sure what to say about this one beyond what's in the ticket, but you probably don't want to review this one on a commit-by-commit basis. Also we should definitely squash-and-merge on this one.


assets/src/disruptions/newDisruption.tsx
```diff
@@ -0,0 +1,296 @@
...
+                id={"adjustment-delete-" + i}
+                onClick={() => deleteAdjustment(i)}
+              >
+                delete
+              </a>
+            </Form.Row>
+          )
+        })}
+        {!isAddingAdjustment ? (
+          <a
```
paulswartz: Since this isn't actually a link, can we make it a `<button>` instead?

assets/tests/setup.ts
```diff
@@ -1 +1,3 @@
-// configure enzyme here
+import { configure } from "enzyme"
+import Adapter from "enzyme-adapter-react-16"
+configure({ adapter: new Adapter() })
```
paulswartz: Do we need to configure coverage for this? I don't see it getting picked up by Codecov.

assets/src/disruptions/newDisruption.tsx
losvedir: Is the idea that any link with `href="#"` should be a button? I can change this one, but we have a few others, too.

paulswartz: Yeah; it's not really a link. If you aren't linking to something else, you're using `<a>` for styling, which can also be accomplished with CSS.

losvedir: Updated all the faux-links to buttons.

PR #62, opened 2020-01-31
Load adjustments on startup

lemald:
Asana ticket: [🏹 Arrow loads adjustments on startup](https://app.asana.com/0/584764604969369/1157557244099559/f)

Two commits should be reviewable independently. The fact that the app now loads stuff into the database on startup makes testing a little weird. I've tried to make it all make sense, but some things like testing that the database starts out with no adjustments just don't really apply anymore.

### DIFF
--- a/lib/arrow/adjustment.ex
+++ b/lib/arrow/adjustment.ex
@@ -33,6 +33,6 @@ defmodule Arrow.Adjustment do
     adjustment
     |> cast(attrs, [:source, :source_label, :route_id])
     |> validate_required([:source, :source_label, :route_id])
-    |> unique_constraint(:source_source_label, name: :adjustments_source_source_label)
+    |> unique_constraint(:source_label)
   end
 end
--- a/lib/arrow/adjustment_fetcher.ex
+++ b/lib/arrow/adjustment_fetcher.ex
@@ -1,0 +1,50 @@
+defmodule Arrow.AdjustmentFetcher do
+  @moduledoc """
+  Reads in information on what adjustments are present in GTFS Creator
+  and saves them to the database. Currently reads from the `priv`
+  directory once on startup, but eventually this will poll some sort
+  of API on a regular basis.
+  """
+
+  use GenServer, restart: :transient
+
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, [])
+  end
+
+  @impl true
+  def init(opts \\ []) do
+    :ok =
+      opts[:path]
+      |> File.read!()
+      |> Jason.decode!()
+      |> update_adjustment_data!()
+
+    :ignore
+  end
+
+  @spec update_adjustment_data!(map()) :: :ok
+  defp update_adjustment_data!(parsed_adjustments) do
+    timestamp = DateTime.truncate(DateTime.utc_now(), :second)
+
+    adjustments =
+      Enum.map(parsed_adjustments, fn parsed_adjustment ->
+        %{
+          source: "gtfs_creator",
+          source_label: parsed_adjustment["id"],
+          route_id: parsed_adjustment["attributes"]["route_id"],
+          inserted_at: timestamp,
+          updated_at: timestamp
+        }
+      end)
+
+    {_, _} =
+      Arrow.Repo.insert_all(Arrow.Adjustment, adjustments,
+        on_conflict: [set: [source: "gtfs_creator"]],
+        conflict_target: :source_label
+      )
+
+    :ok
+  end
+end
--- a/lib/arrow/adjustment_fetcher.ex
+++ b/lib/arrow/adjustment_fetcher.ex
@@ -8,6 +8,8 @@ defmodule Arrow.AdjustmentFetcher do
 
   use GenServer, restart: :transient
 
+  import Ecto.Query, only: [from: 2]
+
   @spec start_link(Keyword.t()) :: GenServer.on_start()
   def start_link(opts) do
     GenServer.start_link(__MODULE__, opts, [])
@@ -44,6 +46,25 @@ defmodule Arrow.AdjustmentFetcher do
         on_conflict: [set: [source: "gtfs_creator"]],
         conflict_target: :source_label
       )
+
+    adjustment_source_labels = Enum.map(adjustments, & &1.source_label)
+
+    unused_adjustment_ids =
+      Arrow.Repo.all(
+        from d in Arrow.Disruption,
+          right_join: a in assoc(d, :adjustments),
+          select: a.id,
+          where: is_nil(d.id)
+      )
+
+    Arrow.Repo.delete_all(
+      from(
+        a in Arrow.Adjustment,
+        where:
+          a.source == "gtfs_creator" and a.source_label not in ^adjustment_source_labels and
+            a.id in ^unused_adjustment_ids
+      )
+    )
 
     :ok
   end
--- a/lib/arrow/application.ex
+++ b/lib/arrow/application.ex
@@ -13,6 +13,7 @@ defmodule Arrow.Application do
       [
         # Start the Ecto repository
         Arrow.Repo,
+        {Arrow.AdjustmentFetcher, path: Application.app_dir(:arrow, "priv/repo/shuttles.json")},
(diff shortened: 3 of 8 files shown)


priv/repo/migrations/20200129212636_update_adjustments_unique_index.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Arrow.Repo.Migrations.UpdateAdjustmentsUniqueIndex do
+  use Ecto.Migration
+
+  def change do
+    drop index(:adjustments, [:source, :source_label], name: :adjustments_source_source_label)
+
+    create unique_index(:adjustments, [:source_label], name: :adjustments_source_label)
```
losvedir: Providing the `:name` was necessary before because it was using a combination of columns. You should be able to just use the implicit name now, I think: [docs](https://hexdocs.pm/ecto/Ecto.Changeset.html#unique_constraint/3).

lemald: Ah, yup, that seemed to work.

test/arrow/adjustment_fetcher_test.exs
```diff
@@ -0,0 +1,60 @@
...
+        source: "arrow",
+        source_label: "foo",
+        route_id: "bar"
+      }
+
+      {:ok, _new_adj} = Arrow.Repo.insert(adj)
+
+      :ignore = Arrow.AdjustmentFetcher.init(path: @test_json_filename)
+
+      foo_adjustment =
```
losvedir: This test would still pass, I believe, if it inserted a new record rather than updating the existing one. We know that's not the case here because of the unique index, but that seems a little implicit to me. What about directly asserting that there's just one record after the fact, and it looks how we expect? Something like:

```ex
assert [%Arrow.Adjustment{source_label: "foo", source: "gtfs_creator"}] = Arrow.Repo.all(Arrow.Adjustment)
```

lemald: Ah, so previously that wouldn't have worked because I wasn't clearing out old entries, but now that I am I think it will.

lib/arrow/adjustment_fetcher.ex
```diff
@@ -0,0 +1,59 @@
...
+  def init(opts \\ []) do
+    opts[:path]
+    |> File.read!()
+    |> Jason.decode!()
+    |> update_adjustment_data()
+
+    :ignore
+  end
+
+  @spec update_adjustment_data(map()) :: nil
```
paulswartz: The general pattern in Erlang/Elixir is for functions which succeed or crash is:
- end with "!"
- return `:ok`

lib/arrow/adjustment_fetcher.ex
```diff
@@ -0,0 +1,59 @@
...
+  defp update_adjustment_data(parsed_adjustments) do
+    timestamp = DateTime.utc_now() |> DateTime.truncate(:second)
```
paulswartz: I'm not sure why Credo doesn't notice this, but generally we don't have pipelines with a single pipe.

lib/arrow/application.ex
```diff
@@ -14,9 +14,16 @@ defmodule Arrow.Application do
         # Start the Ecto repository
         Arrow.Repo,
         # Start the endpoint when the application starts
-        ArrowWeb.Endpoint
+        ArrowWeb.Endpoint,
         # Starts a worker by calling: Arrow.Worker.start_link(arg)
         # {Arrow.Worker, arg},
+        %{
+          id: Arrow.AdjustmentFetcher,
```
paulswartz: Can you switch this to `{Arrow.AdjustmentFetcher, path: Application.app_dir(...)}` and have the rest in `AdjustmentFetcher.child_spec/1`?

priv/repo/migrations/20200129212636_update_adjustments_unique_index.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Arrow.Repo.Migrations.UpdateAdjustmentsUniqueIndex do
+  use Ecto.Migration
+
+  def change do
+    drop index(:adjustments, [:source, :source_label], name: :adjustments_source_source_label)
+
+    create unique_index(:adjustments, [:source_label])
```
paulswartz: It probably doesn't matter now, but do we need to handle the case where `source_label` isn't unique when this migration happens?

lib/arrow/adjustment.ex
```diff
@@ -33,6 +33,6 @@ defmodule Arrow.Adjustment do
     adjustment
     |> cast(attrs, [:source, :source_label, :route_id])
     |> validate_required([:source, :source_label, :route_id])
-    |> unique_constraint(:source_source_label, name: :adjustments_source_source_label)
+    |> unique_constraint(:source_label, name: :adjustments_source_label)
```
losvedir: You don't need to specify the name here anymore, either.

lib/arrow/application.ex
losvedir: In fact, since the only non-standard thing here is `restart: :transient`, I don't think you need to override `AdjustmentFetcher.child_spec/1` but can actually just do `use GenServer, restart: :transient` in that file.

priv/repo/migrations/20200129212636_update_adjustments_unique_index.exs
lemald: Hmm, that's a good point, but I don't see any good way of handling that from looking at the docs for `unique_index/3` and in turn `index/3` [here](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#unique_index/3)...

lib/arrow/adjustment.ex
lemald: So interestingly here if I remove the name it actually changes the behavior - it throws an exception when the constraint is violated rather than returning an `{:error, _}` tuple. I'm not sure if that's actually what we want to do.

lib/arrow/adjustment_fetcher.ex
lemald: Fixed

lib/arrow/adjustment_fetcher.ex
lemald: Fixed

lib/arrow/application.ex
lemald: Fixed, ended up going with @losvedir's suggestion that avoided `child_spec/1` altogether.

priv/repo/migrations/20200129212636_update_adjustments_unique_index.exs
losvedir: That table should be empty for all us developers as well as on arrow-dev, and no previous migrations populate it so future arrow-prod won't have the problem either. So I don't think we have to worry about it here.

lib/arrow/adjustment_fetcher.ex
```diff
@@ -0,0 +1,60 @@
...
+      Arrow.Repo.insert_all(Arrow.Adjustment, adjustments,
+        on_conflict: [set: [source: "gtfs_creator"]],
+        conflict_target: :source_label
+      )
+
+    adjustment_source_labels = Enum.map(adjustments, & &1.source_label)
+
+    from(
+      a in Arrow.Adjustment,
+      where: a.source == "gtfs_creator" and a.source_label not in ^adjustment_source_labels
```
losvedir: We also don't want to do delete any adjustments which are part of old disruptions, even if they're not in the file anymore. I believe, in that situation actually it would 💥 here because of FK constraints.

lemald: Oh, good catch.

lib/arrow/adjustment.ex
losvedir: Hmm... that seems very odd to me. Did you run the migration down and up on the test DB? I see [this question](https://elixirforum.com/t/unique-constraint-i-dont-know-how-to-use-name/16339) which seems to imply that the behavior differs not on whether or not `:name` is included, but if it matches a DB constraint. So I'm wondering if the test DB migrations are consistent with the schema.

lemald: Ah, you're right, I tried just doing `MIX_ENV=test mix ecto.drop` and then recreating the DB and running the migration and it all worked.

PR #79, opened 2020-02-20
feat: agenda view

fjlanasa:
Asana ticket: [🏹 agenda view](https://app.asana.com/0/584764604969369/1148734524638915/f)

### Changes
Added UI components to display "agenda view" of disruptions. 

Once the shape of the eventual API response feeding this view is determined, will need to add logic (either in `DisruptionIndex` component or a separate container component) to fetch this data and and parse it into the format defined by the interface `Disruption`

```
export interface Disruption {
  id: string
  route: Routes
  label: string
  startDate: Date
  endDate: Date
  daysAndTimes: string
}
```

### Base view
Screen Shot 2020-02-21 at 9 57 38 AM

### Filter by search query
Screen Shot 2020-02-21 at 9 57 48 AM

### Filter by route
Screen Shot 2020-02-21 at 9 57 59 AM


assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -0,0 +1,224 @@
...
+  | "Green-C"
+  | "Green-D"
+  | "Green-E"
+  | "Commuter"
+
+type RouteFilterState = {
+  [route in Routes]?: boolean
+}
+
+export interface Disruption {
```
losvedir: I'm not sure the best way to organize this, but calling this `Disruption` seems like it be a namespace conflict in the future. In particular, when we flesh out creating a new disruption or editing one, we'll probably one an `interface` for that Disruption as well, and it carries with it a lot more information (exception dates, days and times as a list of tuples rather than a simple string, etc.).

The structure here is basically exactly what's needed for the Index page table. So what about calling this `DisruptionRow` or something like that?

Alternatively, we could bite the bullet and figure out what we want a `Disruption` to actually be, and then you could update this PR to take one of those and do the wrangling logic to get, e.g. `daysAndTimes` out of the `DayOfWeekTimeRanges` that they are likely to be.

But I think it's fine to go with the simpler approach now, and we'll unify them (if it makes sense) once they're all fleshed out, to see what the needs are.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -0,0 +1,224 @@
...
+  id: string
+  route: Routes
```
losvedir: A Disruption can actually have multiple routes and labels. In the terminology of our DB and models an "adjustment" is a (label, route) pair, and a "disruption" has many adjustments.

losvedir: And I don't mean necessarily that you need to do the type unification I mention in my other comment, so much as the functionality in this table here is that the filtering should be able to filter against any of the adjustments. Can you also update the mocked data to demonstrate this ability?

Also, I'm not super sure how sorting will work, now that I think about it... it's not super well defined to sort the collection of adjustments, so just do whatever sort of works, or if not we can revisit whether we want sorting there at all.

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: Yeah, that makes sense. I had originally called it `DisruptionListItem` (can't really remember why I changed it) so it'd be clear that it's describing the expected shape for this UI component. I'll change to `DisruptionRow`.

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: OK, will update.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -0,0 +1,224 @@
...
+  label: string
+  startDate: Date
+  endDate: Date
+  daysAndTimes: string
+}
+
+interface RouteFilterToggleProps {
```
losvedir: We like to keep the `Props` interface right above the component it's for, since if you change one you usually have to change the other.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -0,0 +1,224 @@
...
+    } else {
+      setView("table")
+    }
+  }, [view, setView])
+
+  const [searchQuery, setSearchQuery] = React.useState<string>("")
+  const [routeFilters, updateRouteFilters] = React.useState<RouteFilterState>(
+    {}
+  )
+  const toggleRouteFilterState = React.useCallback(
```
losvedir: This is neat. My previous experience with React was before hooks, so I'm still getting up to speed with all of them. From the docs it looks like this is a small wrapper around `useMemo`, and `useMemo` is simply a performance optimization. For my own edification, how/why did you decide to use them here? Performance issues? Best practice to just use them regardless?

Also, the docs say:

> We recommend using the exhaustive-deps rule as part of our eslint-plugin-react-hooks package. It warns when dependencies are specified incorrectly and suggests a fix.

We use `tslint` rather than `eslint`, but if we're going to be using these hooks, can you maybe spend a couple minutes to see if it's easy to set up our linter to catch these like it suggests?

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -0,0 +1,224 @@
...
+type Routes =
+  | "Red"
+  | "Blue"
+  | "Mattapan"
+  | "Orange"
+  | "Green-B"
+  | "Green-C"
+  | "Green-D"
+  | "Green-E"
+  | "Commuter"
```
losvedir: "Commuter" reads a little funny to me. Also, in contrast to the others "commuter rail" is a mode (or sometimes "type") akin to subway rather than a route. The routes are, e.g., `CR-Franklin`, `CR-Worcester`, etc.

I think it's fine to call this type "Routes", and in the future we might break up the CR into its respective routes. Maybe we can call this `"CR-All"` or something to signify that it represents all the CR routes?

And then the filtering will have to have special handling there, too. If you look in `priv/repo/shuttles.json`, you'll see our currently defined adjustments. (Yes, each "shuttle" – what `gtfs_creator` calls it – is an "adjustment" here.) All of the currently defined CR ones have a `route_id` like `CR-[....]`. So maybe the filtering can look for the route prefix of `CR-`.

Also, can you update the mocked data to have a CR adjustment to demonstrate this?

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: Yeah, I tend to default to using them, just as to not perform any unnecessary computation (`useMemo`) or references to functions (`useCallback`), though it's probably premature in this case. I think of `useCallback` as similar to binding a function in a class-based parent component to pass to a child, rather than passing the child an inline anonymous function, preserving referential equality of the function passed by the parent.

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: Yeah I guess I was envisioning that the real route IDs for commuter rail routes would be mapped to this (or `"CR-ALL"`) when we are parsing the API response into the structure this UI component is using, similar to how we'll have to format the data showing up in the table columns.

fjlanasa: The thought was that this table doesn't really care about the specific commuter rail line, and mapping all commuter rail routes into one bucket makes the filtering logic easier.

losvedir: Ah, I see. Yeah, that makes sense.

assets/src/disruptions/disruptionIndex.tsx
losvedir: Interesting, thanks.

assets/src/disruptions/disruptionTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+    </th>
+  )
+}
+
+interface SortState {
+  by: "label" | "startDate"
+  order: "asc" | "desc"
+}
+
+const formatDate = (date: Date): string =>
```
losvedir: We have a similar function `formatDisruptionDate` in `disruptions.ts`. Maybe we could consolidate them?

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  const [searchQuery, setSearchQuery] = React.useState<string>("")
+  const [routeFilters, updateRouteFilters] = React.useState<RouteFilterState>(
+    {}
+  )
+  const toggleRouteFilterState = React.useCallback(
+    (route: keyof RouteFilterState) => {
+      updateRouteFilters({ ...routeFilters, [route]: !routeFilters[route] })
+    },
+    [routeFilters, updateRouteFilters]
```
losvedir: These `useCallback` and `useMemo` hooks are really interesting. I was thinking through how they worked a little more, and had the question of whether the `updateRouteFilters` function returned by `useState` is the same each time. If it _is_ the same, then it's not necessary to list here. If it's _not_ the same, then it seems like the `useCallback` doesn't really do anything in this case, since the `updateRouteFilters` function would be new each time. But I found this note in the [docs](https://reactjs.org/docs/hooks-reference.html#usestate):

> React guarantees that `setState` function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the `useEffect` or `useCallback` dependency list.

So that answers that. And I think that means in general we don't have to list `updateRouteFilters` and similar state-updating callbacks in any of these dependency lists.

fjlanasa: Fair enough, though I'm pretty sure if we were to get the "exhaustive dependencies" rule working with ts-lint it would require it.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -0,0 +1,222 @@
...
+            type="text"
+            value={searchQuery}
+            onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+              setSearchQuery(e.target.value)
+            }
+            placeholder="Search Disruptions"
+          />
+          <h6>Filter by route</h6>
+          <div className="mb-1">
+            <RouteFilterToggle
```
losvedir: Is it maybe worth pulling all these into a `["Blue", "Mattapan", ...].map(....)`? Or would that end up causing more re-rendering than it needs otherwise? I'm not sure how exactly React would keep track of the mapped elements across re-renders. Maybe it would just rely on the `key=`?

assets/src/disruptions/disruptionIndex.tsx
losvedir: Oh, we should follow the lint. Were you able to set that up then?

fjlanasa: I haven't been able to set it up with tslint. Moving from tslint to eslint seems simple enough, but I see there is a ticket for that in the backlog. Maybe we should just handle this then?

losvedir: Yeah, that's fine. Can you make a note in that ticket to set this up then?

fjlanasa: Yes, will do

PR #91, opened 2020-03-03
feat: Style disruption 'day of week' selector

losvedir:
Asana: [🏹 Days of week should be stylized circles rather than checkboxes](https://app.asana.com/0/584764604969369/1158326786257803).

Screen Shot 2020-03-03 at 3 40 45 PM

### DIFF
--- a/assets/css/_forms.scss
+++ b/assets/css/_forms.scss
@@ -17,6 +17,6 @@
   }
 
   & input:checked + label {
-    background-color: rgb(182, 225, 178);
+    background-color: rgb(46, 139, 224);
   }
 }
--- a/assets/css/_forms.scss
+++ b/assets/css/_forms.scss
@@ -1,3 +1,22 @@
 .m-forms__sublegend {
   font-weight: bold;
 }
+
+.m-forms__day-of-week-bubble {
+  & input {
+    display: none;
+  }
+
+  & label {
+    border: 1px solid #aaa;
+    border-radius: 1.25rem;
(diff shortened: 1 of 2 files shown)


assets/css/_forms.scss
```diff
@@ -1,3 +1,22 @@
...
+    border: 1px solid #aaa;
+    border-radius: 1.25rem;
+    height: 2.5rem;
+    padding-top: 0.4rem;
+    text-align: center;
+    width: 2.5rem;
+  }
+
+  & input:checked + label {
+    background-color: rgb(182, 225, 178);
```
lemald: Just curious, why use the green-ish color? The Figma mockup has an RGB value of 46, 139, 224.

losvedir: Aha! I knew there was a way to see the color in the Figma mockup, but I couldn't find it. How'd you get that? Are you logged in?

I got that number using the built in Digital Color Meter app and pointing at it with the dropper. But I think that might do funky things like pick up modified colors via my monitor's color profile or something.

Will change.

lemald: I think you have to be logged in, but if you are you can just click on an element on the page and it will bring up the relevant CSS styling for it.

losvedir: Oooh, I see! I was looking at the older mockups far off to the right, where they _are_ greenish.

lemald: Oh, hah, I totally forgot those other mockups existed.

PR #92, opened 2020-03-03
feat: disruption time of day picker

fjlanasa:
Asana ticket: [🏹 Time of day selectors should have AM/PM widget](https://app.asana.com/0/584764604969369/1158326786257805/f)

Screen Shot 2020-03-03 at 4 49 56 PM

### DIFF
--- a/assets/src/disruptions/disruptionTimePicker.tsx
+++ b/assets/src/disruptions/disruptionTimePicker.tsx
@@ -7,7 +7,31 @@ import DatePicker from "react-datepicker"
 
 import { indexToDayOfWeekString } from "./disruptions"
 
-type TimeRange = [string, string]
+type HourOptions =
+  | "12"
+  | "1"
+  | "2"
+  | "3"
+  | "4"
+  | "5"
+  | "6"
+  | "7"
+  | "8"
+  | "9"
+  | "10"
+  | "11"
+
+type MinuteOptions = "00" | "15" | "30" | "45"
+
+type PeriodOptions = "AM" | "PM"
+
+interface Time {
+  hour: HourOptions
+  minute: MinuteOptions
+  period: PeriodOptions
+}
+
+type TimeRange = [Time | null, Time | null]
 
 type DayOfWeekTimeRanges = [
   TimeRange | null,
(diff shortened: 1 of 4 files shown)


assets/src/disruptions/disruptionTimePicker.tsx
```diff
@@ -7,7 +7,31 @@ import DatePicker from "react-datepicker"
 
 import { indexToDayOfWeekString } from "./disruptions"
 
-type TimeRange = [string, string]
+type HourOptions =
+  | "12"
```
losvedir: I remember bringing this up in grooming that maybe `"12"` should go first, since that's how it works with am/pm, but I thought the consensus was that people wanted it to go last, since that's how numbers work.

Also, what do you think about these being numbers instead of strings? My inclination is to have the types be semantically meaningful _within_ the application and serialize/deserialize at the edges. In this case, that would mean stringifying them into the selects (since those _have_ to be strings), and converting them to numbers in the event handlers before storing them in state. I'm not married to the idea, though, if that doesn't play super well with react conventions and would cause issues.

fjlanasa: I'm fine with changing the order so `"12"` is last if that's what we want to do.

I don't have strong opinions about whether we store them as numbers vs strings, but wondering what you're envisioning the format would be if we went the numbers route. 

Would we still keep the range of hour options from `1-12`, or would we be storing them from `0-23`? And will the minutes be `0, 15, 30, 60` (`"00"` would be parsed into `0`)?. It seems like either way we'll need to do some manipulation at the time of submit/request to get the values into the format the server will be expecting.

losvedir: My gut feeling for "semantic" state would be an hour of `0-23`, and minutes of `0, 15, 30, 45`, and AM/PM inferred from the hour (unless is it possible to have AM/PM selected but no time?). That would make the API call, which we'll probably use ISO-8601 (`"23:30:00"`) for, pretty straightforward.

On the other hand, it would make interactions with the form more complex. I suppose having the state more directly match what's being displayed, and tucking the conversion complexity away to just when it's time to hit the API, makes sense.

fjlanasa: OK - that makes sense. If it's all the same to you, I'll just leave as is and leave the conversion to be handled on submit.

PR #93, opened 2020-03-04
feat: front-end edit page

lemald:
This should be all of the front-end logic that's feasible to do for now without having the API to actually save changes.

Right now I just have the "cancel" button redirecting to the view page URL for the corresponding disruption, but we might want to change that in the future if there are multiple ways that the user can reach the edit page.

Also note that I've included some dummy data just so that we can actually see the functionality of the edit page. We should remove this as soon as we have an API we can read from.

### DIFF
--- a/assets/src/disruptions/disruptionSummary.tsx
+++ b/assets/src/disruptions/disruptionSummary.tsx
@@ -2,10 +2,12 @@ import * as React from "react"
 import { Adjustment } from "./disruptions"
 
 interface DisruptionSummaryProps {
+  disruptionId?: string
   adjustments: Adjustment[]
 }
 
 const DisruptionSummary = ({
+  disruptionId,
   adjustments,
 }: DisruptionSummaryProps): JSX.Element => {
   return (
@@ -15,6 +17,13 @@ const DisruptionSummary = ({
           <span className="m-disruption-summary__adjustment_label">
             {adjustment.label}
           </span>
+          {disruptionId && (
+            <div>
+              <span className="m-disruption-summary__disruption_id">
+                Disruption ID: {disruptionId}
+              </span>
+            </div>
+          )}
           <div>
             <span className="m-disruption-summary__adjustment_route">
               {adjustment.route}
--- a/assets/src/disruptions/editDisruption.tsx
+++ b/assets/src/disruptions/editDisruption.tsx
@@ -2,8 +2,11 @@ import * as React from "react"
 import Button from "react-bootstrap/Button"
 import ButtonGroup from "react-bootstrap/ButtonGroup"
 import { RouteComponentProps } from "react-router-dom"
+import { Redirect } from "react-router"
 
 import Header from "../header"
+import { DisruptionPreview } from "./disruptionPreview"
+import { DisruptionSummary } from "./disruptionSummary"
 import {
   DayOfWeekTimeRanges,
   DisruptionTimePicker,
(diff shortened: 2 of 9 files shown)


assets/src/disruptions/disruptionSummary.tsx
```diff
@@ -15,6 +17,13 @@ const DisruptionSummary = ({
           <span className="m-disruption-summary__adjustment_label">
             {adjustment.label}
           </span>
+          {disruptionId ? (
```
fjlanasa: small nitpick, but you can just do this rather than using a ternary:
```
{ disruptionId && 
     <SomeConditionallyRenderedContent />
}
```

assets/src/disruptions/disruptionSummary.tsx
```diff
@@ -2,10 +2,12 @@ import * as React from "react"
 import { Adjustment } from "./disruptions"
 
 interface DisruptionSummaryProps {
+  disruptionId: string | null
```
fjlanasa: optional: set `disruptionId?: string` rather than `string | null` and we wont have to explicitly pass `null` in the various places we are right now, and this component will otherwise work the same.

assets/src/disruptions/editDisruption.tsx
```diff
@@ -13,53 +16,153 @@ interface TParams {
...
+  disruptionId: string | null
+  fromDate: Date | null
+  setFromDate: React.Dispatch<Date | null>
+  toDate: Date | null
+  setToDate: React.Dispatch<Date | null>
+  exceptionDates: Date[]
+  setExceptionDates: React.Dispatch<Date[]>
+  disruptionDaysOfWeek: DayOfWeekTimeRanges
+  setDisruptionDaysOfWeek: React.Dispatch<DayOfWeekTimeRanges>
+  redirectToView: boolean
```
fjlanasa: Do we need this prop (`redirectToView`), seems like we are using it to set the initial value of `redirect`, but it's hardcoded as `false` when we use this component above. Maybe we can just set it to `false` down below like `isPreview`?

assets/src/disruptions/editDisruption.tsx
```diff
@@ -13,53 +16,153 @@ interface TParams {
...
+  const [redirect, setRedirect] = React.useState(redirectToView)
+  const [isPreview, setIsPreview] = React.useState<boolean>(false)
+
+  const adjustment = {
+    label: "Kenmore - Newton Highlands",
+    route: "Green-D",
+  }
+
+  if (redirect && disruptionId) {
+    return <Redirect to={"/disruptions/" + encodeURIComponent(disruptionId)} />
```
fjlanasa: You could also use the `useHistory` hook provided by `react-router-dom` to imperatively redirect on clicking save. 

https://reacttraining.com/react-router/web/api/Hooks/usehistory

assets/src/disruptions/editDisruption.tsx
```diff
@@ -13,53 +16,153 @@ interface TParams {
...
   )
 }
 
 const EditDisruption = ({
   match,
 }: RouteComponentProps<TParams>): JSX.Element => {
-  const [fromDate, setFromDate] = React.useState<Date | null>(null)
-  const [toDate, setToDate] = React.useState<Date | null>(null)
-  const [exceptionDates, setExceptionDates] = React.useState<Date[]>([])
+  const [fromDate, setFromDate] = React.useState<Date | null>(
```
fjlanasa: I don't know how we'd normally handle this, but maybe a TODO comment explaining these are dummy values that should be removed later?

assets/src/disruptions/editDisruption.tsx
lemald: Good catch, it's not needed.

assets/src/disruptions/editDisruption.tsx
lemald: Interesting... are there any particular advantages to this approach besides readability?

fjlanasa: No, I don't think so. I just find it more readable.

lemald: So after poking at this a bit, it looks like the resulting code would be slightly more readable but also harder to test, and I'm not sure that the change would be on net worth it. I think I'll leave it as-is.

PR #103, opened 2020-03-09
feat: disruptions list api

fjlanasa:
Asana ticket: [🏹 provide an API for fetching disruptions active in a date range](https://app.asana.com/0/584764604969369/1145845426232967/f)

Changes:
* added [`ja_serializer`](https://github.com/mbta/ja_serializer)
* added `DisruptionApiController` and index route, which can include specific relationships and filter disruptions by date
* added views for disruption and relationships


lib/arrow_web/router.ex
```diff
@@ -10,7 +10,9 @@ defmodule ArrowWeb.Router do
   end
 
   pipeline :api do
-    plug :accepts, ["json"]
+    plug :accepts, ["json-api"]
+    plug JaSerializer.ContentTypeNegotiation
+    plug JaSerializer.Deserializer
```
losvedir: It seems like in [the docs](https://hexdocs.pm/ja_serializer/JaSerializer.Deserializer.html) all this does is set hyphens to underscores? If so, I think we shouldn't do that. I think it's better to be strict about what we accept, and just request the underscores to begin with. Unless it does other stuff, too?

lib/arrow_web/router.ex
```diff
@@ -57,6 +59,12 @@ defmodule ArrowWeb.Router do
     get("/:provider/callback", AuthController, :callback)
   end
 
+  scope "/api", ArrowWeb do
+    pipe_through([:redirect_prod_http, :api, :browser])
```
losvedir: It shouldn't pipe through, `:browser`, I think, since this is a JSON API and won't render HTML or accept forms or anything like that.

lib/arrow_web/router.ex
```diff
@@ -57,6 +59,12 @@ defmodule ArrowWeb.Router do
...
+
+    get("/disruptions/", DisruptionApiController, :index)
```
losvedir: Since we'll end up using most of the "restful" routes here, maybe better to do `resources("/disruptions", DisruptionApiController, only: [:index])` here? Then we can just update the `:only` list when we add creating, editing, etc.

lib/arrow_web/views/api_views/disruption_view.ex
```diff
@@ -0,0 +1,38 @@
...
+  has_many :exceptions,
+    serializer: ArrowWeb.ExceptionApiView,
+    include: true
+
+  has_many :trip_short_names,
+    serializer: ArrowWeb.TripShortNameApiView,
+    include: true
+
+  def adjustments(disruption, _conn) do
+    Arrow.Repo.all(Ecto.assoc(disruption, :adjustments))
```
losvedir: I'm not sure the best way to do this with JaSerializer, but in general I don't like tucking a bunch of DB work into the view. Can these be loaded in the controller somehow instead?

losvedir:
I'm going to suggest a tweak to the scoping / folders / names a little. If you take a look at, say, our API repo, it follows the convention of module namespacing with dots, and matching folder structure. However, I know that'll be sort of an annoying update, so I'm not super sure if it's worth it. But in any case, FYI, I think this is sort of how we'd do it conventionally in Phoenix:

Rather than `ArrowWeb.DisruptionApiController` in `arrow_web/controllers/disruption_api_controller.ex` and `ArrowWeb.DisruptionApiView` in `arrow_web/views/api_views/disruption_view.ex`, we could scope everything in an `API` namespace.

It would begin in the router with:

```elixir
scope "/api", ArrowWeb.API do
  resources("/disruptions", DisruptionController)
  # ..
end
```

Note `ArrowWeb.API` instead of `ArrowWeb` by itself. That means Phoenix will look for `ArrowWeb.API.DisruptionController`, which in turn should use `ArrowWeb.API.DisruptionView`, etc.

And in terms of file structure, it would be like `arrow_web/controllers/api/disruption_controller.ex`, etc.

I think providing a module pseudo-namespace (since dots don't really _mean_ anything in Elixir) and matching folder structure for each `scope` in the router is sort of conventional.

lib/arrow_web/router.ex
fjlanasa: Yeah, that's right. I'll remove.

lib/arrow_web/views/api_views/disruption_view.ex
fjlanasa: Yeah, I suppose I can `preload` all of these fields in the controller, if that's what we'd want to do.

fjlanasa:
> I'm going to suggest a tweak to the scoping / folders / names a little. If you take a look at, say, our API repo, it follows the convention of module namespacing with dots, and matching folder structure. However, I know that'll be sort of an annoying update, so I'm not super sure if it's worth it. But in any case, FYI, I think this is sort of how we'd do it conventionally in Phoenix:
> 
> Rather than `ArrowWeb.DisruptionApiController` in `arrow_web/controllers/disruption_api_controller.ex` and `ArrowWeb.DisruptionApiView` in `arrow_web/views/api_views/disruption_view.ex`, we could scope everything in an `API` namespace.
> 
> It would begin in the router with:
> 
> ```elixir
> scope "/api", ArrowWeb.API do
>   resources("/disruptions", DisruptionController)
>   # ..
> end
> ```
> 
> Note `ArrowWeb.API` instead of `ArrowWeb` by itself. That means Phoenix will look for `ArrowWeb.API.DisruptionController`, which in turn should use `ArrowWeb.API.DisruptionView`, etc.
> 
> And in terms of file structure, it would be like `arrow_web/controllers/api/disruption_controller.ex`, etc.
> 
> I think providing a module pseudo-namespace (since dots don't really _mean_ anything in Elixir) and matching folder structure for each `scope` in the router is sort of conventional.

Yeah, I went back and forth about this. I'm happy to change it, if you and @lemald don't mind adjusting your branches too.

losvedir:
I don't mind. File renames are sort of annoying to handle automatically (though maybe if you `git mv ...` it will be smart enough to merge cleanly, but I don't have a ton of files I'm working on, and at worst could just copy/paste over my work.

fjlanasa:
> I don't mind. File renames are sort of annoying to handle automatically (though maybe if you `git mv ...` it will be smart enough to merge cleanly, but I don't have a ton of files I'm working on, and at worst could just copy/paste over my work.

Done

lib/arrow_web/views/api/days_of_week_view.ex
```diff
@@ -0,0 +1,17 @@
+defmodule ArrowWeb.API.DaysOfWeekView do
+  use ArrowWeb, :view
+  use JaSerializer.PhoenixView
+
+  attributes([
+    :id,
+    :start_time,
+    :end_time,
+    :monday,
```
losvedir: @lemald and I had a slack convo the other day, and I think we had talked about changing this to `day_of_week: "monday"` or something to that effect (the model / DB being a day of week enum, rather than a bunch of booleans). @lemald - do you remember where we decided on that? I think I lean towards the enum approach, to help avoid the issue of denormalized / inconsistent data: what if two records have overlapping days of the week with conflicting times?

lemald: I think we landed on the decision that the enum approach would be good but didn't take any other steps (like create a ticket).

test/arrow_web/controllers/api/disruption_controller_test.exs
```diff
@@ -0,0 +1,214 @@
+defmodule ArrowWeb.DisruptionControllerTest do
```
losvedir: Should be `ArrowWeb.API.DisruptionControllerTest`.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -0,0 +1,53 @@
...
+    Enum.reduce(filters, from(d in Disruption), &compose_query/2)
+  end
+
+  defp compose_query({filter, date}, query) when filter in ["min_start_date", "min_end_date"] do
+    filter_field = String.to_atom(String.slice(filter, 4..-1))
+    where(query, [d], field(d, ^filter_field) > ^date)
+  end
+
+  defp compose_query({filter, date}, query) when filter in ["max_start_date", "max_end_date"] do
+    filter_field = String.to_atom(String.slice(filter, 4..-1))
```
losvedir: Per [the docs](https://hexdocs.pm/elixir/String.html#to_atom/1), this is unsafe:

> Warning: this function creates atoms dynamically and atoms are not garbage-collected. Therefore, string should not be an untrusted value, such as input received from a socket or during a web request. Consider using `to_existing_atom/1` instead.

`String.to_existing_atom` would work in this case, since we have `:start_date` used elsewhere in the application. However, and this is more of a personal preference thing, I kind of like being explicit about how exactly the different filters translate to queries, and wouldn't mind just being explicit about it:

```elixir 
def build_query(filters) do
  Enum.reduce(filters, from(d in Disruption), &compose_query/2)
end

defp compose_query({"min_start_date", date}, query), do: from(d in query, where: d.start_date >= ^date)
defp compose_query({"min_end_date", date}, query), do: from(d in query, where: d.end_date >= ^date)
defp compose_query({"max_start_date", date}, query), do: from(d in query, where: d.start_date <= ^date)
defp compose_query({"max_end_date", date}, query), do: from(d in query, where: d.end_date <= ^date)
```

But that's optional.

We do need to change from `String.to_atom`, though. And also, I think of "max date" as being the largest date that should be returned, so I think the `<` and `>` should be `<=` and `>=`.

losvedir: Edit: Ah, just realized this is only happening after you've already done `Map.take` on the filters, so I suppose in this case it's not unsafe. I still think we should probably avoid the function, though, since it feels like a trap lying in wait, if we add filters poorly in the future or something.

fjlanasa: Makes sense, good catch on `String.to_atom`. I'll go ahead and make the change to having a different function definition for each filter.

losvedir:
I just realized we don't have the GitHub PR template for this repo, but one of the checkboxes for the other apps is that every function should have a `@spec`.

fjlanasa:
> I just realized we don't have the GitHub PR template for this repo, but one of the checkboxes for the other apps is that every function should have a `@spec`.

Ah, right. On it.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -0,0 +1,61 @@
...
+
+    render(conn, "index.json-api",
+      data:
+        Repo.all(query)
+        |> Repo.preload([:adjustments, :days_of_week, :exceptions, :trip_short_names]),
+      opts: [include: Map.get(params, "include")]
+    )
+  end
+
+  @spec build_query(list(tuple())) :: Ecto.Query.t()
```
losvedir: Instead of `list` we generally use `[]`, and since we know the shape of the tuple, we would usually use it: `[{String.t(), Date.t()}]`.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -0,0 +1,61 @@
...
+  defp build_query(filters) do
+    Enum.reduce(filters, from(d in Disruption), &compose_query/2)
+  end
+
+  @spec compose_query(tuple(), Ecto.Query.t()) :: Ecto.Query.t()
```
losvedir: Similarly `{String.t(), Date.t()}`.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -0,0 +1,61 @@
...
+  defp compose_query({"min_end_date", date}, query),
+    do: from(d in query, where: d.end_date >= ^date)
+
+  defp compose_query({"max_start_date", date}, query),
+    do: from(d in query, where: d.start_date <= ^date)
+
+  defp compose_query({"max_end_date", date}, query),
+    do: from(d in query, where: d.end_date <= ^date)
+
+  @spec take_filters(map) :: map
```
losvedir: We use the function form for all specs: `map()`.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -0,0 +1,61 @@
...
+  defp take_filters(params) do
+    Map.take(Map.get(params, "filter", %{}), @filters)
+  end
+
+  @spec format_filters(map) :: list(tuple())
```
losvedir: `map()` and `[{String.t(), String.t()}]`.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -0,0 +1,61 @@
...
+  defp format_filters(filters) do
+    Enum.reduce(filters, [], fn filter, acc -> acc ++ do_format_filter(filter) end)
+  end
+
+  @spec do_format_filter(tuple()) :: list(tuple())
```
losvedir: Takes a `{String.t(), String.t()}` and returns a `[{String.t(), Date.t()}]`. This is an example where specifying the structure of the tuple is particularly helpful since you can see at a glance that this function probably does a date conversion.

PR #104, opened 2020-03-12
feat: Add models for json:api resources

losvedir:
Just some basic classes to represent the resources we'll be sending over the wire via json:api.


lemald:
Looks pretty good overall, minus the small formatting issue with the build. What's the plan for deserializing? Given that all of the scalar fields are optional, are we just assuming that there won't be much to actually translate or check when deserializing? What about array fields, given that they might be missing entirely from the JSON we get rather than just being an empty list?

PR #106, opened 2020-03-12
feat: Adjustment list api

fjlanasa:
Asana ticket: [🏹 JSON API to list all adjustments](https://app.asana.com/0/584764604969369/1157557244099563/f)

Changes:
* Added `ArrowWeb.API.AdjustmentController`
  * filter by `route_id`
  * filter by `source`
* Updated `Arrow.Application` to not read in adjustments on startup when testing

### DIFF
--- a/lib/arrow/application.ex
+++ b/lib/arrow/application.ex
@@ -7,18 +7,19 @@ defmodule Arrow.Application do
 
   def start(_type, _args) do
     run_migrations_at_startup? = Application.get_env(:arrow, :run_migrations_at_startup?)
-
+    run_adjustment_fetcher_at_startup? = Application.get_env(:arrow, :env) != :test
     # List all child processes to be supervised
     children =
       [
         # Start the Ecto repository
         Arrow.Repo,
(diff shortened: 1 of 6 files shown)


lib/arrow/application.ex
```diff
@@ -7,18 +7,19 @@ defmodule Arrow.Application do
 
   def start(_type, _args) do
     run_migrations_at_startup? = Application.get_env(:arrow, :run_migrations_at_startup?)
-
+    run_adjustment_fetcher_at_startup? = Application.get_env(:arrow, :env) != :test
```
losvedir: We use this pattern of selectively enabling features by environment in a few places (e.g. `:auto_process_gtfs?` in RTR is only true in prod). But rather than having a generic `:env` value which the application then infers its behavior from, can you make this a flag specific for its use, something like `Application.get_env(:arrow, :fetch_adjustments_at_startup?)`. It makes it clearer when looking at the config file what it's actually used for, and if we want to toggle the behavior of the app (say, doing this thing or not doing that thing in some environment) then it's just flipping a boolean in a config file, rather than re-writing code.

PR #140, opened 2020-04-01
Detail view

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 detail page ](https://app.asana.com/0/584764604969369/1163421197919737/f)

This one should definitely be reviewed commit-by-commit.

#### Reviewer Checklist

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -27,7 +27,7 @@ const apiCall = <T>({
     .then((response: Response) => response.json())
     .then((json: any) => parser(json))
     .catch(error => {
-      if (defaultResult === undefined) {
+      if (typeof defaultResult === "undefined") {
         throw error
       } else {
         return defaultResult
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -1,0 +1,37 @@
+const checkResponseStatus = (response: Response) => {
+  if (response.status === 200) {
+    return response
+  }
+
+  if (Math.floor(response.status / 100) === 3 || response.status === 403) {
+    // If the API sends us a redirect or forbidden, the user needs to
+    // re-authenticate. Reload to go through the auth flow again.
+    window.location.reload(true)
+  }
+
+  throw new Error(`Response error: ${response.status}`)
+}
+
+const apiCall = <T>({
+  url,
+  parser,
+  defaultResult,
+}: {
+  url: string
+  parser: (json: any) => T
+  defaultResult?: T
+}): Promise<T> =>
+  window
+    .fetch(url)
+    .then(checkResponseStatus)
+    .then((response: Response) => response.json())
+    .then((json: any) => parser(json))
+    .catch(error => {
+      if (defaultResult === undefined) {
+        throw error
+      } else {
+        return defaultResult
+      }
+    })
+
+export { apiCall }
--- a/assets/src/disruptions/disruptionTimePicker.tsx
+++ b/assets/src/disruptions/disruptionTimePicker.tsx
@@ -5,47 +5,16 @@ import Row from "react-bootstrap/Row"
 
 import DatePicker from "react-datepicker"
 
+import {
+  Time,
+  HourOptions,
+  MinuteOptions,
+  PeriodOptions,
+  TimeRange,
+  DayOfWeekTimeRanges,
+  isEmpty,
+} from "./time"
 import { indexToDayOfWeekString } from "./disruptions"
-
-type HourOptions =
-  | "12"
-  | "1"
-  | "2"
-  | "3"
-  | "4"
-  | "5"
-  | "6"
-  | "7"
-  | "8"
-  | "9"
-  | "10"
-  | "11"
-
-type MinuteOptions = "00" | "15" | "30" | "45"
-
-type PeriodOptions = "AM" | "PM"
-
-interface Time {
-  hour: HourOptions
-  minute: MinuteOptions
-  period: PeriodOptions
-}
-
-type TimeRange = [Time | null, Time | null]
-
-type DayOfWeekTimeRanges = [
-  TimeRange | null,
-  TimeRange | null,
-  TimeRange | null,
-  TimeRange | null,
-  TimeRange | null,
-  TimeRange | null,
-  TimeRange | null
-]
-
-const isEmpty = (days: DayOfWeekTimeRanges): boolean => {
-  return days.filter(d => d !== null).length === 0
-}
 
 interface DisruptionDateRangeProps {
   fromDate: Date | null
@@ -525,9 +494,4 @@ const DisruptionTimePicker = ({
   )
 }
 
-export {
-  DisruptionTimePicker,
-  DisruptionTimeRange,
-  TimeRange,
-  DayOfWeekTimeRanges,
-}
+export { DisruptionTimePicker, DisruptionTimeRange }
--- a/assets/src/disruptions/viewDisruption.tsx
+++ b/assets/src/disruptions/viewDisruption.tsx
@@ -1,7 +1,5 @@
 import * as React from "react"
-import Button from "react-bootstrap/Button"
-import { RouteComponentProps } from "react-router-dom"
-import { Redirect } from "react-router"
+import { RouteComponentProps, Link } from "react-router-dom"
 
 import { apiCall } from "../api"
 
@@ -18,20 +16,20 @@ interface TParams {
 }
 
 interface EditDisruptionButtonProps {
-  setRedirect: React.Dispatch<boolean>
+  disruptionId: string
 }
 
 const EditDisruptionButton = ({
-  setRedirect,
+  disruptionId,
 }: EditDisruptionButtonProps): JSX.Element => {
   return (
-    <Button
-      variant="primary"
-      onClick={() => setRedirect(true)}
-      id="edit-disruption-button"
+    <Link
+      to={"/disruptions/" + encodeURIComponent(disruptionId) + "/edit"}
+      id="edit-disruption-link"
+      className="btn btn-primary"
     >
       edit disruption times
-    </Button>
+    </Link>
   )
 }
 
@@ -48,7 +46,6 @@ interface ViewDisruptionFormProps {
 const ViewDisruptionForm = ({
   disruptionId,
 }: ViewDisruptionFormProps): JSX.Element => {
-  const [redirect, setRedirect] = React.useState<boolean>(false)
   const [disruption, setDisruption] = React.useState<
     Disruption | "error" | null
   >(null)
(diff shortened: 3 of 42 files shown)


assets/src/api.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (Math.floor(response.status / 100) === 3 || response.status === 403) {
+    // If the API sends us a redirect or forbidden, the user needs to
+    // re-authenticate. Reload to go through the auth flow again.
+    window.location.reload(true)
+  }
+
+  throw new Error(`Response error: ${response.status}`)
+}
+
+const apiCall = <T>({
```
losvedir, line 15: This looks good for the use case of fetching data and parsing it into our models, but there are a few ways it wouldn't work out of the box for my branch:  1) in my case I POST (meaning the `fetch` call would need to take a second parameter with options), 2) a successful response has status `201`, and 3) in the case of the error, I actually need the result back to display the errors on the page.

But I'm not sure I see a clear way of making this function work in both scenarios, so I think maybe this is fine for now, and once it's in and I rebase, I'll fiddle with it then.

lemald: For what it's worth, the code in this module and its tests are mostly ~stolen from~ inspired by Skate, which obviously doesn't need to do any POSTing in their current state. I'd be open to modifying it to make that easier if you have specific changes in mind.

assets/src/disruptions/disruptionTimePicker.tsx
```diff
@@ -43,10 +45,109 @@ type DayOfWeekTimeRanges = [
...
+    "10",
+    "11",
+  ].includes(hour)
+}
+
+const isMinuteOption = (minute: string): minute is MinuteOptions => {
+  return ["00", "15", "30", "45"].includes(minute)
+}
+
+const timeStringToTime = (
```
losvedir, line 83: I think this is the sort of helper function that would be nice to test independently of the component. Maybe break it out into a `disruptions/time.ts` file or something, and then have separate tests for that module? I just stashed my reverse of this function, `timeToTimeString` in this component module as well, but I had been wanting to move it into a time utilities module, too.

lemald: Yeah, that's a good idea. I'll go ahead and make a `disruptions/time.ts` file to store it.

lemald: @losvedir how do you feel about also moving the definitions of the `Time` and `DayOfWeekTimeRanges` types to that separate file?

losvedir: Sure, that makes sense to me.

assets/src/disruptions/disruptionTimePicker.tsx
```diff
@@ -43,10 +45,109 @@ type DayOfWeekTimeRanges = [
...
+  ]
+  let invalidTimeFound = false
+
+  daysOfWeek.forEach(dayOfWeek => {
+    if (dayOfWeek.day) {
+      const index = dayNamesToIndices[dayOfWeek.day]
+      const startTime = timeStringToTime(dayOfWeek.startTime)
+      const endTime = timeStringToTime(dayOfWeek.endTime)
+
+      if (startTime !== "error" && endTime !== "error") {
```
losvedir, line 136: Does this handle the case of either the start time or end time being `null` (to represent beginning/end of service)? Also, this is another function that would be nice to be tested separately.

losvedir: Ah, right, I see now `timeStringToTime` returns `Time | null | error`. Nice.

assets/src/disruptions/viewDisruption.tsx
```diff
@@ -0,0 +1,113 @@
...
+        setDisruption(result)
+      } else {
+        setDisruption("error")
+      }
+    })
+  }, [disruptionId])
+
+  if (redirect) {
+    return (
+      <Redirect
```
losvedir, line 72: Rather than having a click handler change the state which causes the component to render a redirect (which I find quite weird, but alas, have had to do this exact thing in my branch after creating a disruption), can you have the edit link be a `<Link to={...}>` instead?

assets/src/disruptions/disruptionTimePicker.tsx
lemald: Okay, just pushed up a commit that should make the change.

assets/src/jsonApi.ts
```diff
@@ -0,0 +1,54 @@
...
+
+type ModelObject =
+  | Adjustment
+  | DayOfWeek
+  | Disruption
+  | Exception
+  | TripShortName
+
+const toModelObject = (response: any): ModelObject | "error" => {
+  const includedObjects = response.included.map((raw: any) =>
```
losvedir, line 15: I'm a little surprised typescript isn't complaining here. We're saying `response` has type of `any`, but if its value is `"foo"` then won't `response.included` be `undefined` and then `undefined.map(...)` be a runtime error?

lemald: Oh, huh, that is surprising. I'll throw in a runtime check to make sure `response.included` is an array.

assets/src/disruptions/viewDisruption.tsx
lemald: I'm following the approach I took in #93 but I agree it's pretty weird, I'll experiment with the `<Link .../>` idea instead and see if I can get that working.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -17,6 +17,21 @@ defmodule ArrowWeb.API.DisruptionController do
...
+  @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def show(conn, params) do
+    try do
+      render(conn, "index.json-api",
+        data:
+          Repo.get!(Disruption, params["id"])
+          |> Repo.preload([:adjustments, :days_of_week, :exceptions, :trip_short_names]),
+        opts: [include: Map.get(params, "include")]
+      )
+    rescue
```
losvedir, line 29: Phoenix has some plumbing to automatically catch exceptions at the top of the application and render the appropriate error message. See [here](https://hexdocs.pm/phoenix/views.html#the-errorview). It also out-of-the-box handles identifying `Ecto.NoResultsError` as a `404`. So the idiomatic way to handle this would be to just `Repo.get!/2` here and let it take care of the displaying the 404.

That said, I'm not entirely sure how that interacts with `ja_serializer`, so if that's why you have the logic in this function, I think it might be better to just do a `Repo.get` and handle with normal case statements.

lib/arrow_web/views/error_view.ex
```diff
@@ -13,4 +14,13 @@ defmodule ArrowWeb.ErrorView do
   def template_not_found(template, _assigns) do
     Phoenix.Controller.status_message_from_template(template)
   end
+
+  def render("404.json" <> _, _assigns) do
```
losvedir, line 18: Oh, interesting, it looks like you've already updated this. Then.. I think handling this in the controller is unnecessary.

losvedir:
Whew, first pass done, and other than a few comments, looks good to me!

I have kind of a lingering question in my head about the `ModelObject` type and `JsonApiResourceObject` abstract class. It might be nice to make the `fromJson` parsing in each model a part of the abstract class interface. I'm not sure what the return value should be, though, maybe `JsonApiResourceObject` itself? Can you return an abstract class? But then that's not a good class to represent model objects, so maybe it should be renamed... I dunno. I'll noodle it a bit over lunch.

lemald:
@losvedir yeah that's a really good question with regards to making `fromJsonObject` part of an abstract class, I also wondered about that myself. I'll be curious to hear what further thoughts you have.

lib/arrow_web/controllers/api/disruption_controller.ex
lemald: Hmm, so if I remove the `try` / `rescue` block it does correctly turn the exception into a 404 response, but it returns a generic 404 rather than the results of the code I wrote in `error_view.ex` to return the JSON:API 404 message. Is there something else I need to do to make Phoenix realize that that's the right error message to return?

losvedir: Try changing `debug_errors: false` in `config/dev.exs` to see it. Otherwise in dev it doesn't go through this flow and tries to show the nice stack traces and such.

I just did that and got the json:api error response with your code as-id (other than removing the `try` / `rescue`).

assets/src/jsonApi.ts
lemald: Okay I just pushed up a change to fix this as well. I don't _love_ it since it involves invoking `includedObjects as ModelObject[]` once we've filtered out any `"error"` values, but I'd prefer if I could do that in some way that the type system can actually verify rather than just taking my word for it.

lib/arrow_web/controllers/api/disruption_controller.ex
lemald: Ahh, I see. Okay, I'll remove it then.

losvedir:
> yeah that's a really good question with regards to making `fromJsonObject` part of an abstract class, I also wondered about that myself. I'll be curious to hear what further thoughts you have.

Hmm, I was just looking into this more, and I guess TypeScript classes can also [implement interfaces](https://www.typescriptlang.org/docs/handbook/interfaces.html#class-types). So that could be another path to explore. But I don't have a clear preference for any approach here, so I think it's probably fine as-is for now.

assets/src/disruptions/viewDisruption.tsx
lemald: Okay, so I got it working, but rather than using `<Link .../>` I used the `useHistory` hook provided by `react-router`. Let me know what you think.

lemald: (It's a different approach but accomplishes the same basic goal of getting rid of the weird conditional rendering of the `<Redirect .../>` component)

assets/src/disruptions/viewDisruption.tsx
```diff
@@ -18,16 +17,22 @@ interface TParams {
...
+  disruptionId,
 }: EditDisruptionButtonProps): JSX.Element => {
+  const history = useHistory()
+
   return (
     <Button
       variant="primary"
-      onClick={() => setRedirect(true)}
+      onClick={() =>
+        history.push(
```
losvedir, line 32: So were you not able to replace this button with a `<Link to={`/disruptions/${disruptionId}/exit`}>` component (a la the "create new disruption" button in `disruptionIndex.tsx`? What was the issue with that?

Otherwise, I'd say you can revert this commit. The React approach is all about reactive rendering based on state. So in that sense `<Redirect>` is preferred to `history.push`, I believe. I was just thinking we could skip the callback and everything if we used a `Link`, which is what seemed semantic here to me.

lemald: The main issue I'm running into is giving the `Link` the right class to emulate the look of the `Button`... just giving it `btn-primary` doesn't work on its own. I can try to figure that out.

lemald: Oh hah I just had to give it `btn` as well. One sec and I'll check that in.

PR #147, opened 2020-04-06
Create disruption on frontend via API call

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 API to create disruption, frontend uses it](https://app.asana.com/0/584764604969369/1157557244099553)

I'm still trying to finesse how I organize the frontend changes around the network call, for the purpose of abstraction and testing, but the functionality should be there, and I think the backend is done.

The commits to review in order:

* A small refactor to rename `apiCall` to `apiGet`, since my code works pretty differently. I plan to move my API calling logic into this module as `apiPost`, which should work fairly similarly, but will, e.g., look at different response statuses and take an error parser in addition to a success parser.
* The backend changes to accept a json:api POST and create the relevant records.
* The frontend changes for the `NewDisruption` component to digest its state and POST it to the backend.

#### Reviewer Checklist

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -1,15 +1,51 @@
+interface Result<T, E> {
+  ok?: T
+  error?: E
+}
+
+const redirectIfUnauthorized = (status: number) => {
+  if (Math.floor(status / 100) === 3 || status === 403) {
+    // If the API sends us a redirect or forbidden, the user needs to
+    // re-authenticate. Reload to go through the auth flow again.
+    window.location.reload(true)
+  }
+}
+
 const checkResponseStatus = (response: Response) => {
   if (response.status === 200) {
     return response
   }
 
-  if (Math.floor(response.status / 100) === 3 || response.status === 403) {
-    // If the API sends us a redirect or forbidden, the user needs to
-    // re-authenticate. Reload to go through the auth flow again.
-    window.location.reload(true)
+  redirectIfUnauthorized(response.status)
+  throw new Error(`Response error: ${response.status}`)
+}
+
+const apiPost = async <T, E>({
+  url,
+  json,
+  successParser,
+  errorParser,
+}: {
+  url: string
+  json: any
+  successParser: (json: any) => T
+  errorParser: (json: any) => E
+}): Promise<Result<T, E>> => {
+  const response = await fetch(url, {
+    method: "POST",
+    credentials: "include",
+    headers: { "Content-Type": "application/vnd.api+json" },
+    body: json,
+  })
+  redirectIfUnauthorized(response.status)
+  const responseData = await response.json()
+  if (response.status === 201) {
+    return { ok: successParser(responseData) }
+  } else if (response.status === 400) {
+    return { error: errorParser(responseData) }
   }
 
-  throw new Error(`Response error: ${response.status}`)
+  return Promise.reject("fetch/parse error")
 }
 
 const apiGet = <T>({
@@ -34,4 +70,4 @@ const apiGet = <T>({
       }
     })
 
-export { apiGet }
+export { apiGet, apiPost }
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -12,7 +12,7 @@ const checkResponseStatus = (response: Response) => {
   throw new Error(`Response error: ${response.status}`)
 }
 
-const apiCall = <T>({
+const apiGet = <T>({
   url,
   parser,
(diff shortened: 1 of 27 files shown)


assets/src/disruptions/newDisruption.tsx
```diff
@@ -218,6 +225,40 @@ const AdjustmentForm = ({
...
+    ...(toDate && { endDate: toDate }),
+    adjustments,
+    daysOfWeek: dayOfWeekTimeRangesToDayOfWeeks(disruptionDaysOfWeek),
+    exceptions: Exception.fromDates(exceptionDates),
+    tripShortNames: [],
+  })
+
+  const data = disruption.toJsonApi()
+
+  return fetch("/api/disruptions", {
```
fjlanasa, line 254: Do you think there is any value in making a more generic function like `apiGet`? Not sure myself.

assets/src/disruptions/newDisruption.tsx
```diff
@@ -228,62 +269,102 @@ const NewDisruption = ({}): JSX.Element => {
...
+          }
+        })
+      }
+      setIsPreview(false)
+      setValidationErrors(errors)
+    }
+  }
+
+  if (doRedirect !== null) {
+    return <Redirect to={`/`} />
```
fjlanasa, line 311: should we redirect to detail page?

assets/src/disruptions/time.ts
```diff
@@ -139,6 +139,74 @@ const fromDaysOfWeek = (
...
+    case 2:
+      return "wednesday"
+    case 3:
+      return "thursday"
+    case 4:
+      return "friday"
+    case 5:
+      return "saturday"
+  }
+  return "sunday"
```
fjlanasa, line 186: nit: could just make this the `default` case in the switch statement

assets/src/api.ts
```diff
@@ -1,18 +1,54 @@
...
+  successParser: (json: any) => T
+  errorParser: (json: any) => E
+}): Promise<Result<T, E>> => {
+  const response = await fetch(url, {
+    method: "POST",
+    credentials: "include",
+    headers: { "Content-Type": "application/vnd.api+json" },
+    body: json,
+  })
+  redirectIfUnauthorized(response.status)
```
fjlanasa, line 40: What do you think about using an interceptor to handle this `redirectIfUnauthorized` logic? Might not be worth it since we shouldn't have too many other places we use it, and interceptors aren't built into `fetch` like libraries like `axios`, but it's not very hard to implement your own ([source](https://blog.logrocket.com/axios-or-fetch-api/)). Could hold off until if/when we start sprinkling it around to more places.
```
fetch = (originalFetch => {
  return (...arguments) => {
    const result = originalFetch.apply(this, arguments);
      return result.then(console.log('Request was sent'));
  };
})(fetch);

fetch('https://api.github.com/orgs/axios')
  .then(response => response.json())
  .then(data => {
    console.log(data) 
  });
```

assets/src/disruptions/newDisruption.tsx
losvedir: Yep, I did this. It's actually a post, so I made an `apiPost`, and changed the existing `apiCall` to `apiGet`.

assets/src/disruptions/newDisruption.tsx
losvedir: I'm not sure. I know that's conventional, but at this point the detail page is basically the same thing as the page you're on when you click the "create" button (both are rendering `<DisruptionPreview>` with only minor differences. I thought redirecting back to the index page to see it in the list would be clearer that something actually happened. A "flash" could help here, too, but I wasn't sure about hooking that all up right now.

fjlanasa: Gotcha, makes sense. If we want to add a success message at some point, I've used this before, and seems pretty easy to set up: https://github.com/fkhadra/react-toastify#demo

assets/src/disruptions/time.ts
losvedir: Thanks! I was trying to appease TypeScript, and it didn't like `case 6: return "sunday"` but I didn't realize it was smart enough to know the `default` case implied the right type.

assets/src/api.ts
losvedir: Oh, that's pretty nifty. Yeah, I'd be in favor of moving to something like that if we need it in a 3rd place.

PR #148, opened 2020-04-06
Adjustments flow to front end

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 List of adjustments flows through to new/edit pages](https://app.asana.com/0/584764604969369/1157557244099561/f)

The two commits should be reviewable independently.

#### Reviewer Checklist

### DIFF
--- a/assets/src/disruptions/disruptions.ts
+++ b/assets/src/disruptions/disruptions.ts
@@ -27,15 +27,10 @@ const indexToDayOfWeekString = (index: number): string => {
 }
 
 const modeForRoute = (route: string): TransitMode => {
-  switch (route) {
-    case "Red":
-      return TransitMode.Subway
-    case "Green-D":
-      return TransitMode.Subway
-    case "CR-Fairmount":
-      return TransitMode.CommuterRail
-    default:
-      return TransitMode.Subway
+  if (route.startsWith("CR-")) {
+    return TransitMode.CommuterRail
+  } else {
+    return TransitMode.Subway
   }
 }
 
--- a/assets/src/disruptions/newDisruption.tsx
+++ b/assets/src/disruptions/newDisruption.tsx
@@ -5,13 +5,15 @@ import Form from "react-bootstrap/Form"
 
 import { DayOfWeekTimeRanges, TimeRange } from "./time"
 import { DisruptionTimePicker } from "./disruptionTimePicker"
+import { apiCall } from "../api"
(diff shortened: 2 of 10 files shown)


assets/src/disruptions/disruptions.ts
```diff
@@ -27,15 +27,10 @@ const indexToDayOfWeekString = (index: number): string => {
...
-  switch (route) {
-    case "Red":
-      return TransitMode.Subway
-    case "Green-D":
-      return TransitMode.Subway
-    case "CR-Fairmount":
-      return TransitMode.CommuterRail
-    default:
-      return TransitMode.Subway
+  if (/^CR-/.exec(route)) {
```
losvedir, line 30: Maybe `.test(...)` if we don't actually need any of the capture results?

lemald: When I updated it to use `.test(...)` eslint actually objected and said I should use the string `.startsWith` method, so I just did that instead.

assets/src/disruptions/newDisruption.tsx
```diff
@@ -227,63 +229,77 @@ const NewDisruption = ({}): JSX.Element => {
...
-          fromDate={fromDate}
-          toDate={toDate}
-          disruptionDaysOfWeek={disruptionDaysOfWeek}
-          exceptionDates={exceptionDates}
-        />
-      ) : (
-        <>
-          <h1>Create new disruption</h1>
-          <AdjustmentForm
+  if (allAdjustments && allAdjustments !== "error") {
```
losvedir, line 253: Just an optional style thing, but I'm a fan of the "guard clause" approach to easily handle errors and small situations like that up front, rather than indenting the main body of the code further. Something like:

```js
if (allAdjustments === null) {
  return <Loading/>
}

if (allAdjustments === "error") {
  return <div>Error loading or parsing adjustments.</div>
}

return (
  <div>
    // ...
  </div>
)
```

lemald: Ah, yeah, that's definitely nicer. I'm still getting used to TypeScript and figuring out when it's smart enough to figure out, in this example, that in the bottom `return` statement `allAdjustments` won't be `null` or `"error"`.

lemald:
Created a [followup ticket](https://app.asana.com/0/881264583703207/1170122874093785/f) for the tests.

PR #155, opened 2020-04-09
feat: load disruption to front-end of edit page

lemald:
#### Summary of changes
**Asana Ticket:** Part of [🏹 API to edit disruption, frontend uses it](https://app.asana.com/0/584764604969369/1163421197919734/f)

This ended up being a fair bit of code that will be largely independent of the rest of the work on this ticket, so I decided to open a separate PR for it. It's all in one commit, unfortunately, but I don't think it's _too_ unwieldy to review.

#### Reviewer Checklist

### DIFF
--- a/assets/src/disruptions/editDisruption.tsx
+++ b/assets/src/disruptions/editDisruption.tsx
@@ -21,7 +21,7 @@ import { DisruptionTimePicker } from "./disruptionTimePicker"
 import Adjustment from "../models/adjustment"
 import Disruption from "../models/disruption"
 import Exception from "../models/exception"
-import { ModelObject, toModelObject } from "../jsonApi"
+import { JsonApiResponse, toModelObject } from "../jsonApi"
 import DayOfWeek from "../models/dayOfWeek"
 
 interface TParams {
@@ -65,11 +65,11 @@ const EditDisruption = ({
   >(null)
 
   React.useEffect(() => {
-    apiGet<ModelObject | ModelObject[] | "error">({
+    apiGet<JsonApiResponse>({
       url: "/api/disruptions/" + encodeURIComponent(match.params.id),
       parser: toModelObject,
       defaultResult: "error",
-    }).then((result: ModelObject | ModelObject[] | "error") => {
+    }).then((result: JsonApiResponse) => {
       if (result instanceof Disruption) {
         setDisruption(result)
       } else {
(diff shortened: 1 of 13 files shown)


assets/src/disruptions/editDisruption.tsx
```diff
@@ -34,68 +46,189 @@ const SaveCancelButton = ({
...
-      { hour: "1", minute: "00", period: "PM" },
-      { hour: "11", minute: "00", period: "PM" },
-    ],
-  ])
+  const [disruption, setDisruption] = React.useState<
+    Disruption | "error" | null
+  >(null)
+
+  React.useEffect(() => {
+    apiGet<ModelObject | ModelObject[] | "error">({
```
fjlanasa, line 68: optional, but maybe we can make this something like `type JSONApiReponse` since I see it all over now.

assets/src/disruptions/editDisruption.tsx
```diff
@@ -34,68 +46,189 @@ const SaveCancelButton = ({
...
+const setExceptionDatesForDisruption = (
+  disruption: Disruption,
+  setDisruption: React.Dispatch<Disruption>
+): React.Dispatch<Date[]> => {
+  return newExceptionDates => {
+    const newExceptionDatesAsTimes = newExceptionDates.map(date =>
+      date.getTime()
+    )
+    const currentExceptionDates = (disruption.exceptions
+      .map(exception => exception.excludedDate)
```
fjlanasa, line 149: Seems like this could just be `.filter().map()`, but not a big deal

assets/src/disruptions/editDisruption.tsx
lemald: Hmm, what type is that exactly? I can't seem to find it, even `git grep -i jsonapiresponse` (to catch different ways of capitalizing it) doesn't turn up any results.

assets/src/disruptions/editDisruption.tsx
lemald: Yeah, the types just make it a pain, since it's nice to be able to say after the `filter` that you're just dealing with a `Date[]`. Otherwise I'd definitely change it.

assets/src/disruptions/editDisruption.tsx
fjlanasa: Oh, sorry I was suggesting we make it and just set it equal to `ModelObject | ModelObject[] | "error"`

lemald: Ah, yeah, makes sense. Just pushed an additional commit to do just that. I think I got all of the existing places where we had `ModelObject | ModelObject[] | "error"`.

PR #160, opened 2020-04-13
Edit disruption API

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 API to edit disruption, frontend uses it](https://app.asana.com/0/584764604969369/1163421197919734/f)

Commits should be reviewable individually. ~`apiPatch` is largely the same as `apiPost`, but I wasn't sure if it would make sense to combine them into a single function and, if so, what that function's name should be.~ Update: I consolidated it to `apiSend` per @losvedir's suggestion.

#### Reviewer Checklist

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -20,26 +20,28 @@ const checkResponseStatus = (response: Response) => {
   throw new Error(`Response error: ${response.status}`)
 }
 
-const apiPost = async <T, E>({
+const apiSend = async <T, E>({
   url,
+  method,
   json,
   successParser,
   errorParser,
 }: {
   url: string
+  method: "POST" | "PATCH"
   json: any
   successParser: (json: any) => T
   errorParser: (json: any) => E
 }): Promise<Result<T, E>> => {
   const response = await fetch(url, {
-    method: "POST",
+    method,
     credentials: "include",
     headers: { "Content-Type": "application/vnd.api+json" },
     body: json,
   })
   redirectIfUnauthorized(response.status)
   const responseData = await response.json()
-  if (response.status === 201) {
+  if (response.status === 200 || response.status === 201) {
     return { ok: successParser(responseData) }
   } else if (response.status === 400) {
     return { error: errorParser(responseData) }
@@ -70,4 +72,4 @@ const apiGet = <T>({
       }
     })
 
-export { apiGet, apiPost }
+export { apiGet, apiSend }
--- a/assets/src/disruptions/editDisruption.tsx
+++ b/assets/src/disruptions/editDisruption.tsx
@@ -1,13 +1,16 @@
 import * as React from "react"
+
+import Alert from "react-bootstrap/Alert"
 import Button from "react-bootstrap/Button"
 import ButtonGroup from "react-bootstrap/ButtonGroup"
+
+import { Redirect } from "react-router-dom"
 import { RouteComponentProps, Link } from "react-router-dom"
 
-import { apiGet } from "../api"
+import { apiGet, apiSend } from "../api"
 
 import Header from "../header"
 import Loading from "../loading"
-import { DisruptionPreview } from "./disruptionPreview"
 import { DisruptionSummary } from "./disruptionSummary"
 import {
   Time,
@@ -21,7 +24,7 @@ import { DisruptionTimePicker } from "./disruptionTimePicker"
 import Adjustment from "../models/adjustment"
 import Disruption from "../models/disruption"
 import Exception from "../models/exception"
-import { JsonApiResponse, toModelObject } from "../jsonApi"
+import { JsonApiResponse, toModelObject, parseErrors } from "../jsonApi"
 import DayOfWeek from "../models/dayOfWeek"
 
 interface TParams {
@@ -30,20 +33,16 @@ interface TParams {
 
 interface SaveCancelButtonProps {
   disruptionId: string
-  setIsPreview: React.Dispatch<boolean>
+  saveFn: () => void
 }
 
 const SaveCancelButton = ({
   disruptionId,
-  setIsPreview,
+  saveFn,
 }: SaveCancelButtonProps): JSX.Element => {
   return (
     <ButtonGroup vertical>
-      <Button
-        variant="primary"
-        onClick={() => setIsPreview(true)}
-        id="save-changes-button"
-      >
+      <Button variant="primary" onClick={saveFn} id="save-changes-button">
         save changes
       </Button>
       <Link
@@ -63,6 +62,24 @@ const EditDisruption = ({
   const [disruption, setDisruption] = React.useState<
     Disruption | "error" | null
   >(null)
+  const [validationErrors, setValidationErrors] = React.useState<string[]>([])
+  const [doRedirect, setDoRedirect] = React.useState<boolean>(false)
+
+  const saveDisruption = React.useCallback(async () => {
+    const result = await apiSend({
+      url: "/api/disruptions/" + encodeURIComponent(match.params.id),
+      method: "PATCH",
+      json: JSON.stringify((disruption as Disruption).toJsonApi()),
+      successParser: toModelObject,
+      errorParser: parseErrors,
+    })
+
+    if (result.ok) {
+      setDoRedirect(true)
+    } else if (result.error) {
+      setValidationErrors(result.error)
+    }
+  }, [disruption, match])
 
   React.useEffect(() => {
     apiGet<JsonApiResponse>({
@@ -77,6 +94,24 @@ const EditDisruption = ({
       }
     })
   }, [match.params.id])
+
+  if (validationErrors.length > 0) {
+    return (
+      <Alert variant="danger">
+        <ul>
+          {validationErrors.map(err => (
+            <li key={err}>{err} </li>
+          ))}
+        </ul>
+      </Alert>
+    )
+  }
+
+  if (doRedirect) {
+    return (
+      <Redirect to={"/disruptions/" + +encodeURIComponent(match.params.id)} />
+    )
+  }
 
   if (disruption === "error") {
     return <div>Error loading disruption.</div>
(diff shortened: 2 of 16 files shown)


losvedir:
Maybe consolidate `apiPost` and `apiPatch` as... `apiSend`.. and have `method` be a required argument? I think that's basically the only difference, right? I guess the status should be `200` vs. `201` for `POST`, but probably that function can just consider either of those as success.

lemald:
Oops, well that reminds me that I forgot to send a 200 status code for `update` (rather than 201). I like your idea for `apiSend` as the name, though, I'll go with that.

assets/src/disruptions/editDisruption.tsx
```diff
@@ -133,6 +152,21 @@ const EditDisruption = ({
         new Disruption({ ...disruption }),
         setDisruption
       )}
+      saveDisruption={async () => {
```
losvedir, line 155: By defining this inline, the function will be re-created on every render, and the `EditDisruptionForm` component will have a prop there that doesn't compare equal. Maybe pull this out to a `const saveDisruption = React.useCallback(..., [disruption, match])`, like Frank has done in some places? I realize defining inline is how we've done it for the other props, but I think Frank's way is the best practice going forward.

lib/arrow/adjustment.ex
```diff
@@ -29,13 +35,15 @@ defmodule Arrow.Adjustment do
   end
 
   @doc false
+  @spec changeset(t(), map()) :: Ecto.Changeset.t()
   def changeset(adjustment, attrs) do
     adjustment
-    |> cast(attrs, [:source, :source_label, :route_id])
+    |> cast(attrs, [:id, :source, :source_label, :route_id])
```
losvedir, line 41: Is this necessary? I don't think this feature should be changing adjustments, and we have the `changeset_assoc` function which whitelists the `:id` (though now that I think about it, I think that function is unnecessary, since it was only used in the `with: ...` option that was actually ignored.

When I was updating the `Disruption.changeset` to handle creation, I had to pass adjustments separately, rather than as part of the changeset, since I wanted to query the DB to retrieve them based on the label, and also to be prepared in the future for creating new ones at the same time.

lib/arrow/adjustment.ex
```diff
@@ -20,6 +20,12 @@ defmodule Arrow.Adjustment do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @type t :: %__MODULE__{
+          route_id: String.t() | nil,
+          source: String.t() | nil,
+          source_label: String.t() | nil
```
losvedir, line 26: There are also `:inserted_at` and `:updated_at` fields from the `timestamps/1` call in `schema`.

lib/arrow/disruption.ex
```diff
@@ -11,20 +11,39 @@ defmodule Arrow.Disruption do
...
   alias Arrow.Disruption.{DayOfWeek, Exception, TripShortName}
 
+  @type t :: %__MODULE__{
+          end_date: Date.t() | nil,
+          start_date: Date.t() | nil,
+          days_of_week: [DayOfWeek.t()] | Ecto.Association.NotLoaded.t(),
+          exceptions: [Exception.t()] | Ecto.Association.NotLoaded.t(),
+          trip_short_names: [TripShortName.t()] | Ecto.Association.NotLoaded.t(),
+          adjustments: [Arrow.Adjustment.t()] | Ecto.Association.NotLoaded.t()
+        }
```
losvedir, line 21: Ditto on the timestamps.

lib/arrow/adjustment.ex
lemald: Yeah good catch. I don't think it's needed.

lib/arrow/disruption.ex
```diff
@@ -38,12 +57,33 @@ defmodule Arrow.Disruption do
...
+      for dow <- attrs["days_of_week"] || [],
+          do: DayOfWeek.changeset(%DayOfWeek{}, dow)
+
+    exceptions =
+      for exception <- attrs["exceptions"] || [],
+          do: Exception.changeset(%Exception{}, exception)
+
+    trip_short_names =
+      for name <- attrs["trip_short_names"] || [],
+          do: TripShortName.changeset(%TripShortName{}, name)
```
losvedir, line 81: Could this be a part of the shared `changeset` function? Or is it slightly different in a way I'm missing?

lib/arrow/disruption/day_of_week.ex
```diff
@@ -2,6 +2,13 @@ defmodule Arrow.Disruption.DayOfWeek do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @type t :: %__MODULE__{
+          day_name: String.t() | nil,
+          start_time: Time.t() | nil,
+          end_time: Time.t() | nil,
+          disruption: Arrow.Disruption | Ecto.Association.NotLoaded.t()
+        }
```
losvedir, line 10: Timestamps.

lib/arrow/disruption/exception.ex
```diff
@@ -2,6 +2,11 @@ defmodule Arrow.Disruption.Exception do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @type t :: %__MODULE__{
+          excluded_date: Date.t() | nil,
+          disruption: Arrow.Disruption | Ecto.Association.NotLoaded.t()
```
losvedir, line 7: Timestamps.

lib/arrow/disruption/trip_short_name.ex
```diff
@@ -2,6 +2,11 @@ defmodule Arrow.Disruption.TripShortName do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @type t :: %__MODULE__{
+          trip_short_name: String.t() | nil,
+          disruption: Arrow.Disruption | Ecto.Association.NotLoaded.t()
```
losvedir, line 7: Timestamps.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -59,6 +59,40 @@ defmodule ArrowWeb.API.DisruptionController do
...
+
+    relationships = ArrowWeb.Utilities.get_json_api_relationships(params_relationships)
+    attrs = JaSerializer.Params.to_attributes(params_data)
+
+    attrs = Map.merge(attrs, relationships)
+
+    changeset =
+      Disruption.changeset_for_update(
+        Repo.get(Disruption, id)
+        |> Repo.preload(:adjustments)
```
losvedir, line 75: Editing a disruption shouldn't allow changing its adjustments right? Maybe we shouldn't preload those, and then also not mention them in the `changeset`.

assets/src/api.ts
```diff
@@ -20,26 +20,28 @@ const checkResponseStatus = (response: Response) => {
...
-const apiPost = async <T, E>({
+const apiSend = async <T, E>({
   url,
+  method,
   json,
   successParser,
   errorParser,
 }: {
   url: string
+  method: string
```
losvedir, line 31: Maybe `"POST" | "PATCH"`?

assets/src/disruptions/editDisruption.tsx
lemald: That's mostly easy, but to avoid running afoul of the Rules of Hooks I need to put it way up above where `disruption` is possibly `null` or `"error"` and then I have to have an `if` statement that explicitly handles that and (this is the tricky part) somehow test it. I'll push up what I've got and come back to the testing issue.

losvedir: I don't quite follow. Which hook are you talking about? Why can't you put the `const saveDisruption = ...`, say, right under the `doRedirect` / `setDoRedirect` ?

lemald: That's where it is, but then `disruption` isn't necessarily a `Disruption`, it could also be `null` or `"error"`. Previously this function was just defined inline and in that case we already had `if` statements that would have returned different things if `disruption` had one of those values, so we were clear.

lib/arrow/disruption.ex
lemald: The only difference is whether the results are fed into `put_assoc` or `cast_assoc`. I could maybe just write a helper function that returns these or something.

assets/src/disruptions/editDisruption.tsx
losvedir: I see. Well, I _think_ from a runtime perspective, it's the same if you just don't include the `if` statements and assert `as Disruption`?

lib/arrow/disruption.ex
losvedir: Ah, right, I guess it doesn't pipe into the functions quite as naturally (unless you pack those values in a tuple or something). But, yeah, I think pulling it out into a little helper function might be good.

lib/arrow_web/controllers/api/disruption_controller.ex
lemald: I can remove this `preload`, it then throws an error unless I explicitly override `adjustments` in the serializer to handle the case when it's not preloaded. I'm not sure how much difference that change on its own makes. `changeset_for_update` is already free of references to adjustments.

lib/arrow/disruption.ex
lemald: Will do.

assets/src/disruptions/editDisruption.tsx
lemald: Sure, it makes me a little sad to not be using the power of TypeScript, but it ought to work.

lemald:
@losvedir okay, whew, I think I've responded to all of your comments.

lib/arrow_web/controllers/api/disruption_controller.ex
losvedir: Ah, right, you have to preload it anyway for the success response. Yeah, makes sense to keep doing it here then.

assets/src/disruptions/editDisruption.tsx
losvedir: Ah, right, because TypeScript wasn't inferring that by the time the component `render` the `useEffect` would have run, etc... it's just because it had been type narrowed by the time it was down there. So it is kind of cheating. Bummer.

PR #166, opened 2020-04-16
test: remove Enzyme, move existing tests to React Testing Library

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 Transition existing Arrow tests from Enzyme to React Testing Library](https://app.asana.com/0/584764604969369/1170122874093785/f)

This is mostly pretty rote replacement of Enzyme with React Testing Library. The one bit I'm still not super happy with is the test `selecting a day of the week enables updating time range`. It used to do a lot of asserting based on the props of the component to make sure it was doing the right thing with the day of week time ranges. However, React Testing Library is a lot more opinionated about not really letting you pry into the contents of React components and instead preferring you to strictly test the front-end. I'm open to suggestions on how to re-do that test in a nicer way if anyone has any.

#### Reviewer Checklist

### DIFF
--- a/assets/tests/disruptions/disruptionTimePicker.test.tsx
+++ b/assets/tests/disruptions/disruptionTimePicker.test.tsx
@@ -1,4 +1,4 @@
-import { mount } from "enzyme"
+import { render, fireEvent } from "@testing-library/react"
 import * as React from "react"
 
 import { DayOfWeekTimeRanges } from "../../src/disruptions/time"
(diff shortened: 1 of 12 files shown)


assets/tests/disruptions/disruptionTimePicker.test.tsx
```diff
@@ -28,334 +28,347 @@ const DisruptionTimePickerWithProps = ({}): JSX.Element => {
...
+
+    const mondayEndPeriodSelect = container.querySelector(
+      "select#time-of-day-end-period-0"
+    )
+    if (!mondayEndPeriodSelect) {
+      throw new Error("Monday end period select not found")
+    }
+    fireEvent.change(mondayEndPeriodSelect, { target: { value: "PM" } })
+
+    fireEvent.click(mondayEndTypeCheck)
```
fjlanasa, line 108: Are these last two click events necessary? Seems like there should be an assertion afterwards, or else we can remove.

lemald: Without these last couple of clicks there's a line or two that goes uncovered. This is the test that my note in the PR description is referring to, previously there were some more assertions in the test but there isn't really a good way to directly test the value of the props anymore, at least that I'm aware of.

fjlanasa: Could you just assert on the state of the input? Like I assume clicking `mondayEndTypeCheck` or `mondayCheck` at that point disables and/or changes the value of some inputs.

lemald: Yeah, it should disable the dropdowns. That's probably the most sensible thing to assert.

PR #168, opened 2020-04-17
feat: Don't allow changing past data

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 make sure edits don't change past disruptions](https://app.asana.com/0/584764604969369/1154489975402115)

The ticket ended up being trickier than I anticipated. The goal of the ticket is to preserve the historical record; not allowing any data in the past to change. The asana task ended up being underspecified for this goal. These were the scenarios I brainstormed and handled:

1. Cannot create Disruption with start date in the past
1. Cannot edit Disruption start date in the past to any value
1. Cannot edit Disruption start date in the future to the past
1. Cannot create Disruption with end date in the past
1. Cannot edit Disruption end date in the past to any value
1. Cannot edit Disruption end date in the future to the past
1. Cannot create Exception date in the past
1. Cannot edit Exception date from the future to the past
1. Cannot edit Exception date from the past to any time
1. Cannot delete Exception date from the past
1. Cannot change Adjustment if disruption start date in the past

For values that end up in the past, I wrote a custom validation `validate_not_in_past`. This runs whenever a changeset's value changes, so it's easy to identify that a value has changed (or been created in the past).

To handle values that change _from_ the past to the future, I wrote another custom validation `validate_not_changing_past` that looks at the previous data value and the new value.

Checking for deletions of exceptions required a custom validation function on `Disruption`.

We don't allow editing Adjustments at all, so that was taken care of by existing behavior.

#### Reviewer Checklist


lib/arrow/disruption.ex
```diff
@@ -36,20 +36,35 @@ defmodule Arrow.Disruption do
   end
 
   @doc false
-  @spec changeset(t(), map()) :: Ecto.Changeset.t()
-  def changeset(disruption, attrs) do
+  @spec changeset(t(), map(), Date.t()) :: Ecto.Changeset.t()
+  def changeset(disruption, attrs, today) do
```
lemald, line 40: This is something I should have considered in my PR for the edit functionality, but can we maybe make `changeset/3` a private function if nothing outside this module calls it directly anymore?

lib/arrow/disruption.ex
```diff
@@ -58,32 +73,34 @@ defmodule Arrow.Disruption do
...
 
     disruption
-    |> changeset(attrs)
-    |> cast_assoc(:days_of_week, days_of_week)
-    |> cast_assoc(:exceptions, exceptions)
-    |> cast_assoc(:trip_short_names, trip_short_names)
+    |> changeset(attrs, today)
+    |> Arrow.Validations.validate_not_changing_past(:start_date, today)
+    |> Arrow.Validations.validate_not_changing_past(:end_date, today)
+    |> cast_assoc(:days_of_week)
```
losvedir, line 84: Just a quick note about what's going on here. The way `cast_assoc` seems to work is akin to `Changeset.cast` - it takes a params map, and casts that data into a changeset. 

In this case, though, rather than passing in the params map as an argument, it gets the raw data out of the changeset, where it was evidently stored earlier in the `cast(disruption, attrs, [:field1, :field2])` call. That is, that `cast` call filters the params for `:field1` and `:field2` for the purposes of the *disruption* changeset, but still stores the full map somewhere in the changeset.

Since those params looked like `%{"field1" => ..., "field2" => ..., "days_of_week" => [...]}`, the `cast_assoc(changeset, :days_of_week)` calls looks in the params map at the `"days_of_week"` key.

So, basically, before when we were doing `cast_assoc(changeset, :days_of_week, days_of_week)` the `days_of_week` argument was being treated as an options keyword list and not doing anything.

That means, the `assoc_changeset` function is actually only needed in `changeset_for_create`, where we use `put_assoc`, which _doesn't_ do casting, and just puts the data right in for the associations. So I reverted that change I asked you to do, deleting that function, and just moving it back into `changeset_for_create`.

lemald: Yeah, I went back and read over the Ecto docs on `cast_assoc/3` and it started to make much more sense.

lemald:
One other thing that wasn't specified in the ticket: do we want to provide any real-time front-end validation of the form as the user is editing it? Obviously we still want to validate on the backend but it might be a nicer user experience if you see some messaging immediately when you try to make an invalid change. If so that's its whole own can of worms and should probably be a separate ticket though.

losvedir:
Yeah, that would be nice. I'll make a ticket so we can discuss it as a team.

PR #180, opened 2020-04-24
feat: load secrets from AWS Secrets Manager

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 Retrieve Arrow secrets from Secrets Manager](https://app.asana.com/0/584764604969369/1162039690437500/f)

Largely just a copy of the code from Delta but modified for our particular secrets. I opted not to remove the code that pulls these values from environment variables to make life easier if we want to do local testing that involves changing them, and I don't think that will cause any problems in the EBS environments, but we'll see.

#### Reviewer Checklist


lemald:
@paulswartz yup, I'm in the process of deploying it to dev, and I'll talk to Ian about removing the old environment variables and making sure everything still works before merging.

losvedir:
Is `SECRET_KEY_BASE` in SecretsManager? If so, I think we need to remove the code in the `init/2` callback in `arrow_web/endpoint.ex`.

lemald:
@losvedir it is, but Delta uses the same approach and keep that code in its corresponding `init/2` function.

losvedir:
I guess it doesn't overwrite it if it's empty, so that's fine.

PR #246, opened 2020-05-28
feat: disruption validations

fjlanasa:
#### Summary of changes
**Asana Ticket:** [🏹 Arrow has validation for disruptions](https://app.asana.com/0/584764604969369/1173526948047133/f)

Validations added:
* start date shouldn't be after end date
* days of week should fall between start and end date
* any exception dates should fall between start and end date
* exception dates should be unique, and should be applicable to days of week 
* start time happens before end time 

Screen Shot 2020-06-02 at 11 05 06 PM

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -69,7 +69,6 @@ defmodule Arrow.Disruption do
     today = DateTime.to_date(current_time)
 
     disruption
-    |> Arrow.Repo.preload([:days_of_week, :exceptions, :trip_short_names])
     |> changeset(attrs, today)
     |> Arrow.Validations.validate_not_changing_past(:start_date, today)
     |> Arrow.Validations.validate_not_changing_past(:end_date, today)
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -41,10 +41,26 @@ defmodule Arrow.Disruption do
   def changeset_for_create(disruption, attrs, adjustments, current_time) do
     today = DateTime.to_date(current_time)
 
+    days_of_week =
+      for dow <- attrs["days_of_week"] || [],
+          do: DayOfWeek.changeset(%DayOfWeek{}, dow)
+
+    exceptions =
+      for exception <- attrs["exceptions"] || [],
+          do: Exception.changeset(%Exception{}, exception, today)
+
+    trip_short_names =
+      for name <- attrs["trip_short_names"] || [],
+          do: TripShortName.changeset(%TripShortName{}, name)
+
     disruption
     |> changeset(attrs, today)
     |> put_assoc(:adjustments, adjustments)
     |> validate_length(:adjustments, min: 1)
+    |> put_assoc(:days_of_week, days_of_week)
+    |> put_assoc(:exceptions, exceptions)
+    |> put_assoc(:trip_short_names, trip_short_names)
+    |> common_validations()
   end
 
   @doc false
@@ -57,9 +73,13 @@ defmodule Arrow.Disruption do
     |> changeset(attrs, today)
     |> Arrow.Validations.validate_not_changing_past(:start_date, today)
     |> Arrow.Validations.validate_not_changing_past(:end_date, today)
+    |> cast_assoc(:days_of_week)
+    |> cast_assoc(:exceptions, with: {Exception, :changeset, [today]})
+    |> cast_assoc(:trip_short_names)
     |> validate_not_deleting_past_exception(today)
     |> validate_not_changing_relationship_in_past(:days_of_week, today)
     |> validate_not_changing_relationship_in_past(:trip_short_names, today)
+    |> common_validations()
   end
 
   @doc false
@@ -68,11 +88,13 @@ defmodule Arrow.Disruption do
     disruption
     |> cast(attrs, [:id, :start_date, :end_date])
     |> validate_required([:start_date, :end_date])
-    |> cast_assoc(:days_of_week)
-    |> cast_assoc(:exceptions, with: {Exception, :changeset, [today]})
-    |> cast_assoc(:trip_short_names)
     |> Arrow.Validations.validate_not_in_past(:start_date, today)
     |> Arrow.Validations.validate_not_in_past(:end_date, today)
+  end
+
+  @spec common_validations(Ecto.Changeset.t()) :: Ecto.Changeset.t(t())
+  defp common_validations(changeset) do
+    changeset
     |> validate_start_date_before_end_date()
     |> validate_days_of_week_between_start_and_end_date()
     |> validate_exceptions_between_start_and_end_date()
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -136,7 +136,7 @@ defmodule Arrow.Disruption do
   defp validate_days_of_week_between_start_and_end_date(changeset) do
     start_date = get_field(changeset, :start_date)
     end_date = get_field(changeset, :end_date)
-    days_of_week = get_field(changeset, :days_of_week)
+    days_of_week = get_field(changeset, :days_of_week, [])
 
     cond do
       is_nil(start_date) or is_nil(end_date) ->
@@ -160,7 +160,7 @@ defmodule Arrow.Disruption do
 
   @spec validate_exceptions_are_unique(Ecto.Changeset.t(t())) :: Ecto.Changeset.t(t())
   defp validate_exceptions_are_unique(changeset) do
-    exceptions = get_field(changeset, :exceptions)
+    exceptions = get_field(changeset, :exceptions, [])
 
     if Enum.uniq_by(exceptions, fn %{excluded_date: excluded_date} -> excluded_date end) ==
          exceptions do
@@ -175,7 +175,7 @@ defmodule Arrow.Disruption do
   defp validate_exceptions_between_start_and_end_date(changeset) do
     start_date = get_field(changeset, :start_date)
     end_date = get_field(changeset, :end_date)
-    exceptions = get_field(changeset, :exceptions)
+    exceptions = get_field(changeset, :exceptions, [])
 
     if Enum.all?(exceptions, fn exception ->
          Enum.member?([:lt, :eq], Date.compare(start_date, exception.excluded_date)) and
@@ -190,8 +190,8 @@ defmodule Arrow.Disruption do
   @spec validate_exceptions_are_applicable(Ecto.Changeset.t(t())) ::
           Ecto.Changeset.t(t())
   defp validate_exceptions_are_applicable(changeset) do
-    days_of_week = get_field(changeset, :days_of_week)
-    exceptions = get_field(changeset, :exceptions)
+    days_of_week = get_field(changeset, :days_of_week, [])
+    exceptions = get_field(changeset, :exceptions, [])
 
     day_of_week_numbers = Enum.map(days_of_week, fn x -> DayOfWeek.day_number(x) end)
 
@@ -206,7 +206,7 @@ defmodule Arrow.Disruption do
 
   @spec validate_start_time_before_end_time(Ecto.Changeset.t(t())) :: Ecto.Changeset.t(t())
   defp validate_start_time_before_end_time(changeset) do
-    days_of_week = get_field(changeset, :days_of_week)
+    days_of_week = get_field(changeset, :days_of_week, [])
 
     if Enum.any?(days_of_week, fn day ->
          not (is_nil(day.start_time) or is_nil(day.end_time)) and
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -177,13 +177,18 @@ defmodule Arrow.Disruption do
     end_date = get_field(changeset, :end_date)
     exceptions = get_field(changeset, :exceptions, [])
 
-    if Enum.all?(exceptions, fn exception ->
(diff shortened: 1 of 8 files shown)


lib/arrow/disruption.ex
```diff
@@ -121,4 +127,108 @@ defmodule Arrow.Disruption do
...
+    if Enum.all?(exceptions, fn exception ->
+         Enum.member?(day_of_week_numbers, Date.day_of_week(exception.excluded_date))
+       end) do
+      changeset
+    else
+      add_error(changeset, :exceptions, "should be applicable to days of week")
+    end
+  end
+
+  @spec validate_start_time_before_end_time(Ecto.Changeset.t(t())) :: Ecto.Changeset.t(t())
```
fjlanasa, line 221: i tried adding this to the `DayOfWeek` module (since it doesn't rely on anything data related to the `Disruption`. The validation itself worked but wouldn't bubble up to be returned by the `Disruption.changeset` function.

lib/arrow/disruption.ex
```diff
@@ -71,9 +71,6 @@ defmodule Arrow.Disruption do
     |> changeset(attrs, today)
     |> Arrow.Validations.validate_not_changing_past(:start_date, today)
     |> Arrow.Validations.validate_not_changing_past(:end_date, today)
-    |> cast_assoc(:days_of_week)
```
losvedir, line 74: What was the reason for moving these? I'm a little surprised this works. 

The reason I put `cast_assoc` in `changeset_for_update` was because (I think/thought) it required that existing values for them be loaded from the database, which implies the Disruption has to have been already created. That's why `changeset_for_create` used `put_assoc` instead.

By moving these three lines to `changeset` it now applies to both of those, and `changeset_for_create` actually calls _both_ `cast_assoc` and `put_assoc`.

fjlanasa: Without the `cast_assoc`, the changes are not added to the `Disruption` changeset. I would have assumed `put_assoc` would do that but something is not working correctly.

lib/arrow/disruption.ex
```diff
@@ -121,4 +127,108 @@ defmodule Arrow.Disruption do
...
+      end) ->
+        changeset
+
+      true ->
+        add_error(changeset, :days_of_week, "should fall between start and end dates")
+    end
+  end
+
+  @spec validate_exceptions_are_unique(Ecto.Changeset.t(t())) :: Ecto.Changeset.t(t())
+  defp validate_exceptions_are_unique(changeset) do
```
losvedir, line 176: Hmm, I was going to say I wonder if we can do a validation at the DB level, and then use `unique_constraint` Ecto check here. But I wonder how we'll handle multiple Disruption versions, so maybe we don't want to go down that route right now.

lib/arrow/disruption.ex
```diff
@@ -121,4 +127,108 @@ defmodule Arrow.Disruption do
...
+  end
+
+  @spec validate_exceptions_between_start_and_end_date(Ecto.Changeset.t(t())) ::
+          Ecto.Changeset.t(t())
+  defp validate_exceptions_between_start_and_end_date(changeset) do
+    start_date = get_field(changeset, :start_date)
+    end_date = get_field(changeset, :end_date)
+    exceptions = get_field(changeset, :exceptions)
+
+    if Enum.all?(exceptions, fn exception ->
```
losvedir, line 194: Per the others, should this be a `cond` that enforces `start_date` and `end_date` are not `nil` here?

lib/arrow/disruption.ex
```diff
@@ -121,4 +127,108 @@ defmodule Arrow.Disruption do
...
+       end) do
+      changeset
+    else
+      add_error(changeset, :exceptions, "should be applicable to days of week")
+    end
+  end
+
+  @spec validate_start_time_before_end_time(Ecto.Changeset.t(t())) :: Ecto.Changeset.t(t())
+  defp validate_start_time_before_end_time(changeset) do
+    days_of_week = get_field(changeset, :days_of_week)
```
losvedir, line 223: `get_field` has an optional default 3rd parameter. Should we provide `[]` as default? Or do you know if it will do that anyway because it's a `has_many` relation?

lib/arrow/disruption.ex
```diff
@@ -121,4 +127,108 @@ defmodule Arrow.Disruption do
       changeset
     end
   end
+
+  @spec validate_start_date_before_end_date(Ecto.Changeset.t(t())) :: Ecto.Changeset.t(t())
+  defp validate_start_date_before_end_date(changeset) do
+    start_date = get_field(changeset, :start_date)
+    end_date = get_field(changeset, :end_date)
+
+    cond do
```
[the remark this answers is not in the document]
losvedir: Now that I've seen the other validations, I think `cond` is fine here, since that's a good general approach that works with all of them.

lib/arrow/disruption.ex
losvedir: Just an idea, maybe it's related to having both `put_assoc` and `cast_assoc` run now on `changeset_for_create`? Maybe one is overriding the other or something. I don't know if you were testing on creating or updating.

lib/arrow/disruption.ex
fjlanasa: of course the changes are not added to the `Disruption` changeset as is, because we run that function before the `put_assoc` statements 🤦

lib/arrow/disruption.ex
fjlanasa: I've tried all combinations (only `put`, only `cast` and both. The changeset of the associated struct (including any validation `errors`) is included in the `changes` field on the `Disruption` changeset, but not in the `errors` field on the `Disruption` changeset.

lib/arrow/disruption.ex
fjlanasa: Yeah I was looking into that, but wasn't sure (also not sure if the error would show up in the `Disruption` changeset like the others).

lib/arrow/disruption.ex
fjlanasa: I changed to just using `cast_assoc` for both. Take a look and let me know if that makes sense.

fjlanasa: Still doing a little reading trying to figure out why `put_assoc` changes aren't being added to the `Disruption` changeset, though (even when the functions were in the right order). My current hunch is that Ecto does something funky when passed structs or changesets to "ignore" the previous value:
https://github.com/elixir-ecto/ecto/blob/890cc735b885bf629f271e2b631fdc49ba1276e1/lib/ecto/changeset.ex#L1269

fjlanasa:
@losvedir some notes following our conversation this morning:
* I was able to get the tests working with `put_assoc` in `changeset_for_create` by pulling out the validations from `changeset` into a seperate function `common_validations` run at the end of both `changeset_for_create` and `changeset_for_update`. I swore I tried this before, but 🤷‍♂️ . I kind of still think it makes more sense to use `cast_assoc` (we wouldn't have to manually create the changesets for the associated entities), but I'll let you look at it and weigh in. 
* I then tried adding the `validate_start_time_before_end_time` into the `DayOfWeek.changeset`. The good news was that the validation did work, in that both the `DayOfWeek` and `Disruption` changesets were marked as invalid. The bad news is that because the errors are nested (pasting an example changeset below), they aren't being serialized correctly, so they are not being displayed in the UI. Not totally sure what lever to pull to fix this, but can dig into it further if we want to make that work:

```
#Ecto.Changeset<
  action: nil,
  changes: %{
    adjustments: [
      #Ecto.Changeset<action: :update, changes: %{}, errors: [],
       data: #Arrow.Adjustment<>, valid?: true>
    ],
    days_of_week: [
      #Ecto.Changeset<
        action: :insert,
        changes: %{
          day_name: "friday",
          end_time: ~T[19:30:00],
          start_time: ~T[20:30:00]
        },
        errors: [days_of_week: {"start_time should be before end_time", []}],
        data: #Arrow.Disruption.DayOfWeek<>,
        valid?: false
      >,
      #Ecto.Changeset<
        action: :insert,
        changes: %{day_name: "saturday"},
        errors: [],
        data: #Arrow.Disruption.DayOfWeek<>,
        valid?: true
      >
    ],
    end_date: ~D[2019-12-12],
    exceptions: [],
    start_date: ~D[2019-10-10],
    trip_short_names: []
  },
  errors: [],
  data: #Arrow.Disruption<>,
  valid?: false
>
```

lib/arrow/disruption.ex
```diff
@@ -68,6 +69,7 @@ defmodule Arrow.Disruption do
     today = DateTime.to_date(current_time)
 
     disruption
+    |> Arrow.Repo.preload([:days_of_week, :exceptions, :trip_short_names])
```
losvedir, line 72: I think we've been trying to keep DB interactions in the controller, with the idea that changesets are pure functions.

losvedir:
I like the way it is now with the shared validations function. Looks like `traverse_errors` is the way to go here.

fjlanasa:
> I like the way it is now with the shared validations function. Looks like `traverse_errors` is the way to go here.

Do you know where is the appropriate place to use `traverse_errors`?

lib/arrow/disruption.ex
fjlanasa: I think we can actually remove this entirely now that I've switched back to using `put_assoc`

losvedir:
> Do you know where is the appropriate place to use `traverse_errors`?

I was just looking at that. `ja_serializer` does some magic there. Looking at the https://github.com/vt-elixir/ja_serializer/blob/master/test/ja_serializer/error_serializer_test.exs, I think we might be able to change the `data: changeset` to `data: [%{detail: ...}, %{detail: ...}]` which might be able to be easily constructed with `traverse_errors`. I don't think we need to provide _all_ the JSON API fields that are possible, since the frontend only uses `detail`, IIRC.

fjlanasa:
> > Do you know where is the appropriate place to use `traverse_errors`?
> 
> I was just looking at that. `ja_serializer` does some magic there. Looking at the https://github.com/vt-elixir/ja_serializer/blob/master/test/ja_serializer/error_serializer_test.exs, I think we might be able to change the `data: changeset` to `data: [%{detail: ...}, %{detail: ...}]` which might be able to be easily constructed with `traverse_errors`. I don't think we need to provide _all_ the JSON API fields that are possible, since the frontend only uses `detail`, IIRC.

I'm taking a look at this, and I'm not sure changing to use `traverse_errors` is worth it. I'd have to reimplement the logic ja_serializer uses to parse the errors from something like:
```
%{
  adjustments: [
    {"should have at least %{count} item(s)",
     [count: 1, validation: :length, kind: :min, type: :list]}
  ],
  end_date: [{"can't be blank", [validation: :required]}],
  start_date: [{"can't be blank", [validation: :required]}]
}
```
to
```
[{
		"detail": "Adjustments should have at least 1 item(s)",
		"source": {
			"pointer": "/data/attributes/adjustments"
		},
		"title": "should have at least 1 item(s)"
	}, {
		"detail": "Start date can't be blank",
		"source": {
			"pointer": "/data/attributes/start_date"
		},
		"title": "can't be blank"
	}, {
		"detail": "End date can't be blank",
		"source": {
			"pointer": "/data/attributes/end_date"
		},
		"title": "can't be blank"
	}]
```

losvedir:
I had a thought in the shower today: Maybe `traverse_errors` in tandem with [`JaSerializer.EctoErrorSerializer.format/2`](https://github.com/mbta/ja_serializer/blob/master/lib/ja_serializer/ecto_error_serializer.ex) could be pretty straightforward?

fjlanasa:
> I had a thought in the shower today: Maybe `traverse_errors` in tandem with [`JaSerializer.EctoErrorSerializer.format/2`](https://github.com/mbta/ja_serializer/blob/master/lib/ja_serializer/ecto_error_serializer.ex) could be pretty straightforward?

I'll give it a look, but have you looked at the last couple of commits which implements your desired behavior? I'm fairly confident `format` and `traverse_errors` won't solve the issue of having to return both errors on the `Disruption` changeset _and_ `DayOfWeek` changeset.

fjlanasa:
> > I had a thought in the shower today: Maybe `traverse_errors` in tandem with [`JaSerializer.EctoErrorSerializer.format/2`](https://github.com/mbta/ja_serializer/blob/master/lib/ja_serializer/ecto_error_serializer.ex) could be pretty straightforward?
> 
> I'll give it a look, but have you looked at the last couple of commits which implements your desired behavior? I'm fairly confident `format` and `traverse_errors` won't solve the issue of having to return both errors on the `Disruption` changeset _and_ `DayOfWeek` changeset.

They don't work well together. I tried `Ecto.Changeset.traverse_errors(changeset, fn x -> x end) |> JaSerializer.EctoErrorSerializer.format()` and the former doesn't return the errors in the correct format.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -91,10 +91,43 @@ defmodule ArrowWeb.API.DisruptionController do
       {:error, changeset} ->
         conn
         |> put_status(400)
-        |> render(:errors, data: changeset)
+        |> render(:errors, errors: format_errors(changeset))
     end
   end
 
+  @spec format_field_name(atom()) :: String.t()
```
losvedir, line 98: I think all these helper functions are fine and look good, but they're not especially `DisruptionController` specific. Maybe put them in `ArrowWeb.Utilities` with our other JSON API serialization helper? I think only `format_errors` needs to be public, and then you can have a test of just the data munging by itself, without dealing with a controller or request/response cycle.

PR #332, opened 2020-07-14
chore: migration to create auth_tokens table

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 figure out authentication for existing arrow API](https://app.asana.com/0/584764604969369/1173297500981963/f)

Putting the migration in a separate PR so we can deploy it first, _then_ the code that actually uses the table.

#### Reviewer Checklist

### DIFF
--- a/priv/repo/migrations/20200713155611_create_auth_tokens.exs
+++ b/priv/repo/migrations/20200713155611_create_auth_tokens.exs
@@ -1,0 +1,14 @@
+defmodule Arrow.Repo.Migrations.CreateAuthTokens do
+  use Ecto.Migration
+
+  def change do
+    create table(:auth_tokens) do
+      add :username, :text, null: false
+      add :token, :text, null: false
+    end
+
+    create unique_index(:auth_tokens, [:username])
+
+    create index(:auth_tokens, [:token])
+  end
+end
--- a/priv/repo/structure.sql
+++ b/priv/repo/structure.sql
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 10.5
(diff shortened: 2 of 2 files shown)


priv/repo/migrations/20200713155611_create_auth_tokens.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Arrow.Repo.Migrations.CreateAuthTokens do
+  use Ecto.Migration
+
+  def change do
+    create table(:auth_tokens) do
+      add :username, :string, null: false
```
losvedir, line 6: I _think_ `:string` maps to `VARCHAR(255)`, while `:text` is preferred for postgres. Performance wise there's no difference, but `text` is more flexible (and can have a length constraint added or changed later.

losvedir: Same for `:token`.

lemald: Ah, after fiddling around a bit with `mix ecto.migrate --log-sql` I think you're right.

losvedir:
Sorry, just remembered something and wanted to get that comment in, to prevent the merge. Can you also update the `priv/repo/structure.sql` file, via `mix ecto.dump`?

lemald:
@losvedir Good catch! Just pushed the update.

digitalcora:
> Can you also update the `priv/repo/structure.sql` file, via `mix ecto.dump`?

We can make this happen automatically by creating appropriate aliases for `ecto.migrate` and `ecto.rollback`. [Example of this on alerts_concierge](https://github.com/mbta/alerts_concierge/blob/master/mix.exs#L38).

PR #333, opened 2020-07-15
Allow API authentication via token

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 figure out authentication for existing arrow API](https://app.asana.com/0/584764604969369/1173297500981963/f)

Should be reviewable commit-by-commit.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/router.ex
+++ b/lib/arrow_web/router.ex
@@ -46,6 +46,7 @@ defmodule ArrowWeb.Router do
     get "/disruptions/new", PageController, :index
     get "/disruptions/:id", PageController, :index
     get "/disruptions/:id/edit", PageController, :index
+    get "/mytoken", MyTokenController, :show
   end
 
   scope "/", ArrowWeb do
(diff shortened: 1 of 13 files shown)


lib/arrow_web/router.ex
```diff
@@ -46,6 +50,7 @@ defmodule ArrowWeb.Router do
     get "/disruptions/new", PageController, :index
     get "/disruptions/:id", PageController, :index
     get "/disruptions/:id/edit", PageController, :index
+    get "/mytoken", MyTokenController, :index
```
losvedir, line 53: I see where you're coming from calling this `:index` since there's no ID, but I think in Phoenix (and Rails before it) convention, this would be considered more of a ["singleton" `:show` path](https://hexdocs.pm/phoenix/Phoenix.Router.html#resources/4-singleton-resources).

lemald: So, this is weird... I pushed a `fixup!` commit, and if you go to the branch in the main GitHub code view it shows up, but it doesn't show up in the PR for some reason... 🤔

lemald: Well I tried squashing down the `fixup!` commit and force-pushing and the changes _still_ aren't reflected here in the PR.

losvedir:
Yep, that sounds like a good plan! Good catch about two steps to deploy Arrow-prod. I'll do the first one now.

PR #343, opened 2020-07-21
feat: add disruptions calendar view

fjlanasa:
#### Summary of changes
**Asana Ticket:** [🏹 Arrow has a calendar view](https://app.asana.com/0/584764604969369/1154489975402109/f)

Changes:
* Added [`@fullcalendar`](https://fullcalendar.io/) (calendar library) and [`rrule`](https://github.com/jakubroztocil/rrule) (a helper library for recurrence rules)
* Because jest doesn't handle our ES6 modules and `@fullcalendar` transpile to ES5, had to edit our tsconfig to transpile that depedency
* Skip loading source map for `rrule`
* Added component `DisruptionCalendar`, which takes a list of `Disruption` objects.
* Refactored `DisruptionIndex` and `DisruptionTable` to let `DisruptionTable` simply take a list of `Disruption` objects and handle the parsing it needs to do itself

Note:
As an alternative to `@fullcalendar` (and it's companion packages) and `rrule`, I looked into using `react-big-calendar`, `moment` and `moment-recur`, but I the bundle size was about the same and the `@fullcalendar` code was already written by the time I found the issue with jest. Could probably switch if people feel strongly though.

filters

#### Reviewer Checklist


losvedir:
Can you talk a little bit about the snapshot testing? How does it work, and what's the process going forward for creating new ones / keeping that one up to date, etc?

losvedir:
Trying locally, ran into this error:

```
disruptionCalendar.tsx:92 Uncaught TypeError: Cannot read property 'getTime' of undefined
    at disruptionCalendar.tsx:92
    at Array.forEach (<anonymous>)
    at disruptionCalendar.tsx:84
    at Array.forEach (<anonymous>)
    at disruptionCalendar.tsx:58
    at Array.reduce (<anonymous>)
    at Object../src/disruptions/disruptionCalendar.tsx.exports.disruptionsToCalendarEvents (disruptionCalendar.tsx:57)
    at disruptionCalendar.tsx:105
    at mountMemo (react-dom.development.js:15442)
    at Object.useMemo (react-dom.development.js:15738)
```

fjlanasa:
> Can you talk a little bit about the snapshot testing? How does it work, and what's the process going forward for creating new ones / keeping that one up to date, etc?



> Can you talk a little bit about the snapshot testing? How does it work, and what's the process going forward for creating new ones / keeping that one up to date, etc?

It just saves the DOM tree that would be rendered by your component given the props provided. I figure it accomplishes the same thing as querying the DOM without me having to dig into what all the CSS selectors `@fullcalendar` adds to individual elements. If you need to update the test you can just run `jest --updateSnapshot`. The snapshot files are human readable, so it's usually pretty easy to tell what changed and why.

https://jestjs.io/docs/en/snapshot-testing

fjlanasa:
> Trying locally, ran into this error:
> 
> ```
> disruptionCalendar.tsx:92 Uncaught TypeError: Cannot read property 'getTime' of undefined
>     at disruptionCalendar.tsx:92
>     at Array.forEach (<anonymous>)
>     at disruptionCalendar.tsx:84
>     at Array.forEach (<anonymous>)
>     at disruptionCalendar.tsx:58
>     at Array.reduce (<anonymous>)
>     at Object../src/disruptions/disruptionCalendar.tsx.exports.disruptionsToCalendarEvents (disruptionCalendar.tsx:57)
>     at disruptionCalendar.tsx:105
>     at mountMemo (react-dom.development.js:15442)
>     at Object.useMemo (react-dom.development.js:15738)
> ```

Do you have disruptions saved with no days of week?

fjlanasa:
> > Can you talk a little bit about the snapshot testing? How does it work, and what's the process going forward for creating new ones / keeping that one up to date, etc?
> 
> > Can you talk a little bit about the snapshot testing? How does it work, and what's the process going forward for creating new ones / keeping that one up to date, etc?
> 
> It just saves the DOM tree that would be rendered by your component given the props provided. I figure it accomplishes the same thing as querying the DOM without me having to dig into what exactly are the CSS selectors `@fullcalendar` adds to individual elements. If you need to update the test you can just run `jest --updateSnapshot`. The snapshot files are human readable, so it's usually pretty easy to tell what changed and why.
> 
> https://jestjs.io/docs/en/snapshot-testing

Additionally, I added a test on the parsing logic in that component, so I already ha

> > Can you talk a little bit about the snapshot testing? How does it work, and what's the process going forward for creating new ones / keeping that one up to date, etc?
> 
> > Can you talk a little bit about the snapshot testing? How does it work, and what's the process going forward for creating new ones / keeping that one up to date, etc?
> 
> It just saves the DOM tree that would be rendered by your component given the props provided. I figure it accomplishes the same thing as querying the DOM without me having to dig into what exactly are the CSS selectors `@fullcalendar` adds to individual elements. If you need to update the test you can just run `jest --updateSnapshot`. The snapshot files are human readable, so it's usually pretty easy to tell what changed and why.
> 
> https://jestjs.io/docs/en/snapshot-testing

Creating a new one just requires adding a new test and running the suite. It is generated if not found.

losvedir:
> Do you have disruptions saved with no days of week?

No. These are my disruptions:

Screen Shot 2020-07-22 at 3 05 25 PM

The first two do not have any date exceptions. The last one has a date exception on `6/17/2020`.

fjlanasa:
> > Do you have disruptions saved with no days of week?
> 
> No. These are my disruptions:
> 
> Screen Shot 2020-07-22 at 3 05 25 PM
> 
> The first two do not have any date exceptions. The last one has a date exception on `6/17/2020`.

I'll see if I can replicate without dealing with making disruptions in the past

assets/src/disruptions/disruptionCalendar.tsx
```diff
@@ -1,5 +1,130 @@
...
 
-const DisruptionCalendar = () => <div>calendar goes here</div>
+interface DisruptionCalendarProps {
+  disruptions: Disruption[]
+  initialDate?: Date
+  timeZone?: string
+}
 
-export default DisruptionCalendar
+export const dayNameToInt = (day?: string): number => {
```
losvedir, line 14: Why is the type `day?: string` here? Are we expecting this function to be called without an argument? In that case is `6` the right return value for that?

fjlanasa: I don't actually expect to call it without an argument. We just have marked almost every field on all models as optional so it was easy enough to keep that up here.

fjlanasa: but if you prefer i can move handling an undefined `dayName` earlier in the chain

losvedir: Yeah, I think I like it better up the chain, keeping this helper function's types semantically correct.

I also notice that this is basically the same as `dayNameToIx`. What do you think about renaming that function and using it instead? On the other hand, in some sense it's sort of "coincidentally" the same mapping between day names and numbers, so I can see an argument for having them separate.

fjlanasa: actually i'm just going to make `dayName` required on the model. seems like it won't break anything and will make this function make more sense.

losvedir: Great, if we can tighten up the model, all the better!

PR #344, opened 2020-07-21
Delete disruption

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 Disruption can be deleted in Arrow](https://app.asana.com/0/584764604969369/1170294334212195/f)

The two commits should be reviewable separately. We only allow users to delete disruptions with a start date in the past, and this is enforced both on the front-end (the button isn't displayed for disruptions that start in the past) and the back-end (it will trigger a 400 error if you try to delete a disruption that began in the past).

#### Reviewer Checklist

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -28,7 +28,7 @@ const apiSend = async <T, E>({
   errorParser,
 }: {
   url: string
-  method: "POST" | "PATCH"
+  method: "POST" | "PATCH" | "DELETE"
   json: any
   successParser: (json: any) => T
   errorParser: (json: any) => E
@@ -40,10 +40,15 @@ const apiSend = async <T, E>({
     body: json,
   })
   redirectIfUnauthorized(response.status)
+
+  if (response.status === 204) {
+    return { ok: successParser(null) }
+  }
+
   const responseData = await response.json()
   if (response.status === 200 || response.status === 201) {
     return { ok: successParser(responseData) }
-  } else if (response.status === 400) {
+  } else if (Math.floor(response.status / 100) === 4) {
     return { error: errorParser(responseData) }
   }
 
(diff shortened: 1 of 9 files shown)


assets/src/disruptions/viewDisruption.tsx
```diff
@@ -86,7 +143,19 @@ const ViewDisruptionForm = ({
             exceptionDates={exceptionDates}
             disruptionDaysOfWeek={disruptionDaysOfWeek}
           />
-          <EditDisruptionButton disruptionId={disruption.id} />
+          <div>
+            <EditDisruptionButton disruptionId={disruption.id} />
+          </div>
+          {disruption.startDate &&
+            disruption.startDate >= new Date(new Date().toDateString()) && (
```
fjlanasa, line 150: weird that apparently this is the best way to get the start of a day with vanilla JS

assets/src/api.ts
```diff
@@ -40,10 +40,15 @@ const apiSend = async <T, E>({
...
+
+  if (response.status === 204) {
+    return { ok: successParser(null) }
+  }
+
   const responseData = await response.json()
   if (response.status === 200 || response.status === 201) {
     return { ok: successParser(responseData) }
-  } else if (response.status === 400) {
+  } else if (Math.floor(response.status / 100) === 4) {
```
fjlanasa, line 51: should we also handle any 500 errors here?

assets/src/disruptions/viewDisruption.tsx
lemald: Yeah I'm not super enamored with it but it came up when I Googled for solutions and all of the alternatives were much more cumbersome...

assets/src/api.ts
lemald: Hmm, good question. Currently a 500 gets handled by the `Promise.reject("fetch/parse error")` bit below, and there's a test specifically for a 500 response to make sure it hits that line. I guess if we ever run into a case where a 500 error returns useful information in the JSON we can revisit that.

PR #355, opened 2020-07-30
feat: add script to pull and copy database from specified environment

fjlanasa:
#### Summary of changes
**Asana Ticket:** [🏹 script to copy arrow prod data to local DB](https://app.asana.com/0/584764604969369/1184957935964297/f)

Changes:
* moved some helper functions in `DisruptionControllerTest` to `Test.Support.Helper`
* add a mix task `Mix.Tasks.CopyDb`, which fetches adjustments and disruptions from a specified origin (via environment variable, defaults to prod), deletes all existing data, and loads new data within a transaction
* added tests

#### Reviewer Checklist


lib/mix/tasks/copy_db.ex
```diff
@@ -0,0 +1,129 @@
...
+
+        adjustments =
+          Enum.map(adjustments, fn x ->
+            %{
+              id: String.to_integer(x["id"]),
+              route_id: x["attributes"]["route_id"],
+              source: x["attributes"]["source"],
+              source_label: x["attributes"]["source_label"],
+              inserted_at: now,
+              updated_at: now
```
lemald, line 41: Huh, do these not get automatically filled in?

lib/mix/tasks/copy_db.ex
```diff
@@ -0,0 +1,129 @@
...
+            %Arrow.Disruption{
+              id: String.to_integer(x["id"]),
+              start_date: Date.from_iso8601!(attrs["start_date"]),
+              end_date: Date.from_iso8601!(attrs["end_date"]),
+              adjustments:
+                repo.all(
+                  from(adj in Arrow.Adjustment,
+                    where: adj.id in ^adjustment_ids
+                  )
+                ),
```
lemald, line 64: Probably not much of a performance concern right now given that we don't have a tone of data, but I wonder if it's better to do this and fetch the adjustments disruption-by-disruption, or to fetch all of the adjustments at the very beginning and then filter the appropriate ones on the Elixir end here.

lib/mix/tasks/copy_db.ex
fjlanasa: We have to kind of circumvent the normal way of applying changesets here because we have the validations preventing creating new disruptions with dates in the past.

lemald: Hmm, shouldn't it only validate fields like start and end date, not inserted / updated at? If it is currently validating these fields I feel like that's a bug in the validation.

fjlanasa: That's not quite the problem. If I add the disruption via the existing `Disruption.changeset` functions, it will add `inserted/updated_at` automatically, but it will also validate start and end date. If I don't add via the `Disruption.changeset`, it won't validate the start and end date, but won't add the `inserted/updated_at` automatically, which then violates the non-null constraint for those fields.

lemald: Ohh, right, I see. Okay, yeah, that's unfortunate but makes sense.

PR #380, opened 2020-08-07
fix: use UTC dates for calendar

fjlanasa:
#### Summary of changes
**Asana Ticket:** [[extra] fix bug in Arrow calendar dates after 11/1](https://app.asana.com/0/584764604969369/1188024321662505/f)

Problem: We were not handling the quirks of the [`rrule`](https://github.com/jakubroztocil/rrule#important-use-utc-dates) library correctly, which caused dates to be off after daylight savings on 11/1/2020.

> The bottom line is the returned "UTC" dates are always meant to be interpreted as dates in your local timezone. This may mean you have to do additional conversion to get the "correct" local time with offset applied.

Solution: convert dates provided to `rrule` following the example in the library's README
```
// RIGHT: Will produce dates with recurrences at the correct time
new RRule({
  freq: RRule.MONTHLY,
  dtstart: new Date(Date.UTC(2018, 1, 1, 10, 30)),
  until: new Date(Date.UTC(2018, 2, 31))
}).all()

[ '2018-02-01T10:30:00.000Z', '2018-03-01T10:30:00.000Z' ]
```

## SEPTEMBER BEFORE
Screen Shot 2020-08-07 at 2 39 49 PM

## SEPTEMBER AFTER (SHOULD BE THE SAME AS IT WAS CORRECT BEFORE)
Screen Shot 2020-08-07 at 2 40 29 PM

## NOVEMBER BEFORE
Screen Shot 2020-08-07 at 2 39 59 PM

## NOVEMBER AFTER
Screen Shot 2020-08-07 at 2 40 36 PM

#### Reviewer Checklist

### DIFF
--- a/assets/src/disruptions/disruptionCalendar.tsx
+++ b/assets/src/disruptions/disruptionCalendar.tsx
@@ -38,6 +38,19 @@ export const dayNameToInt = (day: DayOfWeek["dayName"]): number => {
   }
 }
 
+const toUTCDate = (date?: Date): Date | undefined => {
+  return (
+    date &&
+    new Date(
+      Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
+    )
+  )
+}
+
+const addDay = (date: Date): Date => {
+  return new Date(date.setTime(date.getTime() + 60 * 60 * 24 * 1000))
+}
+
 export const disruptionsToCalendarEvents = (disruptions: Disruption[]) => {
   return disruptions.reduce(
     (
@@ -63,13 +76,14 @@ export const disruptionsToCalendarEvents = (disruptions: Disruption[]) => {
             byweekday: disruption.daysOfWeek?.map((x) =>
               dayNameToInt(x.dayName)
             ),
-            dtstart: disruption.startDate,
-            until: disruption.endDate,
+            dtstart: toUTCDate(disruption.startDate),
+            until: toUTCDate(disruption.endDate),
           })
         )
         disruption.exceptions.forEach((x) => {
-          if (x.excludedDate) {
-            ruleSet.exdate(x.excludedDate)
+          const excludedDate = toUTCDate(x.excludedDate)
+          if (excludedDate) {
+            ruleSet.exdate(excludedDate)
           }
         })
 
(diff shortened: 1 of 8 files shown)


digitalcora:
Would it be feasible to instead use `Date.UTC` at the point where we load dates from an API response? That way the dates would always already be in UTC, and we wouldn't have to remember to pass them through a function at each point where we're interacting with `rrule`.

fjlanasa:
We're interacting with rrule at one point

fjlanasa:
> Would it be feasible to instead use `Date.UTC` at the point where we load dates from an API response? That way the dates would always already be in UTC, and we wouldn't have to remember to pass them through a function at each point where we're interacting with `rrule`.

It's probably feasible, but I'd rather leave the logic here since it's only to handle the jankiness of the `rrule` library

fjlanasa:
> Would it be feasible to instead use `Date.UTC` at the point where we load dates from an API response? That way the dates would always already be in UTC, and we wouldn't have to remember to pass them through a function at each point where we're interacting with `rrule`.

Ended up requiring less changes than I thought so went ahead and did this.

assets/src/disruptions/disruptionCalendar.tsx
```diff
@@ -68,8 +72,9 @@ export const disruptionsToCalendarEvents = (disruptions: Disruption[]) => {
           })
         )
         disruption.exceptions.forEach((x) => {
-          if (x.excludedDate) {
-            ruleSet.exdate(x.excludedDate)
+          const excludedDate = x.excludedDate
+          if (excludedDate) {
+            ruleSet.exdate(excludedDate)
```
digitalcora, line 77: This variable extraction seems unneeded now — we could revert it all the way back?

assets/src/models/disruption.ts
```diff
@@ -85,10 +85,10 @@ class Disruption extends JsonApiResourceObject {
       return new Disruption({
         id: raw.id,
         ...(raw.attributes.start_date && {
-          startDate: new Date(raw.attributes.start_date + "T00:00:00"),
+          startDate: toUTCDate(raw.attributes.start_date),
         }),
         ...(raw.attributes.end_date && {
-          endDate: new Date(raw.attributes.end_date + "T00:00:00"),
+          endDate: toUTCDate(raw.attributes.end_date),
```
digitalcora, line 91: Good refactor! 😃

PR #381, opened 2020-08-10
feat: add react-select for adjustment picker

fjlanasa:
#### Summary of changes
**Asana Ticket:** [🏹 Combo-box for adjustment selection](https://app.asana.com/0/584764604969369/1157557244099569/f)

Changes:
* replace current `AdjustmentPicker` with new version using `react-select`
* change frontend `Adjustment` model to make `id`, `routeId` and `sourceLabel` required
* updated tests as needed

select

#### Reviewer Checklist

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -6,7 +6,6 @@
       "version": "7.10.3",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.3.tgz",
       "integrity": "sha512-fDx9eNW0qz0WkUeqL6tXEXzVlPh6Y5aCDEZesl0xBGA8ndRukX91Uk44ZqnkECp01NAZUdCAl+aiQNGi0k88Eg==",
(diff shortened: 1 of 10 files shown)


fjlanasa:
> I like the amount of deleted code in this one 😄
> 
> As a general comment, we seem to be doing a lot of this in tests:
> 
> ```js
> const thing = container.querySelector("#thing")
> 
> if (thing) {
>   doSomethingWith(thing)
> } else {
>   throw new Error("thing not found")
> }
> ```
> 
> It seems like we could reduce some of this "noise" by instead doing this:
> 
> ```js
> const thing = container.querySelector("#thing")!
> doSomethingWith(thing)
> ```
> 
> ...assuming `doSomethingWith` will crash anyway if `thing` is not defined, giving the same effect as though we had thrown our own error. Would that work?

Yeah I personally probably wouldn't have done it this way in the tests, but I just kept it similar for the sake of consistency.

PR #390, opened 2020-08-18
feat: Add concept of 'revisions' to disruption changes

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 Arrow uses new schema](https://app.asana.com/0/584764604969369/1185700222235211)

Adds the database support for the "published" vs "draft" feature. For now, every change is automatically published, so current app behavior is the same. However, behind the scenes it will now be storing each change as a DisruptionRevision and a cloned set of associated relationships.

Per discussion, we removed some of the time-based validations, since we've found that we sometimes actually _do_ want to make retroactive changes, but have left a lot of the "obviously wrong" checks, like exception dates falling within the date range and on relevant days of the week.

#### Reviewer Checklist


lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -1,36 +1,52 @@
...
+
+    data =
+      query
+      |> Repo.all()
+      |> Repo.preload([:adjustments, :days_of_week, :exceptions, :trip_short_names])
+
+    # IO.inspect(data)
+    # require IEx
+
+    # IEx.pry()
```
lemald, line 26: debugging code

lib/arrow/disruption_revision.ex
```diff
@@ -0,0 +1,94 @@
...
+          trip_short_names: [TripShortName.t()] | Ecto.Association.NotLoaded.t(),
+          adjustments: [Arrow.Adjustment.t()] | Ecto.Association.NotLoaded.t(),
+          inserted_at: DateTime.t() | nil,
+          updated_at: DateTime.t() | nil
+        }
+
+  schema "disruption_revisions" do
+    field :end_date, :date
+    field :start_date, :date
+    field :is_active, :boolean
```
fjlanasa, line 27: does this have a default, or is it implicitly false?

test/arrow/disruption_test.exs
```diff
@@ -1,774 +1,403 @@
...
-               disruption
-               |> Disruption.changeset_for_delete(@current_time)
-               |> Repo.delete()
+      Repo.update!(Ecto.Changeset.change(d, %{published_revision_id: dr.id}))
 
-      assert changeset.errors == [
-               start_date: {"can't be deleted when start date is in the past", []}
-             ]
+      assert {:ok, dr} = Arrow.Disruption.delete(dr.id)
+      assert Repo.all(from(dr in Arrow.DisruptionRevision, select: count(dr.id))) == [2]
```
fjlanasa, line 399: is the original `DisruptionRevision` marked as `is_active: false` in this case?

test/arrow/disruption_test.exs
```diff
@@ -1,774 +1,403 @@
...
+               %{detail: "Adjustments should have at least 1 item(s)"}
+             ] = ArrowWeb.Utilities.format_errors(e)
     end
+  end
 
-    test "Can't delete exception date that's in the past" do
-      start_date = ~D[2020-01-01]
-      end_date = ~D[2020-01-20]
-      disruption = build_disruption(%Arrow.Disruption{start_date: start_date, end_date: end_date})
+  describe "update/2" do
```
fjlanasa, line 163: what happens to the `is_active` field on the copied `DisruptionRevision`?

losvedir:
Seeing your questions here, I think `is_active` is a bit confusing. It essentially means `is_deleted: false`, the way we're using it. So, previous disruption revisions are correctly left as `is_deleted: false`. It *doesn't* mean that's the revision that is currently serving as the "draft" or "published" revision or something like that. Its genesis, I think, came from my original proposed "statuses" of `pending_deactivation`, `inactive`, `pending_activation`, `active`, but those ended up being unnecessary other than the one use of flagging a revision as deleted.

Should we consider renaming this field? Is `is_deleted` any better? I want to tease apart the meanings of "is this the revision that's being used" vs. "does this revision indicate that the disruption has been / will be deleted".

And regarding the questions of defaults, if we do change it, it's probably worth adding a postgres default of it not being deleted.

lib/arrow/disruption_revision.ex
lemald: FWIW the migration does set it to default to true in Postgres.

lemald:
If I recall correctly @digitalcora had some thoughts on the `is_deleted` vs. `is_active` issue, but I don't remember exactly why we went with `is_active`.

lib/arrow/disruption/day_of_week.ex
```diff
@@ -24,7 +24,7 @@ defmodule Arrow.Disruption.DayOfWeek do
   @spec changeset(t(), map()) :: Ecto.Changeset.t()
   def changeset(day_of_week, attrs) do
     day_of_week
-    |> cast(attrs, [:id, :day_name, :start_time, :end_time])
```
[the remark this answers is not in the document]
losvedir: I don't think we do. It works without it. In fact, I believe I had to remove that to *make* it work. That said, from what I semantically understand `cast` to mean, is that these are the parameters we want the user to be able to set. We *don't* want them to set the `id`, so I feel like it doesn't belong in the list. I don't remember why we added it initially, but if it's not necessary now (which from testing and experimentation it isn't), I think it's good to remove.

losvedir:
Alternatively, we could retain `is_active` and have it mean something along the lines of "currently serving as the revision for a (published? draft?) disruption". Then we'd have to change the update code to set previous versions to `false` as necessary. However, I like that less because multiple disruption revisions are "active" for different views at the same time. It also seems to me denormalized in what is canonically the revision for a view - is it the one with the `is_active` flag set a certain way or is it the latest one for a disruption?

lemald:
> Alternatively, we could retain `is_active` and have it mean something along the lines of "currently serving as the revision for a (published? draft?) disruption". Then we'd have to change the update code to set previous versions to `false` as necessary. However, I like that less because multiple disruption revisions are "active" for different views at the same time. It also seems to me denormalized in what is canonically the revision for a view - is it the one with the `is_active` flag set a certain way or is it the latest one for a disruption?

Yeah, I share your hesitation about this approach - I would worry that it would be hard to ensure that that sort of denormalization is kept in sync properly.

fjlanasa:
> > Alternatively, we could retain `is_active` and have it mean something along the lines of "currently serving as the revision for a (published? draft?) disruption". Then we'd have to change the update code to set previous versions to `false` as necessary. However, I like that less because multiple disruption revisions are "active" for different views at the same time. It also seems to me denormalized in what is canonically the revision for a view - is it the one with the `is_active` flag set a certain way or is it the latest one for a disruption?
> 
> Yeah, I share your hesitation about this approach - I would worry that it would be hard to ensure that that sort of denormalization is kept in sync properly.

Ah, ok. I was conflating a `DisruptionRevision` set to `is_active: true` with it's associated `Disruption` having a `published_revision_id` matching the revision.

fjlanasa:
In that case, I don't have strong feelings about `is_active` vs `is_deleted`

digitalcora:
> If I recall correctly @digitalcora had some thoughts on the `is_deleted` vs. `is_active` issue, but I don't remember exactly why we went with `is_active`.

I think `is_deleted` is equally reasonable, and if it would be clearer in this context then we should use it.

@losvedir I think it would be a viable approach, since we are already dynamically defining "the latest revision" without storing an ID anywhere, to also define "the published revision" as "the latest revision that has `is_published` set to true". The act of publishing a revision would change from a Disruption update to creating a new revision identical to the previous one but with the flag set (it would have to be cleared by default when creating new revisions). This would also allow us to keep track of the history of published revisions, which the `published_revision_id` column notably doesn't allow on its own.

All that said, this could be done separately from this change in a backwards-compatible way that likely wouldn't interfere with other work too much.

test/arrow/disruption_test.exs
```diff
@@ -1,774 +1,403 @@
...
-               |> Disruption.changeset_for_delete(@current_time)
-               |> Repo.delete()
+      Repo.update!(Ecto.Changeset.change(d, %{published_revision_id: dr.id}))
 
-      assert changeset.errors == [
-               start_date: {"can't be deleted when start date is in the past", []}
-             ]
+      assert {:ok, dr} = Arrow.Disruption.delete(dr.id)
+      assert Repo.all(from(dr in Arrow.DisruptionRevision, select: count(dr.id))) == [2]
+      assert dr.is_active == false
```
fjlanasa, line 400: maybe we should add an assertion that the `Disruption.published_revision_id == dr.id`

test/arrow/disruption_test.exs
```diff
@@ -1,774 +1,403 @@
...
+      assert length(dr_ids) == 2
+      new_dr_id = Enum.find(dr_ids, &(&1 != dr_id))
+
+      new_dr =
+        Arrow.DisruptionRevision
+        |> Repo.get!(new_dr_id)
+        |> Repo.preload([:exceptions, :days_of_week, :trip_short_names])
+
+      assert new_dr.disruption_id == d.id
+      assert new_dr.disruption_id == d.id
```
fjlanasa, line 201: duplicate assertion here. maybe we should add an assertion that the Disruption.published_revision_id == dr.id

losvedir: Oh, good catch on the duplicate. We'll remove.

I thought about the `published_revision_id == id` thing, but while that's currently the case for `Disruption.update`, it won't be going forward. I think we can skip the assertion for now, but I also wouldn't be opposed to adding it, and then updating the test when the behavior changes.

fjlanasa: ah, ok. fair enough about the extra assertion.

lemald: I like the idea of testing the current behavior now and just updating the test when we update the behavior.

PR #391, opened 2020-08-18
feat: fetch adjustments periodically from S3

digitalcora:
**Asana Ticket:** [🏹 Arrow uses gtfs_creator shuttles in S3](https://app.asana.com/0/584764604969369/1186313575712853)

https://github.com/mbta/gtfs_creator/pull/984 started uploading the file that this process downloads.

#### Reviewer Checklist


lib/arrow/application.ex
```diff
@@ -49,7 +43,7 @@ defmodule Arrow.Application do
   end
 
   def adjustment_fetcher_children(true) do
-    [{Arrow.AdjustmentFetcher, path: Application.app_dir(:arrow, "priv/repo/shuttles.json")}]
+    [{Arrow.AdjustmentFetcher, []}]
```
lemald, line 46: Do we still need a separate `adjustment_fetcher_children/1` function? It's no longer needed for passing in the `path` argument, so all it's really doing is conditionally including or excluding the `Arrow.AdjustmentFetcher` process.

test/arrow/adjustment_fetcher_test.exs
```diff
@@ -1,76 +1,76 @@
-defmodule Arrow.PredictionFetcherTest do
```
lemald, line 1: Haha, oops, I must have carried the module name over from Prediction Analyzer or wherever else I was cribbing this from when I first created the module.

lemald:
I think this looks generally good! I have one minor comment / question and there are still a few uncovered lines, but apart from those things I think it's basically ready to approve.

lib/arrow/application.ex
digitalcora: In attempting to improve the test coverage I realized the URL should be dependency-injected here, so it actually serves a purpose again 😄

losvedir:
Hm, I like the upgrade from the ticket of just checking at start up to periodically fetching!

I will say that I'm not really a fan of `Bypass`. I remember being generally annoyed with it when working on `dotcom` but at this point I don't remember why. Since you're already injecting in `:http_client`, can you just provide a `HttpSuccess` and `HttpFailure` (and others if necessary) modules in the test suite, with a `get/1` function returning whatever you need?

Also, I think there might be some conflicts (hopefully not too bad) with the DB Schema PR. Can you wait for that one to go in before merging this one?

digitalcora:
> Hm, I like the upgrade from the ticket of just checking at start up to periodically fetching!

I can't claim credit, just following along with what the moduledoc previously said:

```
  Reads in information on what adjustments are present in GTFS Creator
  and saves them to the database. Currently reads from the `priv`
  directory once on startup, but eventually this will poll some sort
  of API on a regular basis.
```

> I will say that I'm not really a fan of `Bypass`. I remember being generally annoyed with it when working on `dotcom` but at this point I don't remember why. Since you're already injecting in `:http_client`, can you just provide a `HttpSuccess` and `HttpFailure` (and others if necessary) modules in the test suite, with a `get/1` function returning whatever you need?

Probably. I would like to hear a rationale beyond "I remember being annoyed with Bypass but don't remember why", but if it turns out we can do this with one less dependency and without much extra code, that's a decent reason to me 🙂

digitalcora:
I took a quick try at removing Bypass, but it seems rather complicated — it's not simply about having a module with a `get` function that returns something, we also want to verify that the request is being made (e.g. in the `start_link` test, though having the assertion is nice everywhere). In my mind a good reason to use Bypass is the relative simplicity of the built-in assertions like `expect_once` vs. building the state management and cross-process messaging needed to do the same thing ourselves. Is there an easier pattern I'm not seeing here?

digitalcora:
@losvedir Just pushed a separate commit in which I try using Mox to mock HTTPoison directly rather than setting up the fake server using Bypass. Let me know if this seems better to you, or if there's another approach you'd recommend?

test/arrow/adjustment_fetcher_test.exs
```diff
@@ -1,67 +1,88 @@
...
+          {:ok, fetcher} = AdjustmentFetcher.start_link(interval: 100)
+
+          HTTPMock
+          |> expect(:get, successful_get)
+          |> expect(:get, failed_get)
+          |> expect(:get, successful_get)
+          |> allow(parent, fetcher)
+
+          # wait for it to make all three requests, then stop it
+          Enum.each(1..3, fn _ -> assert_receive :requested, 200 end)
```
losvedir, line 38: Do you need these assertions? As I understood it, I thought `expect` and `verify_on_exit` sort of combine to give you this coverage.

digitalcora: If we didn't wait for the requests here, the test would finish immediately before the fetcher had a chance to do anything, tearing down the test process and the fetcher along with it. `verify_on_exit` then fails the test:

```
* expected Arrow.HTTPMock.get/1 to be invoked 3 times but it was invoked 0 times
```

losvedir: Ah, right.

PR #394, opened 2020-08-20
fix: Regenerate AWS IAM auth token on connect

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 Reconnect after DB authentication changes](https://app.asana.com/0/584764604969369/1186678199122409)

The fix here is actually pretty simple. `Ecto.Repo` options are passed to the `Postgrex` adaptor which are passed to [`DBConnection`](https://hexdocs.pm/db_connection/DBConnection.html#start_link/2), which has a nice `:configure` option which takes a MFA tuple for a function it should call before each time it tries to connect.

By moving the `ex_aws_rds` logic into that function, the actual configuration of the DB can now be done with the usual, simple `config :arrow, Arrow.Repo, ...` approach. Since the prod environment will pull from environment variables, I made a `releases.exs` file for `mix release` to use.

I confirmed that I could induce the `PAM authentication failed` error on arrow-dev running the master branch by rebooting the RDS instance from the AWS console, at least [15 minutes](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html) after the last deploy. With this branch running when I reboot the RDS instance it works fine, and I see the `generating_token` log message.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/repo.ex
+++ b/lib/arrow/repo.ex
@@ -3,39 +3,23 @@ defmodule Arrow.Repo do
     otp_app: :arrow,
     adapter: Ecto.Adapters.Postgres
 
-  @default_pool_size 10
+  require Logger
 
-  def init(_context, config) do
-    pool_size_env = System.get_env("DATABASE_POOL_SIZE")
-    pool_size_env = if pool_size_env, do: String.to_integer(pool_size_env)
-    pool_size = pool_size_env || Keyword.get(config, :pool_size) || @default_pool_size
-    config = Keyword.put(config, :pool_size, pool_size)
+  @doc """
+  Set via the `:configure` option in the Arrow.Repo configuration, a function
+  invoked prior to each DB connection. `config` is the configured connection values
+  and it returns a new set of config values to be used when connecting.
+  """
+  def before_connect(config) do
+    :ok = Logger.info("generating_aws_rds_iam_auth_token")
+    username = Keyword.fetch!(config, :username)
+    hostname = Keyword.fetch!(config, :hostname)
+    port = Keyword.fetch!(config, :port)
 
-    config =
-      if Keyword.take(config, [:database, :username, :password, :hostname]) == [] do
-        username = System.get_env("DATABASE_USER")
-        database = System.get_env("DATABASE_NAME")
-        hostname = System.get_env("DATABASE_HOST")
-        port = "DATABASE_PORT" |> System.get_env() |> String.to_integer()
+    mod = Application.get_env(:arrow, :aws_rds_mod)
+    token = apply(mod, :generate_db_auth_token, [hostname, username, port, %{}])
+    :ok = Logger.info("generated_aws_rds_iam_auth_token")
 
-        if !username, do: raise("missing DATABASE_USER environment variable")
-        if !database, do: raise("missing DATABASE_NAME environment variable")
-        if !hostname, do: raise("missing DATABASE_HOST environment variable")
-
-        mod = Application.get_env(:arrow, :aws_rds_mod)
-        token = apply(mod, :generate_db_auth_token, [hostname, username, port, %{}])
-
-        Keyword.merge(config,
-          database: database,
-          username: username,
-          hostname: hostname,
-          password: token,
-          port: port
-        )
-      else
-        config
-      end
-
-    {:ok, config}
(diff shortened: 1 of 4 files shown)


lib/arrow/repo.ex
```diff
@@ -3,39 +3,23 @@ defmodule Arrow.Repo do
...
-    pool_size_env = if pool_size_env, do: String.to_integer(pool_size_env)
-    pool_size = pool_size_env || Keyword.get(config, :pool_size) || @default_pool_size
-    config = Keyword.put(config, :pool_size, pool_size)
+  @doc """
+  Set via the `:configure` option in the Arrow.Repo configuration, a function
+  invoked prior to each DB connection. `config` is the configured connection values
+  and it returns a new set of config values to be used when connecting.
+  """
+  def before_connect(config) do
+    :ok = Logger.info("Arrow.Repo.before_connect generating_token")
```
digitalcora, line 14: We could use `"#{__MODULE__}"` here to keep this accurate regardless of where the code lives — I don't know if it's really necessary to include the module name, though, if the log string is globally unique.

lib/arrow/repo.ex
```diff
@@ -3,39 +3,23 @@ defmodule Arrow.Repo do
...
-    pool_size_env = System.get_env("DATABASE_POOL_SIZE")
-    pool_size_env = if pool_size_env, do: String.to_integer(pool_size_env)
-    pool_size = pool_size_env || Keyword.get(config, :pool_size) || @default_pool_size
-    config = Keyword.put(config, :pool_size, pool_size)
+  @doc """
+  Set via the `:configure` option in the Arrow.Repo configuration, a function
+  invoked prior to each DB connection. `config` is the configured connection values
+  and it returns a new set of config values to be used when connecting.
+  """
+  def before_connect(config) do
```
digitalcora, line 13: Is this named `before_connect` by convention? I would have expected it to be called `configure`, in line with the option.

test/arrow/repo_test.exs
```diff
@@ -1,76 +1,19 @@
 defmodule Arrow.RepoTest do
   use ExUnit.Case, async: false
+  import Test.Support.Helpers
+  import ExUnit.CaptureLog
```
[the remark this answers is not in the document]
losvedir: Good catch, I'll delete. I also added a bonus commit to add `mix compile --force --warnings-as-errors` to one of our semaphore checks.

lib/arrow/repo.ex
losvedir: I initially didn't use `#{__MODULE__}` just to save the unnecessary string allocation, but that was probably not necessary. In any case, I just removed it, and made the unique strings a little more specific.

lib/arrow/repo.ex
losvedir: I thought about calling it `configure`, but I thought that was less immediately obvious about what it did. The option description says:

> `:configure` - A function to run before every connect attempt to dynamically configure the options...

I felt that calling out "run before every connect attempt" was pretty core to what the function did. Whereas `configure` by itself could potentially run when the app starts up or some other time. In addition, there is an option you can specify `after_connect`, which is a function that runs after a connection attempt, so I was "inspired" by that.

semaphore/credo.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 set -ex
 
+mix compile --force --warnings-as-errors
```
digitalcora, line 4: Has the app already been compiled by this point in the build process? In other words, is this re-compiling the app, or compiling it for the first time? If the latter, why would `--force` be necessary?

lib/arrow/repo.ex
digitalcora: Makes sense to me!

semaphore/credo.sh
losvedir: Currently, no, but I figured why not? It seems conceivable we might have a custom `mix` task running in `setup.sh` or something which would incidentally compile the app. Basically, if the point of this line is to trigger `--warnings-as-errors`, I think it should ensure itself that it actually is compiling it. Just felt more robust to future changes, especially in another file which when changed probably won't be by someone thinking about ramifications in this file. I didn't see any downside to having `--force`.

PR #396, opened 2020-08-21
feat: add nav links to toggle between published and draft disruption views

fjlanasa:
#### Summary of changes
**Asana Ticket:** [🏹 Navigation bar with disruption and review pages ](https://app.asana.com/0/584764604969369/1185700222235219/f)

Changes:
* Added `DisruptionListContainer` component containing the new navigation links, which can wrap `DisruptionIndex`, the coming "Needs Review" page or any other page we'd want to have those links for
* Pulled out the logic we had rendering `Header` in each individual page into a `Page` wrapper that handles it for us
* Some small style tweaks 

toggle

#### Reviewer Checklist

### DIFF
--- a/assets/src/disruptions/disruptionListContainer.tsx
+++ b/assets/src/disruptions/disruptionListContainer.tsx
@@ -1,0 +1,33 @@
+import * as React from "react"
+import { withRouter, RouteComponentProps } from "react-router"
+import { NavLink } from "react-router-dom"
+
+const DisruptionListContainer = withRouter(
+  ({ children }: RouteComponentProps & { children: React.ReactNode }) => {
+    return (
+      <div>
+        <div className="d-flex border-bottom">
+          <NavLink
+            to="/"
+            exact={true}
+            className="m-disruption-list-container_nav-link"
+            activeClassName="active"
+          >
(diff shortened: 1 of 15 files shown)


assets/src/disruptions/disruptionListContainer.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { withRouter, RouteComponentProps } from "react-router"
+import { NavLink } from "react-router-dom"
+
+const DisruptionListContainer = withRouter(
+  ({ children }: RouteComponentProps & { children: React.ReactNode }) => {
+    return (
+      <div>
+        <div className="d-flex border-bottom">
+          <NavLink
+            to="/"
```
lemald, line 11: This might be beyond the scope of this PR, but I wonder... if we're creating a new `/disruptions/needs_review` path, should we maybe change the path of the existing published view to somehow match? Something like `/disruptions/published` perhaps?

fjlanasa: Hmm, yeah I suppose I could go either way. I think I'll just punt on changing for not because I expect it would require changing a lot of unrelated tests and whatnot.

PR #399, opened 2020-08-24
Distinguish published / draft disruptions

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 Arrow distinguishes between published and draft versions](https://app.asana.com/0/584764604969369/1185700222235213/f)

This one is probably most easily-reviewed on a commit-by-commit basis, and the order of the commits should match the order of the acceptance criteria bullet points on the ticket.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/api/disruption_controller.ex
+++ b/lib/arrow_web/controllers/api/disruption_controller.ex
@@ -14,41 +14,40 @@ defmodule ArrowWeb.API.DisruptionController do
       |> format_filters
       |> build_query
 
-    with {:ok, query} <-
-           case(Map.get(params, "only_published"),
-             do:
-               (
-                 "true" -> {:ok, DisruptionRevision.only_published(query)}
-                 "false" -> {:ok, DisruptionRevision.latest_revision(query)}
-                 nil -> {:ok, DisruptionRevision.latest_revision(query)}
-                 _ -> :error
-               )
-           ),
-         data <-
-           query
-           |> Repo.all()
-           |> Repo.preload([:adjustments, :days_of_week, :exceptions, :trip_short_names]) do
-      render(conn, "index.json-api",
-        data: data,
-        opts: [include: Map.get(params, "include")]
-      )
-    else
-      _ -> conn |> put_status(400) |> render(:errors, errors: [%{detail: "Invalid request"}])
+    case do_only_published_query(query, params) do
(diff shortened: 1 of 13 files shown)


lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -13,31 +13,53 @@ defmodule ArrowWeb.API.DisruptionController do
...
-      data: data,
-      opts: [include: Map.get(params, "include")]
-    )
+    with {:ok, query} <-
+           case(Map.get(params, "only_published"),
+             do:
+               (
+                 "true" -> {:ok, DisruptionRevision.only_published(query)}
+                 "false" -> {:ok, DisruptionRevision.latest_revision(query)}
+                 nil -> {:ok, DisruptionRevision.latest_revision(query)}
```
fjlanasa, line 23: does this handle the case where we send `only_published=` with no value for the param?

lemald: Yup, `Map.get` will return `nil` in that case.

fjlanasa: Also, any reason not to add this to the `build_query` function? Is it just because we have a different match pattern here?

test/arrow_web/controllers/api/disruption_controller_test.exs
```diff
@@ -84,6 +84,116 @@ defmodule ArrowWeb.API.DisruptionControllerTest do
       end)
     end
 
+    @tag :authenticated
+    test "includes latest revision by default", %{conn: conn} do
```
fjlanasa, line 88: Do you think it'd be valuable to add another Disruption with only one (published) DisruptionRevision to this case, so we see that both the published and latest draft are returned?

lib/arrow_web/controllers/api/disruption_controller.ex
lemald: I don't think it quite makes sense to put into `build_query` but this comment _did_ lead me to realize that a lot of this could be split out into a different helper function, thereby reducing some shared code.

test/arrow_web/controllers/api/disruption_controller_test.exs
lemald: Yup, good idea. I just pushed up a revised test.

PR #405, opened 2020-08-26
fix: accept empty string for only_published query param

lemald:
#### Summary of changes
**Asana Ticket:** Followup to [🏹 Arrow distinguishes between published and draft versions](https://app.asana.com/0/584764604969369/1185700222235213/f)

After some discussion, realized we should be treating an empty query param as falsy. Also broke out the `?only_published=false` case into its own test to match what I did for `index/2`.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/api/disruption_controller.ex
+++ b/lib/arrow_web/controllers/api/disruption_controller.ex
@@ -180,8 +180,7 @@ defmodule ArrowWeb.API.DisruptionController do
   defp do_only_published_query(q, params) do
     case Map.get(params, "only_published") do
       "true" -> {:ok, DisruptionRevision.only_published(q)}
-      "false" -> {:ok, DisruptionRevision.latest_revision(q)}
-      nil -> {:ok, DisruptionRevision.latest_revision(q)}
+      val when val in [nil, "", "false"] -> {:ok, DisruptionRevision.latest_revision(q)}
       _ -> :error
     end
   end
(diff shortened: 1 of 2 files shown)


lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -178,10 +178,10 @@ defmodule ArrowWeb.API.DisruptionController do
 
   @spec do_only_published_query(Ecto.Queryable.t(), map()) :: {:ok, Ecto.Query.t()} | :error
   defp do_only_published_query(q, params) do
-    case Map.get(params, "only_published") do
+    case Map.get(params, "only_published", "") do
       "true" -> {:ok, DisruptionRevision.only_published(q)}
       "false" -> {:ok, DisruptionRevision.latest_revision(q)}
-      nil -> {:ok, DisruptionRevision.latest_revision(q)}
+      "" -> {:ok, DisruptionRevision.latest_revision(q)}
```
digitalcora, line 184: Can be done with a bit less duplication:

```elixir
case Map.get(params, "only_published") do
  "true" -> {:ok, DisruptionRevision.only_published(q)}
  val when val in [nil, "", "false"] -> {:ok, DisruptionRevision.latest_revision(q)}
```

Here I also co-located the distinct values that will result in the latest revision, making it clear there are three of them (nil, false, and empty string).

lemald: Ah, yes, I like that! Just pushed up the change.

PR #408, opened 2020-08-31
feat: "Review changes" page

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 Review page enumerates differences between draft and published](https://app.asana.com/0/584764604969369/1186492610975125)

This adds a simple "review changes" page. I didn't spend too much time on the design because the one in the ticket is tentative and from a quick peek at Ingrid's new designs, I think we're getting rid of this page altogether.

Screen Shot 2020-08-31 at 12 31 12 PM

The first commit is an independent fixup of our JSON:API deserialization. Our old one had an issue where it wouldn't deserialize nested resources. For example, we might have a foo with a bar relationship, and that bar might have a baz. And then in the top level `included` array, we have the relevant `bar` and `baz`. This PR updates the logic to do _two_ passes through the `included` objects. The first time parses all the top level items, and the second time then inserts the related items in the right place.

The second commit is the meat of the PR. At its core is a query that returns the disruptions whose published and draft revisions are different, preloaded with all the revisions between the two. There's also an important function `DisruptionDiff.diff/2` which takes two preloaded revisions and returns a list of strings describing their differences. For now it's a list of lists, though in the future it might be a list of a more structured object, in order to accommodate author information, though at this point we're not tracking that.

#### Reviewer Checklist


assets/tests/models/disruptionDiff.test.ts
```diff
@@ -0,0 +1,7 @@
+import DisruptionDiff from "../../src/models/disruptionDiff"
+
+describe("DisruptionDiff", () => {
+  test("Hanldes parse error", () => {
```
lemald, line 4: Typo

lib/arrow/disruption.ex
```diff
@@ -104,6 +107,53 @@ defmodule Arrow.Disruption do
...
+        where: dr.id >= d.published_revision_id and dr.id <= as(:draft_map).draft_id,
+        preload: [revisions: {dr, [:days_of_week, :exceptions, :trip_short_names, :adjustments]}]
+      )
+      |> Arrow.Repo.all()
+
+    new =
+      from(d in Arrow.Disruption, where: is_nil(d.published_revision_id))
+      |> Arrow.Repo.all()
+      |> Arrow.Repo.preload(
+        revisions: [:days_of_week, :exceptions, :trip_short_names, :adjustments]
```
lemald, line 139: Just thinking out loud - I wonder if it would make sense to have a constant representing all of these associations that we tend to preload for each disruption. I suspect we reference this in several places around the app.

lemald:
Added a couple of comments but I think it's mostly good! I'll want to sleep on it and give it another pass once you've got codecov up to 100%, but that can wait until tomorrow.

lib/arrow/disruption.ex
losvedir: Good idea!

losvedir:
> My one overall thought / comment is that we're breaking up the changes by individual revision when presenting them to the user, rather than thinking of there being one big draft that gets further updated as more edits are made.

Fortunately, because of the clone + edit approach, these two approaches are not so different, really. We can either iterate over a list of revisions to get each transformation (optionally with author information, which I still like, but is not anywhere in the current design thinking, as far as I know!), or simply diff the first and last revision for the "one big draft" approach. The `DR.diff/2` calculation function could handle either approach with basically no change.

> We're going to completely redo how this information is presented anyway, but I wonder if it would make sense to reconsider the implementation of the DisruptionDiff module in light of our current design thinking.

Yeah, you're right to raise the question. In light of the new homepage design a lot of this won't be relevant soon. I don't think the current design is planning to have a "list all changes" page like this. I still want to merge it since it is done and works and will be useful for a week or two, and can be harvested for parts, but probably a few weeks from now we can remove it. (I'll make a Tech Debt ticket to look into this.) On the other hand, maybe from use we'll find it's a more intuitive way of using the app and we'll want to keep it / incorporate it into the design.

lemald:
Right, agreed that it's not a huge change either way. I'd say it's fine to go with this approach for now and we can poke at it more in the future as we continue to implement more of Ingrid's design.

PR #425, opened 2020-09-17
feat: 🏹 Update frontend to handle mega/giga rows + styling updates

fjlanasa:
#### Summary of changes
**Asana Tickets:** 
* [🏹 Update frontend to handle mega/giga rows](https://app.asana.com/0/584764604969369/1191956107867594/f)
* [🏹 Update header styling + font](https://app.asana.com/0/584764604969369/1191956107867637/f)
* [🏹 Styling update](https://app.asana.com/0/584764604969369/1191956107867641/f)
* [🏹 Add filtering by "published/ready" and "needs review"](https://app.asana.com/0/584764604969369/1191956107867610/f)
* [Days of week not combined in Arrow display](https://app.asana.com/0/584764604969369/1194249214405201/f)

Changes:
* Added `file-loader` package to handling loading and applying new fonts
* Added `postcss-loader` package with `autoprefixer` plugin to handle overriding Bootstrap theming
* Added `Button` component, a wrapper around the Bootstrap component, along with implementations for `PrimaryButton`, `SecondaryButton`, and `LinkButton` to match new styles
* Added `status` field to `DisruptionRevision` to be used for filtering in `DisruptionIndex`
* Updated `DisruptionIndex`
  * changed from picking one revision to display to reducing from a `Disruption` to a list of unique `DisruptionRevisions`
  * added buttons to filter revisions by status
* Updated `DisruptionTable`
  * Added new columns and made all columns sortable
  * Added logic to show the "rowlettes" when a revision shares the "label" of its parent
  * fixed parsing of time period column
* other styling enhancements

megarows

#### Reviewer Checklist


losvedir:
Just a quick question based off the gif (very neat, btw!): what view of the data does the calendar show?

fjlanasa:
> Just a quick question based off the gif (very neat, btw!): what view of the data does the calendar show?

The same as would be shown on the table

ingridpierre:
The focus/hover state I originally made for the secondary button is a bit too subtle and judders when the font changes weight. Let's use the updated style here: https://zeroheight.com/656756d9b/p/30324c-button/t/06c501

image

ingridpierre:
Some of the table widths are causing more wrapping within the row than desired. I've got a couple updates/tweaks that might help: https://zeroheight.com/656756d9b/p/76c9a5-home/t/547a39

- Let's remove the "routes" column and just prepend a smaller route icon with the adjustment name. Hoping that should also help when having to show multiple routes associated with an adjustment
- Let's shorten "exceptions" to "except."
- Let's add a min width to the status buttons so needs review doesn't wrap (or alternatively, fix width to accommodate all labels)

ingridpierre:
Some stylistic things that Jessie and I want to update as well:

- Where space is limited (like in the home rows), let's shorten days to 3 letter abbreviations. **Saturday >>> Sat**
- let's format dates as MM/DD/YYYY such that we have leading zeros on single digit dates. **e.g. 09/01/2020**

fjlanasa:
> Just a quick question based off the gif (very neat, btw!): what view of the data does the calendar show?

It shouldn't be too hard to force one view for the calendar page, just wasn't sure which it should be.

fjlanasa:
@ingridpierre just pushed up a few commits that should have addressed your comments so far. will redeploy to dev.

fjlanasa:
> Just a quick question based off the gif (very neat, btw!): what view of the data does the calendar show?

@losvedir, after talking with @ingridpierre in [Slack](https://mbta.slack.com/archives/CF9QKK2AF/p1600434066003300), we decided to only display published revisions on the calendar view, and will disable the status filters when in that view.

losvedir:
Small bug: If you click `published/ready` and then click `needs review` the second circle also lights up (but doesn't appear to do anything). The reverse direction of `needs review` followed by `published/ready` does the right thing of deselecting the first button and selecting the second.

fjlanasa:
> Small bug: If you click `published/ready` and then click `needs review` the second circle also lights up (but doesn't appear to do anything). The reverse direction of `needs review` followed by `published/ready` does the right thing of deselecting the first button and selecting the second.

Oh woops. I think what you are describing as the "right thing" is actually the wrong thing. It's not meant to be a radio type button. It would work like the first case (which doesn't do anything on dev because there are no disruptions that only have a draft version)

fjlanasa:
> > Small bug: If you click `published/ready` and then click `needs review` the second circle also lights up (but doesn't appear to do anything). The reverse direction of `needs review` followed by `published/ready` does the right thing of deselecting the first button and selecting the second.
> 
> Oh woops. I think what you are describing as the "right thing" is actually the wrong thing. It's not meant to be a radio type button. It should work like the first case (which doesn't do anything on dev because there are no disruptions that only have a draft version)

In other words, it should work as the route filters do.

losvedir:
>  It's not meant to be a radio type button.

Is this something that came out of your chat with Ingrid? We groomed [a separate task](https://app.asana.com/0/881264583703207/1191956107867610) for this functionality (but it's fine that you pulled it in here). But that task mentions it should behave like a radio.

fjlanasa:
> > It's not meant to be a radio type button.
> 
> Is this something that came out of your chat with Ingrid? We groomed [a separate task](https://app.asana.com/0/881264583703207/1191956107867610) for this functionality (but it's fine that you pulled it in here). But that task mentions it should behave like a radio.

Yes

fjlanasa:
> > > Small bug: If you click `published/ready` and then click `needs review` the second circle also lights up (but doesn't appear to do anything). The reverse direction of `needs review` followed by `published/ready` does the right thing of deselecting the first button and selecting the second.
> > 
> > 
> > Oh woops. I think what you are describing as the "right thing" is actually the wrong thing. It's not meant to be a radio type button. It should work like the first case (which doesn't do anything on dev because there are no disruptions that only have a draft version)
> 
> In other words, it should work as the route filters do.

@ingridpierre can you confirm that I'm remembering our conversation about this correctly?

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -38,7 +34,7 @@ type RouteFilterState = {
   [route in Routes]?: boolean
 }
 
-const getRouteIcon = (route?: string): Icon => {
```
losvedir, line 41: We've traditionally done only `export { ... }` at the bottom. I remember this coming up before; I think your experience is in `export`'ing as you go. I'm fine switching to that, I just don't like having _both_ approaches in one file.

Seeing the `export { DisruptionIndex, RouteFilterToggle, DisruptionIndexView }` at the bottom, I think strongly implies that those are the only exported functions. Conversely, if you see a `export const getRouteIcon = ...` here but then a `const DisruptionIndex = ...` there, the implication is that the latter is not public.

I don't know if there's an easy linter setting we can enable to enforce a consistent style. If so, let's do that, with whichever style. If not, can you update the files you've touched to go with your approach? (Remove the `export { ... }` at the bottom, and just tag those particular functions with `export` at the definition?

fjlanasa:
Sorry, I just didn’t think about it. I’m fine with either way, so I’ll just
update to match the current approach.

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: i just did this without thinking about it. i'm fine with either way so will just change the ones touched here to match your preferred style.

fjlanasa: i'll make a follow up ticket to look into adding this rule: https://github.com/benmosher/eslint-plugin-import

ingridpierre:
Using the same style for both the active selection and the focus/hover states makes the filter appear as though it's still selected when it isn't. 

For example, here I've clicked this button on and then **off**, but it still looks active since the element is in focus.
image

Instead let's separate the following treatments:

**Styling for the .active.btn-outline-secondary (with or without focus):**
Grey outline, grey background
image

**Styling for the focus state on the btn-outline-secondary buttons that are not active:**
black outline, white background
image

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: https://app.asana.com/0/881264583703207/1194650366902593

fjlanasa:
> Using the same style for both the active selection and the focus/hover states makes the filter appear as though it's still selected when it isn't.
> 
> For example, here I've clicked this button on and then **off**, but it still looks active since the element is in focus.
> image
> 
> Instead let's separate the following treatments:
> 
> **Styling for only .active.btn-outline-secondary (with or without focus):**
> Grey outline, grey background
> image
> 
> **Styling for the inactive focus state on the .btn-outline-secondary buttons:**
> black outline, white background
> image

made this change and am redeploying to dev

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -151,146 +147,254 @@ const RouteFilterToggleGroup = ({
...
+  group: G[]
+): {
+  state: { [filter in G]?: boolean }
+  anyActive: boolean
+  isFilterActive: (filter: G) => boolean
+  toggleFilter: (filter: G) => void
+  clearFilters: () => void
+  updateFiltersState: React.Dispatch<{ [filter in G]?: boolean }>
+} => {
+  const [filtersState, updateFiltersState] = React.useState<
```
losvedir, line 160: Correct me if I'm wrong, but doesn't this violate the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html)? I didn't think you were "allowed" to call `useState` from a function that wasn't a React Component.

losvedir: Ah, nevermind: I missed that this was `useXYZ` and a custom hook.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -151,146 +147,254 @@ const RouteFilterToggleGroup = ({
...
-              adj.routeId &&
-              (routeFilters[adj.routeId as Routes] ||
-                (routeFilters.Commuter && adj.routeId.includes("CR-")))
-          )) &&
-        (x.adjustments || []).some(
-          (adj) =>
-            adj.sourceLabel && adj.sourceLabel.toLowerCase().includes(query)
+              adj.sourceLabel && adj.sourceLabel.toLowerCase().includes(query)
+          ) &&
+          revision.isActive
```
losvedir, line 263: If I'm understanding this correctly, doesn't this mean deleted revisions just won't appear in the table? I think we need draft revisions to appear there, so they can be selected and made ready. Otherwise, if we delete a disruption, I think it disappears, but then we have no way to approve it, right? Maybe I'm misunderstanding the flow here.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -151,146 +147,254 @@ const RouteFilterToggleGroup = ({
...
+              ]}
+              toggleRouteFilterState={routeFilters.toggleFilter}
+              isRouteActive={routeFilters.isFilterActive}
             />
-            {anyRouteFiltersActive && (
-              <a
-                className="m-disruption-index__clear_filter_button"
+            <div>
+              <SecondaryButton
+                id="status-filter-toggle-published-ready"
```
losvedir, line 321: Maybe this is out of date but I had thought a good rule of thumb was to never use `id`s with elements. I suppose that's mostly for styling reasons, though, and I see that this is only used for the test. In the past, I've used `data` attributes to tag and query from in tests. That makes it clear what it's used for. But I suppose if the convention is that's the only thing `id`'s are used for, then that works, too. Do you think we'd use `id` for anything else?

assets/src/disruptions/disruptionTable.tsx
```diff
@@ -94,47 +145,128 @@ const DisruptionTable = ({ disruptionRevisions }: DisruptionTableProps) => {
...
+        return "needs review"
+      }
+      case DisruptionView.Ready: {
+        return "ready"
+      }
+      case DisruptionView.Published: {
+        return "published"
+      }
+    }
+  }, [])
```
losvedir, line 160: Should `status` be a dependency? Or is that unnecessary because it's a parameter of the function?

In any case, I'm not sure this needs to be a react hook at all. I'd think we'd want to leverage react's state management if it's an expensive call, but a switch statement is pretty light weight.

assets/src/disruptions/disruptionTable.tsx
```diff
@@ -94,47 +145,128 @@ const DisruptionTable = ({ disruptionRevisions }: DisruptionTableProps) => {
...
+          />
         </tr>
       </thead>
       <tbody>
-        {sortedDisruptions.map((x) => (
-          <tr key={x.disruptionId}>
-            <td>{x.label}</td>
+        {sortedDisruptions.map((x, i, self) => (
+          <tr
+            key={`${x.id}${i}`}
```
losvedir, line 213: Maybe `${x.id}-${i}`? Otherwise I wonder if there's a scenario where there's a collision between, `10-1` and `1-10`.

losvedir: Er, bad example, maybe `11-1` and `1-11`. (i.e. ID 11 at index 1, vs. ID 1 at index 11).

assets/src/models/disruption.ts
```diff
@@ -7,10 +7,10 @@ import {
 } from "../jsonApi"
 
 import DisruptionRevision from "./disruptionRevision"
+import { DisruptionView } from "../disruptions/viewToggle"
```
losvedir, line 10: Pulling in stuff from a view into the model is a code smell to me. What do you think about inverting this relationship? Defining the status enum here in the model (it seems natural to me to think of the status as a property of the model), and then the view importing it, instead of the other way around.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -151,146 +147,254 @@ const RouteFilterToggleGroup = ({
...
-    return disruptionRevisions.filter((x) => {
-      return (
-        (!anyRouteFiltersActive ||
-          (x.adjustments || []).some(
+    return disruptions.reduce((acc, curr) => {
+      const uniqueRevisions = [
+        revisionFromDisruptionForView(curr, DisruptionView.Published),
+        revisionFromDisruptionForView(curr, DisruptionView.Ready),
+        revisionFromDisruptionForView(curr, DisruptionView.Draft),
+        ,
```
lemald, line 236: I think this line might be unnecessary? Unless there's some TSX syntax I don't know about at play here, which wouldn't surprise me TBH.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -151,146 +147,254 @@ const RouteFilterToggleGroup = ({
...
+                "Red",
+                "Blue",
+                "Orange",
+                "Green-B",
+                "Green-C",
+                "Green-D",
+                "Green-E",
+                "Mattapan",
+                "Commuter",
+              ]}
```
lemald, line 315: Might be worth putting this list of routes into a constant somewhere since I see it referenced in a couple of places.

fjlanasa: it's referenced above as a literal type and here as an actual value. I'd have to think about how to reuse one or the other. don't know off the top of my head.

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: yeah idk where that came from, maybe the formatter?

assets/src/disruptions/disruptionTable.tsx
fjlanasa: good point. made that change.

assets/src/models/disruption.ts
fjlanasa: yeah, good point. i'll move to the model.

assets/src/disruptions/disruptionIndex.tsx
lemald: In any event looks like it's only used in a couple of places so it's not the end of the world if you can't find a good way to reuse the type definition as a constant or whatever.

assets/src/disruptions/disruptionTable.tsx
fjlanasa: ok i can remove the `useCallback` piece here.

fjlanasa: to be clear, the value here is that the function is not redefined on every render, not that the calculation is cached

fjlanasa: but you're right that it can be defined at the module level instead, since there is no dependency

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: I generally try not to use them for styling, and I'm pretty sure the only reason they are used in this codebase is for query in tests. I guess you could use it to query elements within the application, but that's not a super common usecase in React. not opposed to refactoring to use data attributes, but would end up touching almost all the tests.

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: yeah, pushed up a change to make this last clause `(revision.isActive || revision.status !== DisruptionView.Published)`

assets/src/disruptions/disruptionIndex.tsx
losvedir: Yeah, I guess it's fine. I think it's just using them for styling we should avoid, which we're doing.

losvedir (approved):
🚀

PR #429, opened 2020-09-21
fix: sort adjustments for equality comparison

fjlanasa:
#### Summary of changes
**Asana Ticket:** [[extra] sort Arrow adjustment labels in table view](https://app.asana.com/0/584764604969369/1194957087550849)

In the attached screenshot, you'll see two separate rows where there should be a rowlette. This is because the adjustments were saved in different orders for each revision, for whatever reason. This just fixes the logic to sort the adjustments before [comparing whether labels are equal](https://github.com/mbta/arrow/blob/fl/fix-adjustments-sorting/assets/src/disruptions/disruptionTable.tsx#L222), and thus whether we should show a rowlette or new labels.

Screen Shot 2020-09-21 at 2 22 25 PM

#### Reviewer Checklist


losvedir (approved):
Good catch, thanks! 👍

PR #432, opened 2020-09-23
chore: remove "published/ready" status filter

fjlanasa:
This just deletes the published/ready filter. The "needs review" and other filters will continue work as before. 

#### Reviewer Checklist


losvedir (approved):
This looks fine. 👍 

As part of the process of cleaning up the behavior of the filters, I think we also mentioned that on the Calendar view the data should be the published data. Does Arrow already do that?

fjlanasa:
> This looks fine. 👍
> 
> As part of the process of cleaning up the behavior of the filters, I think we also mentioned that on the Calendar view the data should be the published data. Does Arrow already do that?

Yes, this was introduced in the original PR.

PR #433, opened 2020-09-24
feat: add DiffCell component to show difference between two values in DisruptionTable

fjlanasa:
#### Summary of changes
**Asana Ticket:** [🏹 Diff change between rows](https://app.asana.com/0/584764604969369/1191956107867608/f)

Changes: 
* Adds a component `DiffCell` that takes a `baseValue` and `currentValue` (both `string | number | (string|number)[]`) and grays out the text (using a Bootstrap class `text-muted`)  if `baseValue != null &&  baseValue !== currentValue`, checking for deep equality in the case of arrays.
* Wrap the values of the `date range`, `excep`, and `time period` columns with `DiffCell` in the table
* Added a component `DisruptionTableRow` to extract this logic out from the `DisruptionTable` component 
* small refactor to consolidate the sorting of the various `DisruptionRevision` relations (`exceptions`, `daysOfWeek`, `adjustments`) in `DisruptionRevision.fromJsonObject`

Screen Shot 2020-09-23 at 6 51 59 PM

#### Reviewer Checklist


assets/src/diffCell.tsx
```diff
@@ -0,0 +1,28 @@
...
+type Value = string | number | (string | number)[]
+
+interface DiffCellProps<T extends Value[] | Value = string> {
+  currentValue: T
+  baseValue?: T
+  element?: "td" | "div" | "span"
+  children: React.ReactNode | string
+}
+
+const DiffCell = <T extends Value[] | Value = string>({
```
losvedir, line 12: I'm kind of going back and forth on whether I think it's worth it for this to be a higher order component or not. Ultimately, all it does is calculate whether to apply a `text-muted` class, but it feels to me like it has a lot of syntactic and semantic baggage to accomplish that.

What if `isDiff` were broken out as a simple helper function that takes two pieces of data and returns a boolean? Then the callers would be able to specify `<td>` or what have you, and would just themselves do `classname={isDiff(val1, val2) ? "" : "text-muted"`? It feels simpler, at the expense of making `text-muted` no longer an implementation detail.

Just curious if you started with the simple helper function approach, or what your thought process was on going to HOF.

fjlanasa: Yeah, I started by just adding `className={base != null && base != current ? "" : "text-muted"` in the few places where I'm now using `DiffCell`. My only motivation was to make `text-muted` an implementation detail, but I agree that it might be overkill. Could go either way.

fjlanasa: Though I suppose it's not really an implementation detail if it's in the tests https://github.com/mbta/arrow/blob/fl/select/assets/tests/disruptions/disruptionTable.test.tsx#L187

losvedir: Cool, yeah, then in that case could you switch to just a plain function? I have a slight preference towards simpler / more straightforward code over higher level abstractions when it's kind of a toss up, though I agree that either is fine.

losvedir (approved):
👍

PR #440, opened 2020-09-29
feat: Ability to mark multiple disruptions as "ready"

fjlanasa:
#### Summary of changes
**Asana Ticket:** [🏹 Ability to mark multiple disruptions as "ready"](https://app.asana.com/0/584764604969369/1191956107867616/f)

Backend changes:
* Refactor to move `publish_notice_controller(_test)` to `notice_controller(_test`) with `publish` and `ready` actions
* Add `DisruptionRevision.ready!` function to take a list of revision ids and mark all as ready, to be called in `NoticeController.ready`

Frontend changes:
* Add state management and checkboxes to handle selecting multiple draft revisions 
* Add logic to hit `ready` endpoint with list of revision ids to mark as ready, refetching disruptions on success

ready

#### Reviewer Checklist


assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -280,6 +288,92 @@ const DisruptionIndexView = ({ disruptions }: DisruptionIndexProps) => {
     }, [] as DisruptionRevision[])
   }, [disruptions, searchQuery, routeFilters, statusFilters])
 
+  const [selectedRevisionsState, setSelectedRevisionsState] = React.useState<{
```
losvedir, line 291: Maybe the variable names should be `selectedRevisions` and `setSelectedRevisions`? At least, until now it seems we've been treating "state" as implied when using `useState`, it seems.

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -280,6 +288,92 @@ const DisruptionIndexView = ({ disruptions }: DisruptionIndexProps) => {
...
+        successParser: () => true,
+        errorParser: () => true,
+        url: "/api/ready_notice/",
+      })
+        .then(() => {
+          fetchDisruptions()
+        })
+        .catch(() => {
+          // eslint-disable-next-line no-console
+          console.log("error")
```
losvedir, line 336: Can we log something more useful here?

ingridpierre:
I think we ought to have the 'actions' button be disabled on the calendar view, similar to the status filter.
image

ingridpierre:
When the actions button is selected and the bar is visible/expanded it should get the darker "active" button styling similar to :
image

assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -346,22 +435,78 @@ const DisruptionIndexView = ({ disruptions }: DisruptionIndexProps) => {
...
+                  (view === "calendar" ? "list view" : "calendar view")}
+              </SecondaryButton>
+            </div>
           </div>
         </Col>
       </Row>
+      {actionsMenuOpen && (
+        <Row>
+          <Col>
+            <div className="d-flex p-2 mb-3 border-secondary border rounded">
```
losvedir, line 472: Whew. I'm a big fan of utility classes (we should just use tailwind!), but so far I don't think we've been following this approach so much. When do you think about using a collection of utility classes vs. pulling it into a single named ("semantic") class?

lib/arrow_web/controllers/api/notice_controller.ex
```diff
@@ -0,0 +1,48 @@
+defmodule ArrowWeb.API.NoticeController do
+  use ArrowWeb, :controller
+  require Logger
+
+  @spec publish(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def publish(conn, params) do
+    try do
+      revision_ids =
+        params["revision_ids"]
```
losvedir, line 9: Nice to pull these two actions into the same controller. Since both actions do this same parsing, maybe pull into a little private helper function?

losvedir (approved):
👍

assets/src/disruptions/disruptionIndex.tsx
fjlanasa: I think we do in fact use a fair amount of utility classes. I generally use utility classes if the desired effect can be achieved with them and the combination isn't reused (like this case).

PR #457, opened 2020-10-05
refactor: add lint rule to declare module exports at end of file

fjlanasa:
#### Summary of changes
**Asana Ticket:** [💳 Add eslint-plugin-import](https://app.asana.com/0/584764604969369/1194650366902593/f)

Added module `eslint-plugin-import` and added two rules:
* Ensure all exports appear after other statements
* Prefer named exports to be grouped together in a single export declaration

Note I did **not** add a rule forbidding `default` exports, which is another possibility, though would require changing many files. 

#### Reviewer Checklist

### DIFF
--- a/assets/.eslintrc.js
+++ b/assets/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = {
     "@typescript-eslint/tslint",
     "prefer-arrow",
     "react-hooks",
+    "import",
   ],
   rules: {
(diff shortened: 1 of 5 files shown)


losvedir (approved):
👍 . I don't feel strongly about `default` or not. I lean towards linting against them for consistency, but for smaller files that export just a single component, I do think it makes sense there. So... shrug.

PR #517, opened 2020-11-13
fix: Update disruptions that aren't ready

losvedir:
#### Summary of changes
**Asana Ticket:** [🐛 Unable to edit disruption in Arrow](https://app.asana.com/0/281253957434160/1198920748926221)

The fix is pretty small. Basically, the `update` controller action used `DisruptionRevision.only_ready()` to select the latest "ready" revision to clone when updating. However, I think actually we should be cloning the latest revision. The frontend flow lets you edit only based off the latest draft. And the current flow causes a crash if a given disruption doesn't have a ready revision yet, but is updated (i.e.: two drafts in a row before any approvals).

#### Reviewer Checklist

### DIFF
--- a/test/arrow_web/controllers/api/disruption_controller_test.exs
+++ b/test/arrow_web/controllers/api/disruption_controller_test.exs
@@ -533,6 +533,63 @@ defmodule ArrowWeb.API.DisruptionControllerTest do
     end
 
     @tag :authenticated
+    test "can update a disruption that's never been 'ready'", %{conn: conn} do
+      adjustment =
+        insert(:adjustment, %{
+          source_label: "test_adjustment_1",
+          route_id: "test_route_1",
+          source: "arrow"
+        })
+
+      disruption = insert(:disruption)
+
+      disruption_revision =
+        insert(:disruption_revision,
+          disruption: disruption,
+          start_date: ~D[2019-10-10],
+          end_date: ~D[2019-11-01],
+          days_of_week: [
+            build(:day_of_week, day_name: DayOfWeek.date_to_day_name(~D[2019-10-10]))
+          ],
+          adjustments: [adjustment]
+        )
+
+      post_data = %{
+        "data" => %{
+          "type" => "disruption",
+          "id" => disruption.id,
+          "attributes" => %{
+            "end_date" => Date.to_iso8601(~D[2019-12-01])
+          }
+        }
+      }
+
+      conn = patch(conn, "/api/disruptions/" <> Integer.to_string(disruption.id), post_data)
+
+      assert %{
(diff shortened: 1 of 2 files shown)


test/arrow_web/controllers/api/disruption_controller_test.exs
```diff
@@ -532,6 +532,104 @@ defmodule ArrowWeb.API.DisruptionControllerTest do
...
+      disruption_revision =
+        insert(:disruption_revision,
+          disruption: disruption,
+          start_date: ~D[2019-10-10],
+          end_date: ~D[2019-11-01],
+          exceptions: [build(:exception, excluded_date: ~D[2019-10-20])],
+          days_of_week: [
+            build(:day_of_week,
+              day_name: DayOfWeek.date_to_day_name(~D[2019-10-10]),
+              start_time: ~T[20:30:00]
```
digitalcora, line 555: This test is a bit hard to follow (100 lines!) — I think partly because it includes a ton of details that ultimately aren't relevant to the behavior being tested. Maybe we can just pick one attribute to update for this test, like the end date, and strip out everything else that isn't strictly required? Both from the factory calls, and the post data.

losvedir: Good idea. I stripped out some of the unnecessary data (and, added a bit more testing, so the actual lines of code is the same or longer, but more relevant).

test/arrow_web/controllers/api/disruption_controller_test.exs
```diff
@@ -532,6 +532,80 @@ defmodule ArrowWeb.API.DisruptionControllerTest do
...
+        "data" => %{
+          "type" => "disruption",
+          "id" => disruption.id,
+          "attributes" => %{
+            "start_date" => Date.to_iso8601(disruption_revision.start_date),
+            "end_date" => Date.to_iso8601(~D[2019-12-01])
+          },
+          "relationships" => %{
+            "days_of_week" => %{
+              "data" => [%{"type" => "day_of_week", "attributes" => %{"day_name" => "saturday"}}]
```
digitalcora, line 567: Do we need to include all the existing (unchanged) attributes in the update? I don't think we _should_ need to, based on how Ecto updates typically work, and that could improve the readability a bit.

losvedir: Huh, I had thought if we didn't include them that would tell it to *delete* the associations, but from some playing around it looks like that's only if you include the relationship but set `"data" => []` which makes sense. I removed the extraneous attributes.

digitalcora (approved):
⏱️✨

PR #551, opened 2020-12-16
fix: duplicate events in calendar view

digitalcora:
**Asana Ticket:** [🐛 Arrow: multiple revisions of disruption show up simultaneously in calendar view](https://app.asana.com/0/584764604969369/1199340421679669)

If a disruption had a ready and/or draft revision, each of those would display as its own "event" on the calendar along with the published one. This was due to how the filtering logic used in the table view, which always displays all revisions of "matched" disruptions, was carried over to the calendar view.

#### Reviewer Checklist

### DIFF
--- a/assets/src/disruptions/disruptionIndex.tsx
+++ b/assets/src/disruptions/disruptionIndex.tsx
@@ -157,40 +157,35 @@ type FilterGroup<G extends string> = {
   updateFiltersState: React.Dispatch<FilterGroupState<G>>
 }
 
-const anyMatchesFilter = (
-  revisions: DisruptionRevision[],
+const revisionMatchesFilters = (
+  revision: DisruptionRevision,
   query: string,
   routeFilters: FilterGroup<Routes>,
   statusFilters: FilterGroup<"published" | "ready" | "needs_review">,
   dateFilters: FilterGroup<"include_past">,
   pastThreshold: Date
-) => {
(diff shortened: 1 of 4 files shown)


assets/src/disruptions/disruptionIndex.tsx
```diff
@@ -287,18 +282,27 @@ const DisruptionIndexView = ({
...
         )
-      ) {
-        return [...acc, ...uniqueRevisions]
+      )
+
+      // The table view displays *all* revisions of disruptions which have at
+      // least one revision that matches the filters. The calendar view only
+      // displays the matching revisions themselves, otherwise there would be
+      // duplicate entries.
+      if (matchingRevisions.length) {
```
losvedir, line 300: Not really necessary, but I generally prefer `=== 0` here rather than relying on the "falsey-ness" of `0` in JS, since it's a bit of gotcha considering how heavily we use Elixir where `0` is truthy instead.

losvedir (approved):
Just the one small suggestion, but otherwise looks good to me, and pulling it down and running locally it seemed to fix the bug. 👍

assets/src/disruptions/disruptionIndex.tsx
digitalcora: Sure, I did it this way since it seemed to be the convention in the existing code, but being more explicit may be better.

PR #556, opened 2020-12-22
chore: login to Docker Hub in CI

digitalcora:
To unblock deploys.


paulswartz (changes requested):
You also need to remove the `-x` from line 2, to avoid logging the password.

digitalcora:
@paulswartz oop, fixed!

paulswartz (approved):
🍰

PR #557, opened 2020-12-22
fix: "unbound variable" error in build script

digitalcora:
I really should have test-deployed the last branch to dev _before_ merging it 😑 This would also go away once the Docker Hub variables are set up for this project, but in any case the script should work if they're not present.

### DIFF
--- a/semaphore/build_push.sh
+++ b/semaphore/build_push.sh
@@ -7,6 +7,9 @@ set -e -u
 # * DOCKER_REPO
 awsenv=$1
 
+DOCKER_USERNAME=$(printenv DOCKER_USERNAME || echo "")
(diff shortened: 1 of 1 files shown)


paulswartz (approved):
The variables are set up now: I'll defer to you if you still want to merge this.

digitalcora:
I did think the intent of the script was to work regardless of whether the variables were set, since we bothered to test for this with the `if [ -n "$DOCKER_USERNAME" ]`. So I'll counter-defer to you, as the original author of that test 😄 But, since providing the credentials is sort-of required (or deploys will randomly fail), I think it would also be fine if the script required them.

paulswartz:
That was the intention, but it was probably in a script which didn't have `-u` set. I think this is fine :)

PR #568, opened 2021-01-05
fix: only bump published date if revision changes

digitalcora:
`gtfs_creator` always sends the revision IDs of all disruptions it is publishing; Arrow in turn would always update the "published" timestamp on all the corresponding disruptions, even if their published revisions didn't change. This made the timestamp not that useful, since it would always just be the timestamp of the last GTFS deploy for all (published) disruptions.

This changes the logic to only update the "published" timestamp when the published revision changes.

**Asana Ticket:** [🐛 arrow "published" keeping last publish date without new info](https://app.asana.com/0/584764604969369/1199519213885301)

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/disruption_revision.ex
+++ b/lib/arrow/disruption_revision.ex
@@ -2,10 +2,8 @@ defmodule Arrow.DisruptionRevision do
   use Ecto.Schema
   import Ecto.Query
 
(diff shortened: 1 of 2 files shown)


lemald (approved):
👍

PR #680, opened 2021-06-04
Various backend updates

losvedir:
#### Summary of changes
**Asana Ticket:** dependabot

This updates all the open backend dependabot PRs. In so doing, I realized that our credo configuration was not actually activated... and by activating it, there were lots of things to fix up. And in doing that, I came across the old `diff` functionality that we no longer use, so I took that out.

#### Reviewer Checklist


mix.exs
```diff
@@ -60,6 +60,7 @@ defmodule Arrow.MixProject do
       {:phoenix_ecto, "~> 4.0"},
       {:dialyxir, "~> 1.0.0-rc.7", only: [:dev], runtime: false},
       {:ecto_sql, "~> 3.1"},
+      {:jose, "~> 1.11"},
```
digitalcora, line 63: Any reason we're adding this as an explicit dependency? Do we use it ourselves directly?

mix.exs
```diff
@@ -53,7 +53,7 @@ defmodule Arrow.MixProject do
       {:ex_aws_rds, "~> 2.0"},
       {:ex_aws_secretsmanager, "~> 2.0"},
       {:ex_machina, "~> 2.3", only: :test},
-      {:excoveralls, "~> 0.10", only: :test},
+      {:lcov_ex, "~> 0.2", only: [:dev, :test]},
```
digitalcora, line 56: Why `dev`?

.credo.exs
```diff
@@ -3,13 +3,13 @@
     %{
       name: "default",
       files: %{
-        included: ["apps/"],
+        included: ["lib"],
```
digitalcora, line 6: Should we include `test` as well? That seems to be [the default](https://github.com/rrrene/credo/blob/master/.credo.exs) (it also uses trailing slashes on directory names, though not sure if that matters).

.credo.exs
```diff
@@ -3,13 +3,13 @@
...
         excluded: []
       },
       checks: [
         {Credo.Check.Design.AliasUsage, false},
         {Credo.Check.Readability.MaxLineLength, false},
-        {Credo.Check.Consistency.ExceptionNames},
+        {Credo.Check.Consistency.ExceptionNames, false},
```
digitalcora, line 12: What was the warning that prompted this?

.credo.exs
```diff
@@ -42,7 +42,8 @@
         {Credo.Check.Refactor.NegatedConditionsInUnless},
         {Credo.Check.Refactor.NegatedConditionsWithElse},
         {Credo.Check.Refactor.Nesting},
-        {Credo.Check.Refactor.PipeChainStart},
+        # too many false positives with Ecto's `from` macro
+        {Credo.Check.Refactor.PipeChainStart, false},
```
digitalcora, line 46: It looks like this is actually disabled anyway in the current default config, and moved to a "controversial and experimental" section. Maybe we should regenerate our config to pick up the latest defaults?

lib/arrow/disruption/day_of_week.ex
```diff
@@ -1,4 +1,8 @@
 defmodule Arrow.Disruption.DayOfWeek do
+  @moduledoc """
+  The day of the week that disruption takes place.
```
digitalcora, line 3: that `a` disruption takes place?

lib/arrow/disruption/published_after_ready_error.ex
```diff
@@ -11,6 +16,11 @@ defmodule Arrow.Disruption.Error do
   end
 
   defmodule ReadyNotLatest do
+    @moduledoc """
+    An error representing an attempt to set the ready_revision_id higher even than the
```
digitalcora, line 20: extra word? "`even`"

lib/arrow/disruption/trip_short_name.ex
```diff
@@ -1,4 +1,8 @@
 defmodule Arrow.Disruption.TripShortName do
+  @moduledoc """
+  The Commuter Rail train numbers the disruption applies to.
```
digitalcora, line 3: Given the module is `TripShortName` and it could, in principle, refer to any trip short name (not just Commuter Rail), maybe this could be described as: "The `trip_short_name` of a specific trip a disruption applies to. Normally only used for Commuter Rail trips."

lib/arrow/disruption_revision.ex
```diff
@@ -1,4 +1,9 @@
 defmodule Arrow.DisruptionRevision do
+  @moduledoc """
+  A particular version of a Disruption, in that Disruption's creation/edit/deletion
+  life cycle.
```
digitalcora, line 4: Are we hard-wrapping these at 80 characters? It should be 98, for Elixir code.

lib/arrow_web/controllers/api/disruption_controller.ex
```diff
@@ -50,7 +50,7 @@ defmodule ArrowWeb.API.DisruptionController do
       |> Enum.map(& &1["source_label"])
 
     adjustments =
-      Repo.all(from adj in Arrow.Adjustment, where: adj.source_label in ^adjustment_labels)
+      Repo.all(from(adj in Arrow.Adjustment, where: adj.source_label in ^adjustment_labels))
```
digitalcora, line 53: Hm... not sure this is really more readable. Can we exclude `from` for this check?

lib/arrow_web/utilities.ex
```diff
@@ -35,7 +39,11 @@ defmodule ArrowWeb.Utilities do
       %{},
       fn {relationship, %{"data" => data}}, rels ->
         if is_list(data) do
-          Map.put(rels, relationship, Enum.map(data, &JaSerializer.Params.to_attributes/1))
+          Map.put(
+            rels,
+            relationship,
+            Enum.map(data, &JaSerializer.Params.to_attributes/1)
+          )
```
digitalcora, line 46: Mistaken wrapping? This fits on a 98-character line.

lib/arrow/disruption.ex
```diff
@@ -22,9 +22,9 @@ defmodule Arrow.Disruption do
         }
 
   schema "disruptions" do
-    belongs_to :ready_revision, DisruptionRevision
-    belongs_to :published_revision, DisruptionRevision
-    has_many :revisions, DisruptionRevision
+    belongs_to(:ready_revision, DisruptionRevision)
+    belongs_to(:published_revision, DisruptionRevision)
+    has_many(:revisions, DisruptionRevision)
```
digitalcora, line 27: This seems like it belongs in the previous commit.

mix.exs
losvedir: We don't use it directly, but the version of `:jose` we're on prevents us from updating to Erlang 24.

mix.exs
losvedir: Hmm 🤔 I guess no reason. My default is `[:dev, :test]` for things that don't go into production, but you're right that code coverage is specifically about testing...

mix.exs
digitalcora: In that case we should be able to `mix deps.update` it, right? i.e. just updating it in the lockfile, not adding a dependency on it to our app.

.credo.exs
losvedir: We have two exceptions: `Arrow.Disruption.Error.PublishedAfterReady` and `Arrow.Disruption.Error.ReadyNotLatest`. The error was:

> Exception modules should be named consistently. It seems your strategy is to prefix them with `Published`, but `Arrow.Disruption.Error.ReadyNotLatest` does not follow that convention."

I thought that was a bit silly. But rather than disabling this I've now just changed them to `ErrorPublishedAfterReady` and `ErrorReadyNotLatest` (moving them out of the `.Error.` namespace), and that quieted credo.

.credo.exs
losvedir: Regenerated.

lib/arrow/disruption/trip_short_name.ex
losvedir: Do other modes have trip short names? Like, even in principle, I'm not sure exactly what that would mean for subway, bus, or ferry. Can you give an example of what it might be like?

.credo.exs
digitalcora: `PublishedAfterReadyError` and `ReadyNotLatestError` would probably be more "conventional" (Elixir has things like `KeyError`, `BadMapError`, etc.), and to me at least, read a bit better.

lib/arrow/disruption/trip_short_name.ex
digitalcora: They don't right now, in our GTFS, but `trip_short_name` is not a mode-specific concept in general. Mostly I think it makes sense to explicitly draw the connection here to the GTFS concept this module represents, rather than saying it represents a "commuter rail train number" and leaving it as an exercise for the reader to figure out what that has to do with "trip short names". But I have probably already spent more words defending this than it deserves 😆

lib/arrow/disruption.ex
losvedir: It seems like this isn't an issue in the regenerated credo config, so I'll just revert it.

lib/arrow/disruption/trip_short_name.ex
losvedir: I guess that's fair. :) I had it that way because I was tripped up at first by the idea of `trip_short_name`s until I realized that they were just Commuter Rail train numbers, so that's why I focused on that. But I'll just put both bits of information in there.

.credo.exs
losvedir: Oh, I agree. I didn't think credo would understand that, but it turns out that appeases credo as well. It will look for both a shared prefix or suffix, I suppose. But if it doesn't find one, it will assume prefix for the purpose of the error message, I guess.

digitalcora: The behavior can be a little weird if you only define a few exceptions in your codebase and they're not perfectly consistent. On T-Alerts we had a total of 2 exceptions defined, so it had to pick one to break the tie and ended up with "It looks like your strategy is to prefix them with `Mailer`" 😅

mix.exs
losvedir: That would update the dependency, yeah. But to me, I guess I was thinking about it as wanting to express a version constraint. I think if you simply update it in the lock file, the resolver could in the future potentially downgrade it, right?

digitalcora: It could be downgraded if we added a package that required an older version (the requirement would have to be something like `~> 1.0 and < 1.11`), but then we'd discover it was broken at that time and have to deal with it somehow anyway. Unless the breakage is subtle?

If we do keep this here I'd prefer adding an explanatory comment like "dependency of `<package>`, minimum version for OTP24 support".

losvedir: Hmm, I suppose the breakage is not all that subtle. If something downgrades it, the app will just fail to compile. I'll go ahead with simply updating the lock file.

losvedir:
Oof, after rebasing on `master`, I see that we're using `mix credo --strict` which found some more, lower priority errors, which I've fixed up now, too.

digitalcora (approved):
⬆️ ⬆️ ⬆️ ✅

PR #724, opened 2021-07-29
fix: calendar "today" highlight was in UTC

digitalcora:
**Asana Ticket:** [🏹 🐛 Arrow calendar "today" is on UTC time](https://app.asana.com/0/584764604969369/1200312313842498)

The start and end date of a disruption are ultimately just dates, with no time or zone information, but since JavaScript lacks a "plain date" type we parse them on the frontend as `Date` objects with the time part set to 00:00:00 in UTC. To make FullCalendar display these correctly, we set the calendar's `timeZone` to UTC. This had the side effect that the "today" highlight followed the current UTC time, causing "today" to advance to the next day too early (from our time zone's perspective).

FullCalendar also accepts event start/end values as ISO date strings, which are assumed to be in the browser's local time zone. By doing so and setting the calendar to use the `local` time zone (the default), we can avoid dealing with times in the calendar at all, and "today" should align with local time.

#### Reviewer Checklist


digitalcora:
Note: I smoke-tested this by syncing my local Arrow database with production, going to the calendar view in both environments, and flipping back and forth between them. Everything looked identical. Additionally, we already have a test for the calendar component that verifies events are displayed correctly around DST changes.

bklebe (approved):

PR #725, opened 2021-07-31
refactor: No more react-router, routing by Phoenix

losvedir:
#### Summary of changes
**Asana Ticket:** [Move page header and routing to Phoenix](https://app.asana.com/0/584764604969369/1200687441670226)

This removes `react-router` completely and transitions all routing to Phoenix, as prep work for de-SPA-ifying the app. The app otherwise is mostly react: the various routes just load a react component, which then make API calls to get the data for the page.

There are a few parts to it:

* The header is moved to `app.html.eex` and the `<Page />` component was deleted. I made the logo/title clickable to go home, since I think that's pretty well understood and expected behavior at this point. (This also means the "back to home" link is no longer necessary.)
* Previously most requests were sent to `PageController.index/2` which boots up the react app and uses react router to determine the component to display. Now, the requests are fielded by the `DisruptionController`, and each of index/show/new/edit views, which embed their respective react components using `react_phoenix`.
* Within the components, navigation was handled by the all-pervasive react router via `<Link>`, `<Redirect>`, and the like. These were replaced with simple `window.location` manipulation to issue a new request instead. I pulled this logic into a little `redirectTo` function in a helper module in order to mock it with `jest` in tests. Raw manipulation of `window.location` is prohibited in the jsdom and was causing errors.
* The tests were the most annoying part. Because of the `<Link>`s previously, the components had to be embedded in `<BrowserRouter>` or `<MemoryRouter>` components in the test. For the most part, this was unnecessary ceremony once removing the `<Link>`s, except for the case of needing to know the current URL to get the `?v=ready` or `?v=published` parameter, to know which revision to show (since currently the frontend just gets *all* the revisions and decides on the frontend which one to show). I replaced the code's use of react-router's `useHistory` with a simple static `history` and used `window.history.pushState` in tests to set the proper URL.
* Some tests were for things like clicking on a link and navigating, which aren't really the frontend's responsibility anymore, and can't be tested besides (until we set up integration tests). I deleted these. We'll be flying blind for a bit that, e.g., the table on the home page can link to a disruption's show page, but I did test it manually and it works.

#### Reviewer Checklist


assets/webpack.config.js
```diff
@@ -5,6 +5,10 @@ const CopyWebpackPlugin = require("copy-webpack-plugin")
 
 module.exports = (env, options) => ({
   resolve: {
+    alias: {
+      react: path.resolve(__dirname, './node_modules/react'),
+      'react-dom': path.resolve(__dirname, './node_modules/react-dom')
+    },
```
bklebe, line 11: Why is this alias needed?

bklebe: Interesting. Alright, thanks for making that clearer.

bklebe (approved):

PR #729, opened 2021-08-04
New "/feed" page to see changes relative to what's in GTFS

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 Page to review recent changes](https://app.asana.com/0/584764604969369/1200671066645718)

This branch is (at the time of this comment) on https://arrow-dev.mbtace.com/feed. (But note that on Dev we don't mark revisions as "published", and so the feed is basically all disruptions!)

A few tweaks to lay the groundwork for our Arrow work. Since we're no longer having a "request" -> "approve" flow, we need a way to see the recent changes before deploying GTFS from `gtfs_creator`. This PR has a few independent commits:

1. The first commit removes a `/temp_review_changes` route which never really got created but was intended to be something like this back in the day.
2. The second commit actually implements the feed, at `/feed`. I used TailwindCSS, per our retro discussion, but just dropped a link to the CDN in the index template. I updated our upcoming fix-assets task to remove that link once we have Tailwind incorporated correctly.
3. Fixes a small corner case the `/feed` page highlighted related to deleting disruptions. Previously, we'd never move the "published" ID of a disruption to the deleted revision, since gtfs_creator never even knows about it anymore once it's deleted. However, in a sense, if a disruption was there previously and is now removed, the deletion of the disruption (i.e. the absence of the disruption in the data) is part of that published feed, so should be marked as such.

Screen Shot 2021-08-04 at 2 43 08 PM

#### Reviewer Checklist


lib/arrow/disruption.ex
```diff
@@ -146,6 +146,44 @@ defmodule Arrow.Disruption do
     {updated, new}
   end
 
+  @doc """
+  Takes a disruption query (or starts a new one for Arrow.Disruption) and adds a new named join
+  to it with a column `latest_revision_id` which is that Disruption's latest revision.
+
+  Usage: from([disruption, latest_revisions] in with_latest_revision_id(), ...)
+  """
+  def with_latest_revision_id(disruption_query \\ __MODULE__) do
```
bklebe, line 155: Is this missing a spec or does it not need one because it's just some query builder type?

digitalcora: I think it should be:

```elixir
@spec with_latest_revision_id(Ecto.Queryable.t()) :: Ecto.Query.t()
```

(also, stealing this for my current work, thanks~)

losvedir: Nope, just missing one.

bklebe (approved):

PR #735, opened 2021-08-16
test: fix calendar test failing on Mondays

digitalcora:
This test meant to create a 7-day date span, but used "today + 7 days" rather than "today + 6 days", so the end date fell on the same day of the week as the start date. Since the test disruption was active on Monday, when the current day was Monday this caused the disruption to appear in the calendar twice, failing the test.


mkennedm (approved):
Looks good.

PR #738, opened 2021-08-19
Migrate disruptions table/calendar to server-rendered

digitalcora:
**Asana Ticket:** [🏹 Migrate Arrow from SPA to Phoenix-routed](https://app.asana.com/0/584764604969369/1200671066645709) (subtask)

This is primarily a port of the `DisruptionIndex` and `DisruptionTable` components to Phoenix controller and view logic. The filters that can be applied to the index are now expressed as query params, incidentally making it possible to link to a specific filter state of the index. The `DisruptionCalendar` wrapper for `FullCalendar` remains, and now accepts a JSON:API document so the server can easily pass data to it. (I think it would be very possible to port this if we wanted to, but didn't want to expand the scope of this subtask any further)

My feeling is that this could still use some testing, but I think it may be covered by #734. I might add some controller tests for the table view specifically, since no JavaScript is involved and so a full browser isn't required for an "integration test".

#### Reviewer Checklist


lib/arrow_web/controllers/disruption_controller/filters/calendar.ex
```diff
@@ -0,0 +1,14 @@
+defmodule ArrowWeb.DisruptionController.Filters.Calendar do
+  @moduledoc "Handles filters unique to the calendar view (currently none)."
```
losvedir, line 2: I'm not sure this abstraction is pulling its weight for me. Conceptually, I think of the querying as part of the non-web context, which accepts a `Filters` as its interface (which is nice, I like that). In that context, things like what to filter and how to sort are "first class" instructions.

I often like the "make impossible states unrepresentable" approach, but in this there's no impossible state as far as what gets queried by the DB, just some collection of conditions. According to the *view*, we shouldn't bother with sorting if someone is looking at the calendar, but I think that could be handled by the `from_params/1`, and shouldn't really be a part of the *data*.

lib/arrow_web/controllers/disruption_controller/filters.ex
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+  defstruct routes: @empty_set, search: nil, view: %Table{}
+
+  @spec calendar?(%__MODULE__{}) :: boolean
+  def calendar?(%{view: %Calendar{}}), do: true
+  def calendar?(%{view: %Table{}}), do: false
+
+  @spec flatten(%__MODULE__{}) :: %{atom => any}
+  def flatten(%{view: view} = filters) do
```
losvedir, line 36: I think in general it's good with modern Elixir to pattern match on the struct (`def flatten(%__MODULE{...`), since the compiler now catches accessing struct members that aren't present and that sort of thing. I know it's in the `@spec` but.... I just don't trust dialyzer to catch those sorts of type issues all the time.

lib/arrow_web/controllers/disruption_controller/filters/helpers.ex
```diff
@@ -0,0 +1,7 @@
+defmodule ArrowWeb.DisruptionController.Filters.Helpers do
+  @moduledoc "Functions shared by the `Filters` modules."
+
+  @spec put_if(map, boolean, any, any) :: map
```
losvedir, line 4: Huh, nifty little thing.

lib/arrow_web/controllers/disruption_controller/index.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  defp apply_filter({:sort, {direction, :start_date}}, query) do
+    from [revisions: r] in query, order_by: {^direction, r.start_date}
+  end
+
+  defp apply_filter(_, query), do: query
+
+  defp apply_filters(query, nil), do: query
+
+  defp apply_filters(query, filters) do
```
losvedir, line 50: :chefs_kiss: - so cool. I love how all these filters compose!

losvedir (commented):
Overall looks great to me! A couple conceptual comments which aren't really mandatory, more something I feel like should change but am open to discussion about.

lib/arrow_web/views/disruption_view.ex
```diff
@@ -1,3 +1,75 @@
...
+
+  alias ArrowWeb.API
+  alias ArrowWeb.DisruptionController.Filters
+  alias Phoenix.Controller
+
+  defp calendar_props(disruptions) do
+    %{"data" => API.DisruptionView.render("index.json-api", data: disruptions)}
+  end
+
+  defdelegate current_path(conn), to: Controller
```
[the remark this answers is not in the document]
digitalcora: `defdelegate` was more concise, that's all. I could make them private functions if you think that's better.

lib/arrow_web/controllers/disruption_controller/index.ex
```diff
@@ -0,0 +1,71 @@
+defmodule ArrowWeb.DisruptionController.Index do
```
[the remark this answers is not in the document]
digitalcora: I'd agree this module isn't primarily "web" — it's here because it was extracted from the first pass where everything was just tossed in private functions on the controller. On the other hand, it does have some logic that's tied to the view layer (the templates assume `revisions` only contains one revision and it's the latest one, `adjustments` are sorted by name, etc.), so I'd hesitate to say it's not connected at all.

My instinct is that it's fine for it to live here given nothing else uses it; it's basically a big "private module" of the controller. If something _else_ later needed to do this same kind of "disruption indexing" in a different context, that would of course prompt some refactoring based on the needs of that other thing, and perhaps we'd end up pulling out just the view-specific logic to a different place. Right now it seems premature.

lib/arrow_web/views/disruption_view.ex
```diff
@@ -1,3 +1,75 @@
 defmodule ArrowWeb.DisruptionView do
   use ArrowWeb, :view
+
+  alias ArrowWeb.API
+  alias ArrowWeb.DisruptionController.Filters
+  alias Phoenix.Controller
+
+  defp calendar_props(disruptions) do
+    %{"data" => API.DisruptionView.render("index.json-api", data: disruptions)}
```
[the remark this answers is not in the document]
digitalcora: If we did port the logic in the `DisruptionCalendar` component to the server, we wouldn't need JSON:API for this specific case (the server would encode the disruptions directly into FullCalendar object format). As I mentioned in the OP that was a bit more than I wanted to bite off in this task, but in general I'm doing things with an eye to minimizing our use of JSON:API to pass data "to ourselves".

lib/arrow_web/controllers/disruption_controller/filters/calendar.ex
digitalcora: This was initially a "flat" data model (i.o.w. it looked more like what the output of `flatten` now looks like), but adding the calendar view ended up cluttering the code with so many "if calendar then this, else that" conditionals that I started looking for a cleaner alternative. I suppose you'll have to take my word for it, since the original work is now squashed away 😓 But it was decidedly buggy and hard-to-reason-about.

> I often like the "make impossible states unrepresentable" approach, but in this there's no impossible state as far as what gets queried by the DB, just some collection of conditions.

Well, `Filters` is the interface to `Index`, but it's also mainly a representation of the "front-end state" of the view (in fact it'd probably be a `State` module if this were LiveView). In the latter there are nonsensical states, and I think it's valuable to make them unrepresentable, but that means we have this slight tension between the two roles. The current solution is just `flatten/1`, which is the "adapter" from `Filters`-as-frontend-state to `Filters`-as-`Index`-interface. I'm open to different ideas on how to handle this interface; I think it's adequate but I don't love it.

lib/arrow_web/views/disruption_view.ex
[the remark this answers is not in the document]
losvedir: Hm, yeah I think so. Thanks!

losvedir (approved):
Thanks for addressing my questions! Looks good to me. 👍

PR #741, opened 2021-08-24
Migrate disruption details page to server-rendered

digitalcora:
**Asana Ticket:** [🏹 Migrate Arrow from SPA to Phoenix-routed](https://app.asana.com/0/584764604969369/1200671066645709) (subtask)

Since the "delete" button was part of `ViewDisruption` and was not mentioned as a separate item in the migration task, its functionality is also migrated here, with a small change: It redirects back to the disruption details page rather than the index, giving a more explicit confirmation of the action (the "This disruption is deleted" note).

Some basic controller tests are also added for the non-API disruptions controller, to ensure everything works together in the absence of full acceptance tests.

#### Reviewer Checklist


lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -19,4 +22,25 @@ defmodule ArrowWeb.DisruptionController do
   def edit(conn, %{"id" => id}) do
     render(conn, "edit.html", id: id)
   end
+
+  def delete(conn, %{"id" => id}) do
+    _revision = Disruption.delete!(id)
+    redirect(conn, to: Routes.disruption_path(conn, :show, id))
+  end
+
+  defp get!(id) do
```
losvedir, line 31: Maybe this would be good in `Disruption`? Loading a full fledged version of the current version of a disruption seems like a property of the business logic rather than just the web app to me.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -19,4 +22,25 @@ defmodule ArrowWeb.DisruptionController do
...
+    from(
+      [disruptions: d, revisions: r] in Disruption.with_latest_revisions(),
+      where: d.id == ^id,
+      left_join: a in assoc(r, :adjustments),
+      left_join: days in assoc(r, :days_of_week),
+      left_join: e in assoc(r, :exceptions),
+      left_join: t in assoc(r, :trip_short_names),
+      order_by: [a.source_label, e.excluded_date, t.trip_short_name],
+      preload: [
+        revisions: {r, [adjustments: a, days_of_week: days, exceptions: e, trip_short_names: t]}
```
losvedir, line 41: Hm, these are the same associations as `DisruptionRevision.associations/0`. They could be used in a second load: `Repo.one!() |> Repo.preload(...)`, but I guess that doesn't work with the sorting here. But the sorting is of the associations right, since it's just a single revision being returned, so maybe that could be part of the preload call? I haven't tried that.

lib/arrow_web/templates/disruption/show.html.eex
```diff
@@ -1 +1,96 @@
...
+      <% end %>
+    </ul>
+
+    <hr class="my-3" />
+
+    <div class="row">
+      <%= if not @revision.is_active do %>
+        <div class="col-xs-12">
+          <div class="m-disruption-details__deletion-indicator">
+            <b>Note:</b> This disruption is deleted. It will not appear in GTFS.
```
losvedir, line 34: I tried this out locally: very nice treatment of this flow!

lib/arrow_web/views/disruption_view.ex
```diff
@@ -28,6 +29,12 @@ defmodule ArrowWeb.DisruptionView do
   defp published?(%{published_revision_id: id, revisions: [%{id: id}]}), do: true
   defp published?(_disruption), do: false
 
+  defp route_icon(route, size, opts \\ []) when size in ~w(sm lg) do
+    class = Keyword.get(opts, :class, "")
+    style = "background-image: url(/images/icon-#{route_icon_name(route)}.svg)"
```
losvedir, line 34: Can you use `static_path` here? That would then use the hash-digested path instead, which is useful if we change something about the icon files (to prevent a browser using a cached, incorrect icon).

losvedir (approved):
Looks great! Couple non-mandatory questions.

Also, can `src/disruptions/viewDisruption.tsx` and its test be deleted now?

lib/arrow_web/views/disruption_view.ex
digitalcora: Good catch, I just copied this from the React templates without considering it.

lib/arrow_web/controllers/disruption_controller.ex
digitalcora: Separately preloading would result in an additional query for each preloaded resource... of course it doesn't really matter since it's a (low) fixed number of queries and they're all indexed, so maybe there would be a benefit to having a `DisruptionRevision.preloads/0` shared by the index/show logic that contains all these and their desired ordering.

lib/arrow_web/controllers/disruption_controller.ex
digitalcora: I'll try it out.

PR #763, opened 2021-09-16
Clean up unused code post-SPA-migration

digitalcora:
**Asana:** [🏹 Migrate Arrow from SPA to Phoenix-routed](https://app.asana.com/0/584764604969369/1200671066645709) (subtask)

With bonus refactor: calendar events are generated on the server, removing the last use of JSON:API from the frontend and letting us delete all that code ➖➖➖

See commit messages for additional details!

#### Reviewer Checklist


assets/.prettierignore
```diff
@@ -1,3 +1,4 @@
+bootstrap.css
```
losvedir, line 1: Great idea!

assets/src/app.tsx
```diff
@@ -2,14 +2,15 @@
...
 // tslint:disable-next-line
 require("../css/app.css")
 
 import "phoenix_html"
-import ReactPhoenix from "./ReactPhoenix"
 
-import { DisruptionCalendar } from "./disruptions/disruptionCalendar"
-import DisruptionForm from "./disruptionForm"
+import ReactPhoenix from "./ReactPhoenix"
+import DisruptionCalendar from "./components/disruptionCalendar"
```
losvedir, line 12: I love the reorganization. One thing that's sort of bothered me about our current approach is that I'm used to component filenames being capitalized like their components, e.g., `import DisruptionCalendar from "./components/DisruptionCalendar"`. While you're at it, what do you think about changing the file names like that?

lib/arrow_web/views/disruption_view/calendar.ex
```diff
@@ -0,0 +1,68 @@
...
+  alias Arrow.Disruption.DayOfWeek
+  alias ArrowWeb.Endpoint
+  alias ArrowWeb.Router.Helpers, as: Routes
+
+  @doc """
+  Generates props for `DisruptionCalendar`, which has the same interface as `FullCalendar`.
+  Reference: https://fullcalendar.io/docs/event-parsing
+  """
+  @spec props([Disruption.t()]) :: %{atom => any}
+  def props(disruptions), do: %{events: events(disruptions)}
```
losvedir, line 14: I can't get over how much more straightforward this feels to me than all that JS code. Whew.

I really like the convention you've established here and with the previous form, of having a view module that provides the props. What do you think about making it a behaviour? On the one hand, we expect every module to have a `props` function, but on the other hand, the actual type of that function is going to be wildly different. So I don't know that a behaviour is such a good idea. I do like the idea of formalizing this convention _somehow_, I just don't know what sort of mechanism we have for that. 🤔

losvedir (approved):
This looks great to me!

I think there's also a now-unused `ready` column in the DB, right? And some Elixir code to cleanup around that? I'm happy to approve this PR as-is, though, if that would be a much larger undertaking, or if we're still using it in a way I'm not thinking of.

assets/src/app.tsx
digitalcora: Sounds fine, I'm in. 👍

lib/arrow_web/views/disruption_view/calendar.ex
digitalcora: It's an interesting idea, but no clever solution comes to the top of my mind, and I think as-is it should be clear to anyone reading the code what the convention is. Maybe if we add more components in the future, we'll see something that could be formalized, but as it is we only have two examples.

digitalcora:
> I think there's also a now-unused `ready` column in the DB, right?

We already took care of that in #737 — there are no remaining references to "ready" in the code, that I can find.

assets/src/app.tsx
digitalcora: As an aside, boy is this ever a pain on a case-insensitive file system. (the trick, it turns out, is to always use `git mv` and not anything else to do the renames)

losvedir:
> We already took care of that in #737

Oh yeah! Oops.

PR #769, opened 2021-09-21
feat: Add ROW status concept

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 Add ROW status concept](https://app.asana.com/0/584764604969369/1200687441670232)

Adds the concept of "ROW status" to the app, where a Disruption (revision) can be either approved or pending. A subsequent PR will update GTFS Creator to ignore pending disruptions.

#### Reviewer Checklist

### DIFF
--- a/assets/src/components/DisruptionForm.tsx
+++ b/assets/src/components/DisruptionForm.tsx
@@ -89,8 +89,7 @@ const DisruptionForm = ({
     tripShortNames: initialTripShortNames,
   },
 }: DisruptionFormProps) => {
-  const [isRowConfirmed, setIsRowConfirmed] =
-    useState<boolean>(initialRowConfirmed)
+  const [isRowConfirmed, setIsRowConfirmed] = useState(initialRowConfirmed)
 
   const [transitMode, setTransitMode] = useState<TransitMode>(
     initialAdjustments.length === 0
@@ -150,7 +149,7 @@ const DisruptionForm = ({
             <input
               className="form-check-input"
               type="radio"
-              name="row_confirmed"
+              name="revision[row_confirmed]"
               value={`${rowValue}`}
               checked={rowValue === isRowConfirmed}
               onChange={() => {
--- a/assets/src/components/DisruptionForm.tsx
+++ b/assets/src/components/DisruptionForm.tsx
@@ -14,6 +14,7 @@ type DaysOfWeek = { [dayName: string]: TimeRange }
 type DisruptionRevision = {
   startDate: string | null
   endDate: string | null
+  rowConfirmed: boolean
   adjustments: Adjustment[]
   daysOfWeek: DaysOfWeek
   exceptions: string[]
@@ -28,6 +29,11 @@ const days = [
   "friday",
   "saturday",
   "sunday",
+]
+
+const rowStatusLabels: [boolean, string][] = [
+  [true, "Approved"],
+  [false, "Pending"],
 ]
 
 const transitModeLabels: [TransitMode, string][] = [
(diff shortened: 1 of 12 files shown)


lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -32,6 +32,8 @@ defmodule ArrowWeb.DisruptionController do
   end
 
   def create(conn, %{"revision" => attrs}) do
+    attrs = update_in(attrs["row_confirmed"], &boolify_row_value/1)
```
losvedir, line 35: I'm not thrilled with this approach. I sort of expected Ecto's `cast` to handle convert `"true"` and `"false"` into `true` and `false` the way it will convert ISO 8601 strings into DateTimes, but it doesn't.

digitalcora: Are you sure? I just tested this with a boolean field in `alerts_concierge` and it casts the string versions of `true` and `false` to their boolean equivalents. The built-in `Phoenix.HTML.Form.checkbox` helper also suggests it should work this way, as it adds a checkbox input with the value `"true"` and a hidden input for when it's unchecked with the value `"false"`.

assets/src/components/DisruptionForm.tsx
```diff
@@ -133,6 +143,24 @@ const DisruptionForm = ({
...
   return (
     <>
+      <fieldset>
+        <legend>approval status</legend>
+        {rowStatusLabels.map(([rowValue, rowLabel]) => (
+          <label key={rowLabel} className="form-check form-check-label">
+            <input
+              className="form-check-input"
+              type="radio"
+              name="row_confirmed"
```
digitalcora, line 153: I think this might be your issue with the casting — it should be `revision[row_confirmed]`.

priv/repo/migrations/20210921192435_add_row_status.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Arrow.Repo.Migrations.AddRowStatus do
+  use Ecto.Migration
+
+  def change do
+    alter table(:disruption_revisions) do
+      add :row_confirmed, :boolean, null: false, default: true
```
digitalcora, line 6: This says `default: true`, but the field definition says `default: false`.

assets/src/components/DisruptionForm.tsx
```diff
@@ -76,12 +82,16 @@ const DisruptionForm = ({
...
     endDate: initialEndDate,
+    rowConfirmed: initialRowConfirmed,
     adjustments: initialAdjustments,
     daysOfWeek: initialDaysOfWeek,
     exceptions: initialExceptions,
     tripShortNames: initialTripShortNames,
   },
 }: DisruptionFormProps) => {
+  const [isRowConfirmed, setIsRowConfirmed] =
+    useState<boolean>(initialRowConfirmed)
```
digitalcora, line 93: I don't think `<boolean>` here is needed, as the type defaults to the type of the initial value, which in this case is correct (`initialRowConfirmed` is a boolean, and the type of the state should also be a boolean).

digitalcora:
...oops, I fell victim to "one more comment" and basically drive-by reviewed your PR 😅 Forgive me, @paulswartz

assets/src/components/DisruptionForm.tsx
losvedir: Aha! That was it. Good catch, thanks.

paulswartz (approved):
@digitalcora no worries; that means less for me to review!

🍰

PR #770, opened 2021-09-22
feat: disruptions have a description

bklebe:
#### Summary of changes
**Asana Ticket:** [Disruptions have a description](https://app.asana.com/0/584764604969369/1200681721075713/f)

Added a new Ecto Schema field `DisruptionRevision.description`. For existing `disruption_revisions`, this is populated with their adjustments concatenated with `, ` in the migration. This field validates as required by the schema changeset functions. Disruption descriptions show in the disruption table at the index, in the feed, and when viewing, editing, or creating a disruption. Where route icons were shown next to each of a disruption's adjustments, now only unique route icons are shown for the adjustments.

#### Reviewer Checklist


lib/arrow/disruption_revision.ex
```diff
@@ -16,6 +16,7 @@ defmodule Arrow.DisruptionRevision do
           end_date: Date.t() | nil,
           start_date: Date.t() | nil,
           is_active: boolean(),
+          description: String.t(),
```
digitalcora, line 19: Should be `String.t() | nil`?

assets/src/components/DisruptionForm.tsx
```diff
@@ -180,6 +185,25 @@ const DisruptionForm = ({
         ))}
       </fieldset>
 
+      <fieldset>
+        <legend>description</legend>
+        <label htmlFor="description" hidden>
+          description
+        </label>
+        <textarea
```
digitalcora, line 193: We can add a `required` property to this, and the browser will give it a nice client-side error message if it's blank.

mix.exs
```diff
@@ -73,7 +73,7 @@ defmodule Arrow.MixProject do
       {:tzdata, "~> 1.1"},
       {:ueberauth_cognito, "~> 0.1"},
       {:ueberauth, "~> 0.6"},
-      {:wallaby, "~> 0.28.1", runtime: false, only: :test}
+      {:wallaby, "~> 0.28.1", runtime: false, only: [:dev, :test]}
```
digitalcora, line 76: ❓

lib/arrow_web/templates/disruption/show.html.eex
```diff
@@ -16,14 +16,16 @@
       <% end %>
     </div>
 
-    <ul class="m-disruption-details__adjustment-list">
-      <%= for adjustment <- @revision.adjustments do %>
-        <li class="m-disruption-details__adjustment-item">
```
digitalcora, line 21: These two class names (`-list` and `-item`) don't seem to be used anywhere else, so we can remove them from the CSS.

lib/arrow_web/templates/disruption/_table.html.eex
```diff
@@ -14,16 +14,16 @@
...
+                <%= for adjustment <- Enum.uniq_by(revision.adjustments, &(&1.route_id)) do %>
+                    <%= route_icon(@conn, adjustment.route_id, "sm", class: "my-1 mr-2") %>
+                <% end %>
               </div>
             <% end %>
-          <% else %>
-            <i>(no adjustments selected)</i>
-          <% end %>
+            <%= revision.description %>
+          </div>
```
digitalcora, line 26: Seems maybe worth extracting this whole block (the icons-plus-description) into a partial, for de-duplication?

priv/repo/migrations/20210917173348_add_disruption_revision_description.exs
```diff
@@ -0,0 +1,27 @@
...
+    repo().query!("""
+    UPDATE disruption_revisions dr
+    SET description = (SELECT string_agg(a.source_label, ', ')
+                       FROM disruption_revisions d
+                       JOIN disruption_adjustments da
+                       ON d.id = da.disruption_revision_id
+                       JOIN adjustments a
+                       ON da.adjustment_id = a.id
+                       WHERE dr.id = d.id
+                       GROUP BY d.id)
```
digitalcora, line 22: Since we don't exactly have an automated test for it, did you test that this works as expected on a snapshot of the prod database?

assets/css/disruption_form.css
```diff
@@ -27,3 +27,7 @@
   font-size: large;
   font-weight: 200;
 }
+
+.description {
```
digitalcora, line 31: I don't see any element with class `description`. And we should probably make this follow the BEM-ish pattern of the other class names, if we do use it for something.

digitalcora (commented):
General comments:

1. I don't see any changes to the calendar here, though the task specifies that descriptions should display on the calendar. On the other hand, it doesn't really specify _how_ that should happen, and there's no attached mockup. Can we get clarification on this?

2. There seems to be a huge change to the `package-lock.json`, despite no frontend packages changing — let's revert that?

lib/arrow/disruption_revision.ex
digitalcora: This is the change I thought of because that's how we currently handle `start_date` and `end_date`: they are nullable in the database and the schema, but the validations make them required. We could change those ones too, I suppose.

assets/src/components/DisruptionForm.tsx
```diff
@@ -180,6 +185,25 @@ const DisruptionForm = ({
         ))}
       </fieldset>
 
+      <fieldset>
+        <legend>description</legend>
+        <label htmlFor="description" hidden>
```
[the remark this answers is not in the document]
bklebe: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#attr-hidden

It's difficult to write Wallaby tests without a label or an id. What would you recommend?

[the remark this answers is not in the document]
digitalcora: You learn something new every day! For some reason I couldn't find it in my devdocs search.

> It's difficult to write Wallaby tests without a label or an id. What would you recommend?

We sort of discussed this in Slack but in this instance I think rather than making the markup fit the limitations of the testing tool, we should implement test code that accommodates the desired markup. For now I think a `css` finder with `[aria-label="description"]` would be fine, since it seems like augmenting how Wallaby searches for form elements would be a substantial task.

priv/repo/migrations/20210917173348_add_disruption_revision_description.exs
bklebe: No, I didn't. How do I go about doing that? I tried it with some disruptions I created locally.

digitalcora: You can get a copy of the data in production using the `mix copy_db` task, with the environment variable `ARROW_DOMAIN` set to `https://arrow.mbta.com` and `ARROW_API_KEY` set to the value you get when you go to https://arrow.mbta.com/mytoken. Note this wipes out whatever data is already in your local database.

mix.exs
bklebe: I added this to enable experimenting with Wallaby in `iex -S mix`. Would it be better to leave it out and do `MIX_ENV=test iex -S mix`?

digitalcora: I think that's probably what I'd do — in that case you also know you're running in the exact same environment the tests are, so your experimenting will behave the same way.

README.md
```diff
@@ -35,6 +35,8 @@
   - `mix dialyzer` — check typespecs
   - `mix format` — format code
   - `mix credo` — lint code
+  - `mix ecto.setup` — setup database
```
digitalcora, line 38: This is already on L26 as part of the setup instructions.

README.md
```diff
@@ -35,6 +35,8 @@
...
+  - `mix copy_db` — copy database from `$ARROW_DOMAIN` (be sure to setup `.envrc` with the desired domain and you `$ARROW_API_KEY` first)
```
digitalcora, line 39: typo: `you` => `your`

Also, maybe we should explain how to get a key in the first place.

lib/arrow/disruption_revision.ex
```diff
@@ -126,7 +128,7 @@ defmodule Arrow.DisruptionRevision do
   defp clone_fields(%module{} = record) do
     Map.take(
       record,
-      module.__schema__(:fields) -- ~w(id disruption_revision_id inserted_at updated_at)a
+      module.__schema__(:fields) -- ~w(id disruption_revision_id inserted_at updated_at description)a
```
digitalcora, line 131: These are the fields that _shouldn't_ be copied into the new structs when we clone a revision, so I don't think description should be among them. The idea here is that we no longer have to duplicate all the field names in the clone logic (see #773).

lib/arrow/disruption_revision.ex
bklebe: Which should I do? Make it accept `nil` or write a new migration to make the DB reject null? My inclination is towards the latter.

digitalcora: Either way I'm fine as long as the types, schema, and database are all in agreement.

losvedir: I like the latter, too, (having suggested it!), but I'm fine either way. Chef's choice.

bklebe: I added the constraint.

assets/src/components/DisruptionForm.tsx
```diff
@@ -12,6 +12,7 @@ type TimeRange = { start: string | null; end: string | null }
 type Adjustment = { id: number; label: string; routeId: string }
 type DaysOfWeek = { [dayName: string]: TimeRange }
 type DisruptionRevision = {
+  description: string | null
```
digitalcora, line 15: As per the constraint, this should just be `string` (I think allows simplifying L102).

priv/repo/migrations/20210924191802_disruption_description_not_null.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Arrow.Repo.Migrations.DisruptionDescriptionNotNull do
+  use Ecto.Migration
+
+  def up do
+    alter table(:disruption_revisions) do
+      modify :description, :text, null: false
```
digitalcora, line 6: We should be able to include this in the first migration rather than having a separate migration for it (alter / execute / alter again), and then it all happens in a single transaction.

digitalcora (approved):
Looks fine to me, pending looking at whether we can merge those migrations.

PR #772, opened 2021-09-22
fix: ROW status tweaks

losvedir:
#### Summary of changes
**Asana Ticket:** n/a

Tweaks and fix-ups to #769.

The largest change is renaming the column from `row_confirmed` to `row_approved`. I don't know where I got "confirmed" in my head before, but the mocks are pretty consistent in calling it "approved". The presentation layer doesn't *have* to match exactly the data layer, but might as well get started off on the right foot.

I added `row_approved` to a few more places: The API that gtfs_creator uses, `clone!`, `db_structure`, a type spec.

I added a simple `approval status` to the show page. I found mocks to a version of the show page we'll have in the future with a sidebar with notes and some other things. So I didn't copy the design exactly, which won't work now, but I did use the wording. I made a couple tickets regarding that mock to make sure getting it to look like that won't fall through the cracks.

I updated the "status" column on the homepage table to be "GTFS status" to make it a little less confusing until we get to the ROW status on table ticket.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/templates/disruption/show.html.eex
+++ b/lib/arrow_web/templates/disruption/show.html.eex
@@ -40,6 +40,16 @@
 
       <div class="col-md-10 <%= inactive_class %>">
         <div class="mb-3">
+          <h4>approval status</h4>
+
+          <div class="pl-3">
+            <span class="m-disruption-details__row_status">
+              <%= if @revision.row_approved, do: "Approved", else: "Pending" %>
+            </span>
+          </div>
+        </div>
+
+        <div class="mb-3">
           <h4>date range</h4>
 
           <div class="pl-3">
--- a/lib/arrow_web/templates/disruption/show.html.eex
+++ b/lib/arrow_web/templates/disruption/show.html.eex
@@ -43,9 +43,7 @@
           <h4>approval status</h4>
 
           <div class="pl-3">
-            <span class="m-disruption-details__row_status">
-              <%= if @revision.row_approved, do: "Approved", else: "Pending" %>
-            </span>
+            <%= if @revision.row_approved, do: "Approved", else: "Pending" %>
           </div>
         </div>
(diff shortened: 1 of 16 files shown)


lib/arrow_web/templates/disruption/show.html.eex
```diff
@@ -39,6 +39,16 @@
       <% inactive_class = if(@revision.is_active, do: "", else: "text-muted") %>
 
       <div class="col-md-10 <%= inactive_class %>">
+        <div class="mb-3">
+          <h4>approval status</h4>
+
+          <div class="pl-3">
+            <span class="m-disruption-details__row_status">
```
digitalcora, line 46: This span/class doesn't seem to do anything — unless I'm missing that the class already exists in the CSS before this PR?

priv/repo/migrations/20210922191945_row_status_name_change.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Arrow.Repo.Migrations.RowStatusNameChange do
+  use Ecto.Migration
+
+  def change do
+    alter table(:disruption_revisions) do
+      remove :row_confirmed
+      add :row_approved, :boolean, null: false, default: true
```
digitalcora, line 7: There is a `rename/3` which can rename a column without dropping the existing data or having to redefine it. (this remove-and-add way also isn't reversible, though it would be if you used `remove/3`)

lib/arrow_web/templates/disruption/show.html.eex
losvedir: Nope, I'll remove. I had started making it look somewhat like the mocks before deciding to hold off for now.

digitalcora (approved):
🛤️ 🔧 ✅

PR #773, opened 2021-09-23
refactor: use introspection to clone revisions

digitalcora:
Docs: https://hexdocs.pm/ecto/3.7.1/Ecto.Schema.html#module-reflection

This avoids duplicating all the field names of revisions and their associations, so we don't have to remember to also update the clone function whenever a field is added or removed.

By creating a changeset for the clone instead of immediately saving it to the database, we also avoid the need to wrap updates and deletes in a transaction.

We still duplicate the names of associations — theoretically we could also avoid this with introspection, but it's less clear that doing so would be worth the complexity.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/disruption_revision.ex
+++ b/lib/arrow/disruption_revision.ex
@@ -48,9 +48,13 @@ defmodule Arrow.DisruptionRevision do
   @spec associations() :: [atom()]
   def associations, do: ~w(adjustments days_of_week exceptions trip_short_names)a
 
-  @spec changeset(t(), map) :: Changeset.t(t())
-  def changeset(revision, attrs) do
-    revision
+  @doc "Gets a revision by ID with preloaded associations."
+  @spec get!(id()) :: t()
+  def get!(id), do: __MODULE__ |> Repo.get!(id) |> Repo.preload(associations())
+
+  @spec changeset(t() | Changeset.t(t()), map) :: Changeset.t(t())
+  def changeset(data, attrs) do
+    data
     |> Changeset.cast(attrs, [:start_date, :end_date, :row_approved])
     |> Changeset.put_assoc(:adjustments, Adjustment.from_revision_attrs(attrs))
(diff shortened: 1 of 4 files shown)


lib/arrow/disruption_revision.ex
```diff
@@ -144,6 +123,13 @@ defmodule Arrow.DisruptionRevision do
     :ok
   end
 
+  defp clone_fields(%{__struct__: module} = record) do
```
losvedir, line 126: Rather than relying on the special `:__struct__` key we know lies underneath, maybe `defp clone_fields(%module{} = record)`?

lib/arrow/disruption_revision.ex
```diff
@@ -144,6 +123,13 @@ defmodule Arrow.DisruptionRevision do
...
+    Map.take(
+      record,
+      module.__schema__(:fields) -- ~w(id disruption_revision_id inserted_at updated_at)a
```
losvedir, line 129: This feels a little intrusive to me. What about deferring "how to clone" to the object itself? `module.clone(record)`?

losvedir (approved):
Much improved! Thanks. I left a couple suggestions, but if you disagree I'm happy with it being merged like this. 👍

lib/arrow/disruption_revision.ex
digitalcora: For some reason I keep forgetting you can do that. (in fairness, so do a lot of people whose code samples I find on the internet, apparently)

lib/arrow/disruption_revision.ex
digitalcora: What's an "object"? 😛

Japes aside, I think it's a good idea, but I also like how it's currently organized a bit better since all the "clone" logic is localized to the one module where the operation is defined/supported — it wouldn't make sense to, say, clone an `Exception` on its own.

PR #780, opened 2021-10-01
feat: filtering by ROW status

bklebe:
#### Summary of changes
**Asana Ticket:** [ROW status can be filtered](https://app.asana.com/0/584764604969369/1200689411100427/f)

Added a new filter for both the table and calendar views that when activated, only shows ROW approved disruptions.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/disruption_controller/filters.ex
+++ b/lib/arrow_web/controllers/disruption_controller/filters.ex
@@ -22,11 +22,12 @@ defmodule ArrowWeb.DisruptionController.Filters do
 
   @type t :: %__MODULE__{
           routes: MapSet.t(String.t()),
+          only_approved?: boolean(),
           search: String.t() | nil,
           view: Calendar.t() | Table.t()
         }
 
-  defstruct routes: @empty_set, search: nil, view: %Table{}
+  defstruct routes: @empty_set, only_approved?: false, search: nil, view: %Table{}
 
   @spec calendar?(%__MODULE__{}) :: boolean
   def calendar?(%__MODULE__{view: %Calendar{}}), do: true
@@ -43,6 +44,7 @@ defmodule ArrowWeb.DisruptionController.Filters do
 
     %__MODULE__{
       routes: params |> Map.get("routes", []) |> MapSet.new(),
+      only_approved?: not is_nil(params["only_approved"]),
       search: if(params["search"] in [nil, ""], do: nil, else: params["search"]),
       view: view_mod.from_params(params)
     }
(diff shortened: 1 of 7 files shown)


lib/arrow_web/controllers/disruption_controller/filters.ex
```diff
@@ -22,11 +22,12 @@ defmodule ArrowWeb.DisruptionController.Filters do
 
   @type t :: %__MODULE__{
           routes: MapSet.t(String.t()),
+          only_approved: boolean(),
```
paulswartz, line 25: suggestion: Elixir-y booleans end with `?`
```suggestion
          only_approved?: boolean(),
```
(not going to re-tag it everywhere)

lib/arrow_web/controllers/disruption_controller/index.ex
```diff
@@ -15,6 +15,10 @@ defmodule ArrowWeb.DisruptionController.Index do
     from [revisions: r] in query, where: is_nil(r.end_date) or r.end_date > ^cutoff
   end
 
+  defp apply_filter({:only_approved, true}, query) do
+    from [revisions: r] in query, where: r.row_approved
+  end
+
```
paulswartz, line 21: suggestion: This is further down ([line 50](https://github.com/mbta/arrow/pull/780/files#diff-a0076cc310a11366f090a4157b93d61fe97684f4bc56f1bcb8879b589167f87fR50)), but it might be worth not having a fallback case. That way, you get an explicit crash if this file gets an unexpected filter, rather than showing more items than expected.

If that change is made, you'd also want a `{:only_approved?, false}` case here which returns the query as-is.

lib/arrow_web/controllers/disruption_controller/filters.ex
bklebe: Should I also update `Table.include_past`?

paulswartz: In general yes, but I'll leave it to you as to whether that's out-of-scope here.

lib/arrow_web/controllers/disruption_controller/index.ex
bklebe: >If that change is made, you'd also want a `{:only_approved?, false}` case here which returns the query as-is.

Hmmm, currently we use the lack of a query parameter to indicate the filter is inactive, not the presence of a `false` flag. I believe this is because it keeps the index route cleaner. Is this still relevant in that case?

paulswartz: I believe this is after the query parameter parsing, so there's always a `only_approved?` value from the struct.

bklebe: Ah yes, right. Thanks. In that case we would probably need fallbacks for at least `include_past?` too.

paulswartz (approved):
🍰 pending CI

lib/arrow_web/controllers/disruption_controller/index.ex
```diff
@@ -27,10 +35,14 @@ defmodule ArrowWeb.DisruptionController.Index do
     from [adjustments: a] in query, where: ^condition
   end
 
+  defp apply_filter({:routes, routes}, query) when routes == @empty_set, do: query
+
   defp apply_filter({:search, search}, query) when is_binary(search) do
     from [adjustments: a] in query, where: ilike(a.source_label, ^"%#{search}%")
   end
 
+  defp apply_filter({:search, search}, query) when is_nil(search), do: query
```
paulswartz, line 44:
```suggestion
  defp apply_filter({:search, nil}, query), do: query
```

paulswartz (approved):
one small nitpick, otherwise 🍰

PR #782, opened 2021-10-04
fix: don't generate coverage data for deps/

bklebe:
#### Summary of changes
**Asana Ticket:** [fix Arrow coverage reporting](https://app.asana.com/0/584764604969369/1201120775201216/f)

Tells Lcov to ignore the `deps/` path when generating coverage info. We don't want `deps/mox/lib/mox.ex` covered and it breaks coverage reporting.

#### Reviewer Checklist

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -22,7 +22,7 @@ defmodule Arrow.MixProject do
         ignore_warnings: ".dialyzer.ignore-warnings"
       ],
       preferred_cli_env: ["test.integration": :test],
-      test_coverage: [tool: LcovEx]
+      test_coverage: [tool: LcovEx, ignore_paths: ["deps/"]]
     ]
   end
 


mix.exs
```diff
@@ -22,7 +22,7 @@ defmodule Arrow.MixProject do
         ignore_warnings: ".dialyzer.ignore-warnings"
       ],
       preferred_cli_env: ["test.integration": :test],
-      test_coverage: [tool: LcovEx]
+      test_coverage: [tool: LcovEx, ignore_paths: ["deps/"]]
```
paulswartz, line 25: Is this something that we should include in the GitHub Actions migration doc?

paulswartz (approved):
🍰

mix.exs
bklebe: Probably? I think `lcov.info` will contain coverage info for deps that are used in tests if not. `realtime_signs` uses Mox and doesn't seem to have this problem, but `api` does. Not sure what the difference is.

bklebe: Actually `api` seems to have a different problem. Disregard.

PR #790, opened 2021-10-15
feat: read-only Arrow

bklebe:
#### Summary of changes
**Asana Ticket:** [make read-only version of Arrow](https://app.asana.com/0/584764604969369/1200520004636479/f)

Overhauled how the app perceives users from an authorization perspective. All authorization checks are based on the central module `Arrow.Permissions`, and are carried out against a new `User` struct that is loaded into the connection after authentication. Checks are redundant in some places to make it abundantly clear which functionalities are meant to be restricted. I couldn't decide how much to change the parts of the test suite that still treat authentication as authorization, though I introduced the `:authenticated_admin` tag which should be a drop-in replacement for `:authenticated`. Test coverage may not be up to snuff, let me know if you see anything that could be tested more effectively.

#### Reviewer Checklist


lib/arrow/accounts/user.ex
```diff
@@ -0,0 +1,15 @@
...
+
+  @type group() :: :admin | :viewer
+
+  @type t() :: %__MODULE__{
+          id: String.t(),
+          groups: list(group())
+        }
+
+  @enforce_keys [:id]
+  defstruct @enforce_keys ++ [groups: []]
```
digitalcora, line 14: If this were instead a `MapSet` of groups, it would prevent the same group from (nonsensically) appearing multiple times.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -21,55 +26,65 @@ defmodule ArrowWeb.DisruptionController do
 
   @spec new(Conn.t(), Conn.params()) :: Conn.t()
   def new(conn, _params) do
-    changeset = DisruptionRevision.new() |> Changeset.change()
-    render(conn, "new.html", adjustments: Adjustment.all(), changeset: changeset)
+    with :ok <- Permissions.authorize(:create_disruption, conn.assigns.current_user) do
```
digitalcora, line 29: These checks seem redundant with the `CheckPermission` plug already applied above. I think both these and the `action_fallback` could be removed, and everything would still work the same.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -2,11 +2,16 @@ defmodule ArrowWeb.DisruptionController do
...
-  alias Arrow.{Adjustment, Disruption, DisruptionRevision}
+  alias Arrow.{Adjustment, Disruption, DisruptionRevision, Permissions}
   alias ArrowWeb.ErrorHelpers
   alias Ecto.Changeset
   alias Plug.Conn
 
+  action_fallback(ArrowWeb.FallbackController)
+  plug(ArrowWeb.Plug.CheckPermission, :create_disruption when action in [:new, :create])
+  plug(ArrowWeb.Plug.CheckPermission, :update_disruption when action in [:edit, :update])
+  plug(ArrowWeb.Plug.CheckPermission, :delete_disruption when action in [:delete])
```
digitalcora, line 13: We could alias this to reduce the repetition. (surprised Credo isn't bugging us about it already)

lib/arrow_web/plug/load_user.ex
```diff
@@ -0,0 +1,29 @@
+defmodule ArrowWeb.Plug.LoadUser do
```
digitalcora, line 1: I think this is a bit of a misnomer since no resource is being "loaded" from anywhere; we're only assigning a struct into the conn. `AssignUser`?

lib/arrow_web/plug/load_user.ex
```diff
@@ -0,0 +1,29 @@
...
+  @cognito_groups Application.compile_env!(:arrow, :cognito_groups)
+  @default_groups Application.compile_env!(:arrow, :default_groups)
+
+  @spec init(Plug.opts()) :: Plug.opts()
+  def init(options), do: options
+
+  @spec call(Plug.Conn.t(), Plug.opts()) :: Plug.Conn.t()
+  def call(conn, _opts) do
+    claims = Guardian.Plug.current_claims(conn)
+    id = claims["sub"]
```
digitalcora, line 17: It seems like we expect `sub` to always be in the claims (else we could end up with `%User{id: nil}`), so this could be `%{"sub" => user_id} = claims = Guardian.Plug.current_claims(conn)`

lib/arrow_web/plug/load_user.ex
```diff
@@ -0,0 +1,29 @@
...
+
+    groups =
+      claims
+      |> Map.get("groups", [])
+      |> Enum.flat_map(&[Map.get(@cognito_groups, &1, [])])
```
digitalcora, line 22: Since the values of `@cognito_groups` are single items rather than lists, this might be a clearer way to express it:

```elixir
claims |> Map.get("groups", []) |> Enum.map(&@cognito_groups[&1]) |> Enum.reject(&is_nil/1)
```

lib/arrow_web/ensure_admin_group.ex
```diff
@@ -1,4 +1,4 @@
-defmodule ArrowWeb.EnsureArrowGroup do
+defmodule ArrowWeb.EnsureAdminGroup do
```
digitalcora, line 1: As we discussed earlier, I think since we're introducing this new permission-based framework it would make a lot of sense to consolidate this into it. Currently the things this controls access to are:

* viewing the changes feed (the "Josh page")
* generating and viewing one's API token
* the hook that notifies the app when changes have been published in GTFS
* fetching a database dump

Accordingly maybe we could introduce some `Permissions`:

```elixir
@required_groups %{
  create_disruption: [:admin],
  update_disruption: [:admin],
  delete_disruption: [:admin],
  view_change_feed: [:admin],
  use_api_token: [:admin],
  notify_published: [:admin],
  dump_database: [:admin]
}
```

I suggest `use_api_token` here because e.g. if someone were to be removed from the group that allows them to view/generate their API token, they probably shouldn't be allowed to still use their already-generated API token to do other things (the only thing that protects against this currently is all of these actions being bundled under the "ensure admin" check). So this would imply that `TryApiTokenAuth` would itself be protected by the `use_api_token` permission, in addition to the token-auth'd endpoints being protected by their own permissions.

lib/arrow_web/router.ex
```diff
@@ -48,16 +52,27 @@ defmodule ArrowWeb.Router do
   end
 
   scope "/", ArrowWeb do
-    pipe_through [:redirect_prod_http, :browser, :auth, :ensure_auth, :ensure_arrow_group]
+    pipe_through([:redirect_prod_http, :browser, :auth, :ensure_auth, :ensure_arrow_group])
```
digitalcora, line 55: This doesn't have `:load_user`... actually, shouldn't almost every endpoint have `:load_user`? Maybe we can add it to a larger shared authentication/authorization pipeline, since a lot of these things we always want to use together. Something like:

```elixir
pipeline :authenticate do
  plug(ArrowWeb.AuthManager.Pipeline)
  plug(ArrowWeb.TryApiTokenAuth)
  plug(Guardian.Plug.EnsureAuthenticated)
  plug(ArrowWeb.Plug.LoadUser)
end
```

I don't think there's a downside to applying `TryApiTokenAuth` to all authenticated requests, but if we wanted to avoid that, there could be a separate `:token_authenticate` pipeline.

lib/arrow_web/templates/disruption/index.html.eex
```diff
@@ -1,6 +1,8 @@
 <div class="row my-3">
   <div class="col">
-    <a class="btn btn-primary" href="/disruptions/new">+ create new</a>
+    <%= if Permissions.authorize?(:create_disruption, @conn.assigns.current_user) do %>
```
digitalcora, line 3: What do you think of assigning a `@current_user` in views that need it? I'm thinking this would make the view's "dependency" on the current user clearer (and wouldn't be too much extra controller code, since we could use pattern-matching to extract the user in the action's function head).

lib/arrow_web/templates/disruption/show.html.eex
```diff
@@ -86,10 +86,12 @@
 
       <%= if @revision.is_active do %>
         <div class="col-md-2">
-          <%= link("edit",
-                class: "btn btn-primary w-100",
-                to: Routes.disruption_path(@conn, :edit, @id)
-              ) %>
+          <%= if Permissions.authorize?(:update_disruption, @conn.assigns.current_user) do %>
```
digitalcora, line 89: Could be combined with above? (unless the layout somehow breaks otherwise)

```eex
<%= if Permissions.authorize?(:update_disruption, @conn.assigns.current_user) and @revision.is_active do %>
```

lib/arrow_web/templates/disruption/show.html.eex
```diff
@@ -6,7 +6,7 @@
         <h5>ID <span class="ml-2 font-weight-normal"><%= @id %></span></h5>
       </div>
 
-      <%= if @revision.is_active do %>
+      <%= if Permissions.authorize?(:delete_disruption, @conn.assigns.current_user) && @revision.is_active do %>
```
digitalcora, line 9: We should use `and` and `or` when we expect both sides to be booleans — then the code will fail loudly if they aren't. (there's no difference in precedence, it's just that `&&` and `||` allow the operands to be anything)

lib/arrow_web/templates/unauthorized/index.html.eex
```diff
@@ -1,3 +1,8 @@
-You are not authorized to access Arrow. If you believe you
-should have access, please have your supervisor contact
-realtimeappsdl@mbta.com.
+<section class="d-flex flex-column justify-content-center align-items-center h-100">
+  <h4 class="text-danger text-start">Whoops! You are not authorized to access this page.</h4>
```
digitalcora, line 2: Pour one out for "Woops!" 😢

digitalcora (commented):
Comments away! Note: I didn't review the tests yet, since you mentioned still working on those.

lib/arrow_web/controllers/disruption_controller.ex
bklebe: >  kind of like how the paths that map to each controller action all live together in a separate Router module instead of being defined on each action. But I certainly don't mind if it's the other way.

It's purely because this is security sensitive that I chose to do it like this. I actually did both initially because there's no real _harm_, it's just redundant. What I don't want is to look at a restricted operation and have to look elsewhere to determine if it's being authorized properly. Can plugs be put anywhere in a controller or do they need to be at the top? I would be OK with using plugs if we could put them right above every relevant action, with `only:`.

digitalcora: > Can plugs be put anywhere in a controller or do they need to be at the top?

I don't think there's anything stopping us, though I might prefer the `with` approach over that.

PR #799, opened 2021-11-11
Dependabot (+ Elixir, Erlang, and Phoenix 1.6)

losvedir:
#### Summary of changes
**Asana Ticket:** [🤖 tend to dependabot ](https://app.asana.com/0/584764604969369/1201340855635750)

The first commit is a general `mix deps.update --all`, except for `ueberauth` and `ueberauth_cognito` which breaks our local dev "fake ueberauth" strategy. I poked at this a bit, but didn't immediately see a solution (or find exactly the problem, though it's probably related to the new CSRF handling), so I gave up and fixed the versions to where they are, for now. I wanted to move on and update all the other things, but if we get more caught up on deps, this would be something to investigate.

The second commit is a general `npm update`.

The third commit updates Erlang and Elixir to the latest versions available in asdf.

The final commit updates to Phoenix 1.6 🎉 following Chris McCord's [update guide](https://gist.github.com/chrismccord/2ab350f154235ad4a4d0f4de6decba7b).

This branch is on https://arrow-dev.mbtace.com .

#### Reviewer Checklist


mix.exs
```diff
@@ -62,17 +62,22 @@ defmodule Arrow.MixProject do
...
       {:postgrex, ">= 0.0.0"},
       # If react_phoenix changes, check assets/src/ReactPhoenix.js, too
       {:react_phoenix, "1.2.0"},
+      {:telemetry_metrics, "~> 0.6"},
+      {:telemetry_poller, "~> 0.5"},
       {:tzdata, "~> 1.1"},
-      {:ueberauth_cognito, "~> 0.1"},
-      {:ueberauth, "~> 0.6"},
+      {:ueberauth_cognito, "0.2.0"},
+      {:ueberauth, "0.6.3"},
```
bklebe, line 80: Could we make a ticket to fix the problem with these and unpin them? Seems like authentication libraries are important to keep up to date, especially if we're not currently handling CSRF correctly.

losvedir: Yep, good idea. Ticket here: https://app.asana.com/0/881264583703207/1201382502692986

We are currently handling CSRF correctly. It used to be that `ueberauth` punted on it, so Eddie and I added it to `ueberauth_cognito`. Then someone convinced the maintainers of `ueberauth` that they should handle it themselves so all the integrations aren't separately responsible for it. `ueberauth_cognito` was supposed to be updated to work in the new regime, but isn't seeming to out of the box for us. But while we're on these two versions of `ueberauth` and `ueberauth_cognito` we should still be fine.

bklebe:
Ah, I see. So it's more that Ueberauth has yet to be updated?

________________________________________

bklebe:
er, ueberauth_cognito, rather

________________________________________

losvedir:
Well ueberauth_cognito was supposed to have been updated for it here: https://github.com/mbta/ueberauth_cognito/pull/52 (by a 3rd party contributor nonetheless!), and I confirmed it worked, I think, with signs-ui at the time. So I'm not exactly sure why it's not working now with Arrow.

mix.exs
bklebe: I see, so it's more like `ueberauth_cognito` needs to be updated to let ueberauth handle CSRF?

bklebe (approved):

PR #809, opened 2021-11-29
chore: Added color feature for pending disruption status

uchukwu:
#### Summary of changes
**Asana Ticket:** [🏹 ROW status on calendar](https://app.asana.com/0/584764604969369/1200687441670238/f)

[Includes changes in css file for added color events, associated with pending disruption status vs approved disruption status. Change is picked up in DCalendar template.]

#### Reviewer Checklist

### DIFF
--- a/assets/css/disruption_calendar.css
+++ b/assets/css/disruption_calendar.css
@@ -14,26 +14,50 @@
   color: #aaa;
 }
 
-.fc-event.route-Red {
+.fc-event.status-approved.route-Red {
   background-color: #da291c;
 }
 
-.fc-event.route-Blue {
+.fc-event.status-pending.route-Red {
+  background-color: #f4bfbb;
+}
+
+.fc-event.status-approved.route-Blue {
   background-color: #003da5;
 }
 
-.fc-event.route-Orange {
+.fc-event.status-pending.route-Blue {
+  background-color: #b3c5e4;
+}
+
+.fc-event.status-approved.route-Orange {
   background-color: #ed8b00;
 }
 
-.fc-event.route-Mattapan {
+.fc-event.status-pending.route-Orange {
+  background-color: #f9d6a6;
+}
+
+.fc-event.status-approved.route-Mattapan {
   background-color: #da291c;
 }
 
-.fc-event[class*="route-Green"] {
+.fc-event.status-pending.route-Mattapan {
+  background-color: #f4bfbb;
+}
+
+.fc-event.status-approved[class*="route-Green"] {
   background-color: #00843d;
 }
 
-.fc-event[class*="route-CR"] {
+.fc-event.status-pending[class*="route-Green"] {
+  background-color: #cce6d8;
+}
+
+.fc-event.status-approved[class*="route-CR"] {
   background-color: #80276c;
+}
+
+.fc-event.status-pending[class*="route-CR"] {
+  background-color: #d98fd3;
 }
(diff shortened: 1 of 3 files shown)


assets/css/disruption_calendar.css
```diff
@@ -1,23 +1,47 @@
...
+  background-color: #cce6d8;
+}
+
+.fc-event[class*="approved-route-CR"] {
   background-color: #80276c;
 }
+
+.fc-event[class*="pending-route-CR"] {
+  background-color: #d98fd3;
+}
```
paulswartz, line 47: suggestion: needs a trailing newline
```suggestion
}

```

lib/arrow_web/views/disruption_view/calendar.ex
```diff
@@ -42,7 +43,7 @@ defmodule ArrowWeb.DisruptionView.Calendar do
     |> Enum.map(fn {event_start, event_end} ->
       %{
         title: source_label,
-        classNames: "route-#{route_id}",
+        classNames: "#{row_status(row_approved)}-route-#{route_id}",
```
paulswartz, line 46: suggestion: what do you think about having separate classes for the approval status and the route? Something like:
```suggestion
        classNames: "{row_status_class(row_approved)} route-#{route_id}",
```

While it doesn't make much difference here (you still need all the same styles) it's more flexible in the future if we need to target them separately.

lib/arrow_web/views/disruption_view/calendar.ex
```diff
@@ -51,6 +52,14 @@ defmodule ArrowWeb.DisruptionView.Calendar do
     end)
   end
 
+  defp row_status(row_approved) do
+    if row_approved do
+        "approved"
+    else
+        "pending"
+    end
+  end
```
paulswartz, line 61: suggestion: might be better as a pattern match?
```suggestion
  defp row_status_class(true), do: "status-approved"
  defp row_status_class(false), do: "status-pending"
```

lib/arrow_web/views/disruption_view/calendar.ex
uchukwu: It appears that implementing your suggested change doesn't produce the expected color features in the calendar. Everything is set to the default blue. Am I missing something?

losvedir: What he means is instead of having a combined class that you then specifically style, like:

```
html:
<div class="approved-route-Blue"> ... </div>

css:
.approved-route-Blue { .... }
```

to instead add two separate classes, which can then be selected in the CSS by naming both:

```
html:
<div class="approved route-Blue"> ... </div>

css:
.approved.route-Blue { .... }
```

In the HTML, note that spaces delimit different classes, and an element can have multiple classes on it. In the CSS, note that you can style "elements with both this class1 and that class2" by using the `.class1.class2` selector (no space between the classes). (And in the case here, I think it would be `.fc-event.pending-status.route-Blue`.)

Right now, there's not much difference between the approaches, but by making `.pending-status` its own class, we could potentially do something like:

```css
.approved-status { font-weight: bold }
```

and then that would make all the approved events' bold, without having to individually style it for every route.

assets/css/disruption_calendar.css
```diff
@@ -1,23 +1,47 @@
...
+.fc-event.status-approved.route-Mattapan {
   background-color: #da291c;
 }
 
-.fc-event[class*="route-Green"] {
+.fc-event.status-pending.route-Mattapan {
+  background-color: #f4bfbb;
+}
+
+.fc-event[class*="status-approved route-Green"] {
```
paulswartz, line 33: suggestion: the `status-*` part of the class name doesn't need the wild-card matching

```suggestion
.fc-event.status-approved[class*="route-Green"] {
```

(also for the 3 items below)

paulswartz: see above comment: I think we both clicked the 🔃 button around the same time :)

paulswartz (approved):
🍰

PR #811, opened 2021-12-01
feat: show no-adjustments disruptions on calendar 

digitalcora:
**Asana:** [🏹 Disruptions can be created without adjustment](https://app.asana.com/0/584764604969369/1200681721075715)

We didn't say much in the task about _how_ these disruptions should be shown on the calendar, so I made up a distinct appearance for them and used the disruption ID as the title, to provide a way to distinguish between multiple disruptions with no adjustments (and since we already use disruption IDs prominently elsewhere in the UI).

Screen Shot 2021-12-01 at 4 08 02 PM

#### Reviewer Checklist

### DIFF
--- a/assets/css/disruption_calendar.css
+++ b/assets/css/disruption_calendar.css
@@ -1,3 +1,9 @@
+.fc-h-event {
+  /* replace colored border with padding */
+  border: none;
+  padding: 2px;
+}
+
 .fc-event.route-Red {
   background-color: #da291c;
 }
--- a/assets/css/disruption_calendar.css
+++ b/assets/css/disruption_calendar.css
@@ -4,6 +4,16 @@
   padding: 2px;
 }
 
+.fc-event.route-none {
+  background-color: #fff;
+  border: 2px dashed #ccc;
+  padding: 0;
+}
+
+.fc-event.route-none .fc-event-title {
+  color: #aaa;
+}
+
 .fc-event.route-Red {
   background-color: #da291c;
 }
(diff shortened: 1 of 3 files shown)


losvedir (approved):
Looks good to me. I think "disruption ID" is fine for now. We'll probably re-visit later, but better to tackle it at the same time as disruptions that have adjustments, too.

digitalcora:
I think actually we will revisit this very soon, once [🏹 Disruptions-sans-adjustment have a named route or mode](https://app.asana.com/0/584764604969369/1200681721075719), since it seems like that field will be required for disruptions without adjustments. This brings up the question of, when we do that task, how we should handle disruptions that have since been created without adjustments — we can't just arbitrarily assign them one of the available modes (or can we?).

losvedir:
> This brings up the question of, when we do that task, how we should handle disruptions that have since been created without adjustments — we can't just arbitrarily assign them one of the available modes (or can we?).

Yeah, we can't currently. That's a good chunk of what the ticket is about. i.e. the acceptance criteria:

> * At the data level, a disruption without any adjustments has an associated "line/mode", for the purposes of filtering and display. The options are Red, Orange, Blue, Green-B, Green-C, Green-D, Green-E, Mattapan, Bus, CR, Silver Line, Ferry, Green Line.
> * Adding adjustments clears the line.

with

> *IMPLEMENTATION NOTES*
> Perhaps a field on disruption_revision?

digitalcora:
Right, I mean, we _could_ find all disruptions whose current revision has no adjustments, and create a new revision for them with some non-null value for the mode — and then this UI would never again be relevant. But at the database level, if there have ever been any revisions with no adjustments (and no mode), then those will always be there.

PR #813, opened 2021-12-07
Send a fancy slack notification

bklebe:
#### Summary of changes
**Asana Ticket:** [Send a fancy Slack notification](https://app.asana.com/0/584764604969369/1200871105094777/f)

Two new modules, `Arrow.Slack.Notifier` and `Arrow.Slack.DisruptionNotification`:

- `Notifier` is the entry point for sending notifications

-  `DisruptionNotification` formats a Slack message about disruption changes.

I'm looking for feedback on the general direction right now, since we discussed reviewing the design. I haven't written any tests yet and I haven't implemented the secrets configuration either, but I expect the latter to be fairly simple.

#### Reviewer Checklist


digitalcora:
I'm 👍 on the general approach. One suggestion: rather than having the `Notifier` call `DisruptionNotification`, have the notifier accept an already-built notification, and have the controller build the notification. That way `Notifier` can be _just_ the API integration logic, and can be mocked in the test environment, rather than having to use Bypass and swap around env vars to mock the actual HTTP endpoint.

losvedir:
I like this. I second the suggestion to separate the API integration logic from building the message. That way, the message itself can be built using an easily testable pure function.

I think, in trying to use "contexts" as Phoenix guides you to (e.g. with `mix phx.gen context`), I'd think of this as introducing a new `Slack` context, and have all the API calls be functions of `Arrow.Slack`. If that's the case It might look like: `Arrow.Slack.disruption_created(...) |> Arrow.Slack.notify()`.

`Arrow.Slack.disruption_created` would call a function in `Arrow.Slack.DisruptionNotification` as it does now, and `Arrow.Slack.notify()` could maybe use `Application.get_env(:arrow, :slack_notifier)` to decide which module to call:

```ex
defmodule Arrow.Slack do
  # ...

  def notify(msg) do
    notifier = Application.get_env(:arrow, :slack_notifier)
    notifier.notify(msg)
  end
end
```

And what you currently have in `notify` would go in a `Arrow.Slack.HTTPNotifier` module. The dev and test environments could use `Arrow.Slack.ConsoleNotifier` or something like that.

PR #816, opened 2021-12-13
chore: Add submit button for marking disruption approval status on sh…

uchukwu:
…ow page

#### Summary of changes
**Asana Ticket:** [Mark disruption pending/approved from show page](https://app.asana.com/0/584764604969369/1201029740414123/f)

Added new action in DisruptionController for HTML button submit on show page. The HTML button submit allows client to change disruption status from approved to pending and vice versa. 

Screenshot of button with `mark as approved`:

Screen Shot 2021-12-14 at 3 14 30 PM

Screenshot of button switched to `mark as pending` (after clicking on it):

Screen Shot 2021-12-14 at 3 14 46 PM

Screenshot of button switched back to `mark as approved` (after clicking on it):

Screen Shot 2021-12-14 at 3 15 03 PM

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/disruption_controller.ex
+++ b/lib/arrow_web/controllers/disruption_controller.ex
@@ -9,8 +9,17 @@ defmodule ArrowWeb.DisruptionController do
   alias Plug.Conn
 
   plug(Authorize, :create_disruption when action in [:new, :create])
-  plug(Authorize, :update_disruption when action in [:edit, :update])
+  plug(Authorize, :update_disruption when action in [:edit, :update, :update_row_status])
   plug(Authorize, :delete_disruption when action in [:delete])
+
+  @spec update_row_status(Conn.t(), Conn.params()) :: Conn.t()
+  def update_row_status(conn, %{"id" => id, "revision" => attrs}) do
+    {:ok, _} = Disruption.update(id, attrs)
+
+    conn
+    |> put_flash(:info, "Disruption updated successfully.")
+    |> redirect(to: Routes.disruption_path(conn, :show, id))
+  end
 
   @spec index(Conn.t(), Conn.params()) :: Conn.t()
   def index(%{assigns: %{current_user: user}} = conn, params) do
--- a/lib/arrow_web/router.ex
+++ b/lib/arrow_web/router.ex
@@ -47,6 +47,7 @@ defmodule ArrowWeb.Router do
     get("/mytoken", MyTokenController, :show)
     get("/", DisruptionController, :index)
     resources("/disruptions", DisruptionController, except: [:index])
+    put("/disruptions/:id/row_status", DisruptionController, :update_row_status)
   end
 
   scope "/", ArrowWeb do
--- a/lib/arrow_web/templates/disruption/show.html.heex
+++ b/lib/arrow_web/templates/disruption/show.html.heex
@@ -85,7 +85,15 @@
       </div>
 
       <%= if @revision.is_active and Permissions.authorize?(:update_disruption, @user) do %>
-        <div class="col-md-2">
+        <div class="col-md-4">
+          <%= form_tag Routes.disruption_path(@conn, :update_row_status, @revision.disruption_id), method: "put" do %>
+            <%= hidden_input :revision, :row_approved, value: !@revision.row_approved %>
+            <%= submit mark_as_approved_or_pending(@revision.row_approved), class: "btn btn-primary w-100"%>
+          <% end %>
+        </div>
+      <% end %>
+      <%= if @revision.is_active and Permissions.authorize?(:update_disruption, @user) do %>
+        <div class="col-md-3">
           <%= link("edit",
                 class: "btn btn-primary w-100",
                 to: Routes.disruption_path(@conn, :edit, @id)
--- a/lib/arrow_web/views/disruption_view.ex
+++ b/lib/arrow_web/views/disruption_view.ex
@@ -73,4 +73,12 @@ defmodule ArrowWeb.DisruptionView do
   defp update_view_path(conn, %{view: view} = filters, key, value) do
     update_filters_path(conn, %{filters | view: %{view | key => value}})
   end
+
+  defp mark_as_approved_or_pending(true) do
+    "mark as pending"
+  end
+
+  defp mark_as_approved_or_pending(false) do
+    "mark as approved"
+  end
 end
--- a/test/arrow_web/controllers/disruption_controller_test.exs
+++ b/test/arrow_web/controllers/disruption_controller_test.exs
@@ -229,6 +229,27 @@ defmodule ArrowWeb.DisruptionControllerTest do
     end
   end
 
+  describe "update_row_status/2" do
+    @tag :authenticated_admin
+    test "admin can update a disruption row status", %{conn: conn} do
+      %{disruption_id: id} = insert(:disruption_revision, row_approved: false)
+
+      params = %{
+        "revision" => string_params_for(:disruption_revision, row_approved: "true")
+      }
+
+      location =
+        conn
+        |> put(Routes.disruption_path(conn, :update_row_status, id), params)
+        |> redirected_to()
+
+      assert %{row_approved: true} =
+               Repo.get!(DisruptionRevision, Disruption.latest_revision_id(id))
+
+      assert location == Routes.disruption_path(conn, :show, id)
+    end
+  end
+
   defp insert_revision_with_everything do
     insert(:disruption_revision,
       days_of_week: [build(:day_of_week)],


losvedir:
Looks like the branch is out of date. Can you rebase onto the latest master, and also include a screenshot?

losvedir:
Thanks for the screenshot. Can you try styling the button so it looks like the "edit" one? Also, it should say "Mark as Pending" and "Mark as Approved".

lib/arrow_web/views/disruption_view.ex
```diff
@@ -76,4 +76,8 @@ defmodule ArrowWeb.DisruptionView do
 
   defp route_id_uniq("CR-" <> _), do: "CR"
   defp route_id_uniq(route_id), do: route_id
+
+  defp approved_or_pending(status) do
+    if(status == true, do: "Marked as Approved", else: "Marked as Pending")
```
losvedir, line 81: This should be the other way around. If `row_approved` is true, then the button should say to mark it as pending, and if `row_approved` is false, it should say to mark it as approved.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -12,6 +12,37 @@ defmodule ArrowWeb.DisruptionController do
   plug(Authorize, :update_disruption when action in [:edit, :update])
   plug(Authorize, :delete_disruption when action in [:delete])
 
+  @spec approved(Conn.t(), Conn.params()) :: Conn.t()
+  def approved(conn, %{"id" => id}) do
```
losvedir, line 16: Since this toggles the ROW status, maybe a better name for this action and route would be `toggle_row_status`.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -12,6 +12,37 @@ defmodule ArrowWeb.DisruptionController do
...
+    %{revisions: [revision]} = Disruption.get!(id)
+
+    if revision.row_approved == true do
```
losvedir, line 19: I think it would make sense to pull this toggling logic into a function in `Disruption`. So this controller endpoint could be as simple as:

```elixir
def toggle_row_status(conn, %{"id" => id}) do
  :ok = Disruption.toggle_row_status(id)
  redirect(to: Routes.disruption_path(conn, :show, id))
end
```

I can't think of a way toggling the status would fail, so you don't need to handle that error case. The only thing that could go wrong is if `id` doesn't exist, and when that happens `Disruption.get!` raises an error which the app catches at a high level and turns into a `404` response.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -12,6 +12,37 @@ defmodule ArrowWeb.DisruptionController do
   plug(Authorize, :update_disruption when action in [:edit, :update])
```
losvedir, line 12: Can you add the new function to this list? I think it should require `:update_disruption` permissions to toggle the ROW status.

lib/arrow_web/views/disruption_view.ex
uchukwu: Good catch. Made the swap in new commit

lib/arrow_web/controllers/disruption_controller.ex
bklebe: Is there a reason not to do this as a form submission to `update`, and figure out what the toggle state should be client-side? It seems like adding a route for this when the functionality to update a disruption already exists is duplicating functionality.

losvedir: I think that would be another good way to do it. We paired on this some, and that was my first thought, but ultimately it's surprisingly a bit tricky to make work that way.

The main problem is with handling associations and what happens if you submit a request that does not mention any. The default Phoenix behavior is to treat that as "leave them alone". While that would be ideal in *this* case of just wanting to toggle the ROW status, it resulted in a bug a few weeks ago where we were unable to completely remove all the exception dates from a disruption. The semantics of form names and values (e.g. `<input name=disruption[exception][][date]" value="2021-12-25" />` means that it's impossible to submit the ideal format of an empty array.

Consequently, the `update` action is somewhat tailored to working with the form, and handles unspecified associations as deleting them.

So, having this button re-use the `update` route, while consolidating into one route, would introduce some extra complexity there in tweaking how the form works, and maybe introducing a special "delete all" sentinel value to handle the respective use cases. I wouldn't be opposed if the PR went that way, but it doesn't feel like a slam dunk much better way to do it either.

That said, I do like the idea of the client saying what it thinks the value should be, so the form submission is in sync with what's displayed to the user. I wouldn't want a client to load the page, have someone else approve the disruption in the meantime, and then have the original person try to click "approve" only to have it go back to pending.

bklebe: It's not clear to me why the associations would be unspecified — you would specify them as their current value, no?

losvedir: I don't follow. What do you mean marked as un-editable? This is on the show page, where there's not currently any form at all, so if we did go with the "this is really just an update" approach, we'd need to stuff all the disruption's current data in hidden fields (well, all the association data, the direct data you can leave unspecified and that would mean "don't update"). It's... doable, but seems inelegant and like a footgun to me. It also couples the data we need to load for the show page, to the data we need to load for the edit page.

bklebe: > I don't follow. What do you mean marked as un-editable?

I meant hidden form fields with `form_for` and `hidden_inputs_for` like you mention. Apologies, I should have been more precise. 

I've spent a while thinking about why this feels wrong to me. I'm definitely curious about how doing it as an update is a footgun, because this way seems to have more footguns. The biggest issue is the problem you mention about state. The button is currently not quite doing what it claims: it says it is marking a disruption as approved or as pending, but it is actually _toggling_ the approval status. I would expect to get an error if I tried to approve a disruption that was already approved, or at least a no-op like I'd get if I submitted an update like that. (Incidentally, this has exposed that we should probably be doing optimistic locking with all other updates too if we expect Arrow to have concurrent users)

My other concerns are mostly about maintainability and REST API hygiene (not sure if we try to follow REST but it's my preference). On maintainability, we already have a method for updating disruptions, and introducing a different route for updating disruption approval state introduces complexity in the authorization structure (which is what originally caught my eye). On REST, the empty POST here is stateful and not semantic: we have to know the client state in order to do what the client expects us to, and the semantics are defined by the URL structure instead of the HTTP verbs (this could just as easily be a GET as a POST). We get around persisting client state by assuming that the client state is what we see in the database, but that may not always be the case (race conditions are one example outlined above, but they are an instance of a more general problem with this approach).

I still feel that this should be a POST to `/disruptions/:id` and use the existing update action. It will add a bit of complexity client-side, but I think it pays off in terms of keeping the API structure such that it reflects resources and is semantic, and centralizing the logic (e.g. locking) we need to do updates properly.

losvedir: Hm, thinking about this more, I think we can probably reuse the `update` action, without _too_ much tweaking. I think maybe a special "keep_associations" hidden input can skip the logic to delete missing associations.

losvedir:
This works, and is how we talked about doing it in our initial pairing session. So nice work wrapping it up like that. However, from the comment chain (https://github.com/mbta/arrow/pull/816#discussion_r768933359) with Beatrix, I did think of a potential issue with this approach. Consider the following sequence of events:

1. User 1 loads the disruption's show page and sees that it's "pending".
2. They go get a cup of coffee.
3. User 2 loads the disruption's show page and sees that it's "pending".
4. User 2 clicks "Mark as approved".
5. The disruption becomes approved.
6. User 1 sits back down and clicks "Mark as approved".
7. The disruption, which had been approved already, is now toggled *back* to pending.

So I think the idea of the route being just a "toggle", and the controller inferring which status to change to, is a problem. The form should encode the value that the user is trying to change it to. That way, it will always be in sync with the text on the button they're pressing. And in the situation above, the user will simply approve a disruption that is already approved, and not inadvertently unapprove it.

This can be done by giving the form a hidden input with the row status that the user thinks they're changing it to.

We can *almost* do that by just re-using the existing update function, and not needing a special `toggle_row_status` field at all. Try a form like this, I think (but you can make the `true` and `"Mark as approved"` `false` and `"Mark as pending"` based on the current status):

```elixir
<%= form_tag Routes.disruption_path(@conn, :update, @id), method: "put", class: "col-lg-8" do %>
  <%= hidden_input :revision, :row_status, value: true %>
  <%= submit "Mark as approved" %>
<% end %>
```

However, I think what will happen (try it out and let me know!), is that button will correctly update the ROW status, but then also _delete_ any associations (e.g. exception dates).

The reason I think it will delete those is because of the `put_new_assocs` function in the controller. That function will, if the form doesn't mention any associations, put values in the `attrs` of `[]` for them. For example, if the attrs were:

```elixir
%{
  "row_status" => true
}
```

it would change that into

```elixir
%{
  "row_status" => true,
  "adjustments" => [],
  "days_of_week" => [],
  "exceptions" => [],
  "trip_short_names" => []
}
```

before passing them into `Disruption.update`. And the `[]` values mean "there are no (e.g.) exception dates for this disruption, so if a previous version had them, please delete them."

I think we can get around that by doing two things. The first is adding a special value in your form:

```elixir
  <%= hidden_input :revision, :keep_associations, value: true %>
```

and the second is updating `put_new_assocs` to look at that value, and only put all the `[]` values if it's not there:

```
defp put_new_assocs(%{"keep_associations" => true}), do: attrs

defp put_new_assocs(attrs) do
  # existing function and code ....
end
```

Try playing around with that and see if you can make something like that work.

lib/arrow_web/controllers/disruption_controller.ex
losvedir: > My other concerns are mostly about maintainability and REST API hygiene (not sure if we try to follow REST but it's my preference)

I think REST is a good starting point, but in my experience trying to follow it 100% introduces more complexity than it's worth. I think the main benefit of REST is as a shared base of understanding among developers, so someone new to the codebase is already somewhat oriented and probably has some decent guesses about the architecture of the app and what pages and forms correspond to what controllers and functions and such.

But no one else (users, designers, product people) think in terms of REST, and trying to shoehorn designs into that paradigm doesn't always work cleanly. Is "liking" something an update to a comment? Creating a "like" resource? What about, say, a drag and drop interface that lets you sort your collection of books. I gravitate more towards using REST for show/index and for forms, and a sprinkling of specialized routes corresponding to domain actions that happen from a UX perspective. In general, like with functions, I like having a collection of more, simple routes, than fewer, more complex ones. In this case, "approve disruption" kind of toes the line for me. (I agree that my initial "toggle" was a bad idea.)

What if in the future we want to add "mark as approved" to the home page? From a UX perspective, this seems like a pretty straightforward movement of a feature from one page to another. If the code domain concepts track the product domain concepts, this is also straightforward: the concept of "approval" requires essentially just a disruption ID, and so anywhere we're offering a disruption to approve, we'll have that. We just call our "mark as approved" route from the home page. But if we have to serialize all the associations, we _don't_ necessarily have all that information already loaded on the home page.

Or what if we want to "mark multiple as approved" from the homepage list? That's a fairly straightforward extension of marking one. If we have a specified "mark approved" action, then maybe we update the route to be `mark_approved` and the IDs are a parameter list. But if instead, per adherence to the rules of REST, we had a single one which was an update to the disruption resource, how do you update multiple at the same time?

I like to refer to the code behind hex.pm for inspiration on good Phoenix code. Take a look at their [router](https://github.com/hexpm/hexpm/blob/main/lib/hexpm_web/router.ex). I'd say the majority of their routes are RESTful, but there are plenty of specialized routes like `PackageReportController.reject_report` or `OrganizationController.add_seats` or `ReleaseController.publish`.

> I'm definitely curious about how doing it as an update is a footgun,

Specifically, I meant having to serialize the entire state including associations in hidden fields in the show page. Six months from now, when someone is changing the "DisruptionController.update` function, they're obviously going to think of the form on the edit page, but not necessarily the "mark as approved" button over on the show page (but note: if we have a "mark as approved" action, people will probably think of anywhere we, from a product perspective, have a "mark as approved" button, whether that's the show page or the home page). If we add another association, like maybe my notes ticket, and we maintain the same contract that we then have to put all that data in the request, will we remember to add new hidden fields over on the show page? If we forget to, then the "mark as approved" button now deletes that association.

bklebe: > Specifically, I meant having to serialize the entire state including associations in hidden fields in the show page.

> If we forget to, then the "mark as approved" button now deletes that association.

Yes, this all makes sense. I can understand why we wouldn't want to have to maintain all that in two places. One thing we could do that `resources` doesn't do by default (presumably for simplicity's sake) is differentiating `PUT /disruptions/:id` vs `PATCH /diruptions/:id`, the former updating the resource by replacing it with a new version (edit page semantics on associations) and the latter modifying the fields on an existing resource (mark as approved/pending semantics on associations, probably more things in the future).

Insofar as we should only follow any constraint so long as it helps us achieve product goals, I agree that REST is not worth pursuing for its own sake. But in this case, I don't think we're so far from the beaten path of web application development that we gain a lot from deviating. Batch operations like you mention are a good example of something that doesn't fit well in REST, but I also don't think we should be architecting for batch operations right now unless we know we're going to need it soon (it wasn't in the ticket, at least). Either way I'm not going to block merging on this, so long as the toggling issue is fixed. We might want to add a ticket to look at our concurrency control strategy for other updates, though.

uchukwu:
Commented old code out. Will remove once revised code is approved

bklebe:
Heads up that I’m not going to be available to review this until 1/3 and I’m fine with @losvedir’s sign off here.

losvedir:
A couple things:

* Cora has since merged in a PR that has some conflicts here, so why don't you take a stab at trying to rebase onto the latest master branch and resolving them.
* This could use a test. Try looking in `ArrowWeb.DisruptionControllerTest` and seeing if you can add a test for your new controller action.
* The following behavior occurs and is not good: 1) Create or edit a disruption and make sure it has at least one "adjustment location", 2) from the show page, mark it as pending/approved, 3) go back to the edit form and notice that all its adjustment locations have been deleted.

That last behavior comes from `Disruption.update`, and in particular the `DisruptionRevision.changeset` line which does `|> Changeset.put_assoc(:adjustments, Adjustment.from_revision_attrs(attrs))`. The converse behavior that we have to allow for is that we *do* want the form to be able to delete all the associations.

I think we need to change that `put_assoc` line to be instead a function which either calls `put_assoc` as it is now, *if the `attrs` have an `adjustments` key, and to not do anything and just return the changeset as-is, if that key isn't available.

lib/arrow/disruption.ex
```diff
@@ -148,4 +148,28 @@ defmodule Arrow.Disruption do
     )
     |> Arrow.Repo.all()
   end
+
+  # @doc """
```
losvedir, line 152: When the PR is ready for review, commented out code should be deleted.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -9,9 +9,27 @@ defmodule ArrowWeb.DisruptionController do
   alias Plug.Conn
 
   plug(Authorize, :create_disruption when action in [:new, :create])
-  plug(Authorize, :update_disruption when action in [:edit, :update])
+  plug(Authorize, :update_disruption when action in [:edit, :update, :toggle_row_status])
```
losvedir, line 12: I think `toggle` was a good name when the controller automatically toggled the ROW status back and forth. But now that it accepts as a parameter, the value that it should be updated to, maybe `update_row_status` is a better name now? That's more in line with standard terminology where the full update is called `update`.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -9,9 +9,27 @@ defmodule ArrowWeb.DisruptionController do
...
+  #   _disruption = Disruption.toggle_row_status(id)
+
+  #   conn
+  #   |> put_flash(:info, "Disruption updated successfully.")
+  #   |> redirect(to: Routes.disruption_path(conn, :show, id))
+  # end
+
+  @spec toggle_row_status(Conn.t(), Conn.params()) :: Conn.t()
+  def toggle_row_status(conn, %{"id" => id, "revision" => attrs}) do
+    _disruption = Disruption.update(id, attrs)
```
losvedir, line 26: `Disruption.update` returns either `{:ok, _something}` or `{:error, _something_else}`. By assigning this response to `_disruption` we're throwing away that information.

It's better to either handle the success and error cases individually, with something like this:

```elixir
case Disruption.update(id, attrs) do
  {:ok, revision} ->
    # .. handle success case

  {:error, changeset} ->
    # .. handle failure case
end
```

In this case, updating the ROW status is something that can't fail, so we can *assert* that it succeeds (and in the unanticipated case that it doesn't, crash). You can assert by matching against only the success case:

```elixir
{:ok, _} = Disruption.update(id, attrs)
```

This throws away the response (by just matching it to `_`), but it asserts that the shape of the response was a tuple that was `:ok`.

lib/arrow_web/router.ex
```diff
@@ -47,6 +47,7 @@ defmodule ArrowWeb.Router do
     get("/mytoken", MyTokenController, :show)
     get("/", DisruptionController, :index)
     resources("/disruptions", DisruptionController, except: [:index])
+    put("/disruptions/:id/approved", DisruptionController, :toggle_row_status)
```
losvedir, line 50: The URL should maybe be `"/disruptions/:id/row_status"`. (You don't need the `update` in the URL, because that's conveyed by the request being a PUT.

lib/arrow_web/templates/disruption/show.html.heex
```diff
@@ -85,7 +85,19 @@
       </div>
 
       <%= if @revision.is_active and Permissions.authorize?(:update_disruption, @user) do %>
-        <div class="col-md-2">
+        <div class="col-md-4">
+          <%= form_tag Routes.disruption_path(@conn, :toggle_row_status, @revision.disruption_id), method: "put" do %>
+            <%= hidden_input :revision, :row_approved, value: flip_hidden_approved_status(@revision.row_approved) %>
```
losvedir, line 90: "Flipping" the status is the [NOT (`!`) operator](https://hexdocs.pm/elixir/Kernel.html#!/1). So instead of a function, you could do: `value: !@revision.row_approved` here.

lib/arrow_web/views/disruption_view.ex
```diff
@@ -73,4 +73,12 @@ defmodule ArrowWeb.DisruptionView do
   defp update_view_path(conn, %{view: view} = filters, key, value) do
     update_filters_path(conn, %{filters | view: %{view | key => value}})
   end
+
+  defp mark_as_approved_or_pending(status) do
```
losvedir, line 77: The "elixiry" way to do this would be by defining two function clauses here, pattern matching on the status. Try doing it that way.

test/arrow_web/controllers/disruption_controller_test.exs
```diff
@@ -229,6 +229,27 @@ defmodule ArrowWeb.DisruptionControllerTest do
     end
   end
 
+  describe "toggle_row_status/2" do
+    @tag :authenticated_admin
+    test "admin can update a disruption row status", %{conn: conn} do
```
losvedir, line 234: Nice test! 👍

losvedir (changes requested):
Getting there! I think maybe one final round of clean ups and we should be good to merge.

losvedir:
Looks like it needs a rebase again. A couple more PRs have been merged.

Also, don't sell yourself short. I'd call this commit a `feat:` (feature) not a `chore:`!

losvedir (approved):
Looks good to me! 👍

PR #818, opened 2021-12-20
feat: mode/line for adjustment-less disruptions

digitalcora:
**Asana:** [🏹 Disruptions-sans-adjustment have a named route or mode](https://app.asana.com/0/584764604969369/1200681721075719)

Creating a disruption without any adjustments should require selecting a "mode" or "line", which both determines how the UI should handle the disruption (icons, filters, etc.), and indicates to GTFS developers what kind of adjustment is being "requested".

This change introduces the concept of "adjustment kind", an enum of these "mode or line" options. For disruptions without adjustments, the chosen option is stored directly in a new `adjustment_kind` field; for disruptions _with_ adjustments, a kind is derived from each adjustment based on its route. This allows "adjustment-less" and "adjustment-ful" disruptions to be handled similarly w.r.t. adjustment kinds.

Though creating disruptions without any adjustments has been possible since d685ed0, we don't appear to have any of these in production, so this change doesn't attempt to fill in an `adjustment_kind` for them. If such a disruption does exist, it might display oddly, but should not crash anything.

### Notes

* This incidentally completes [🏹 Filtering works with disruptions-sans-adjustments](https://app.asana.com/0/584764604969369/1200726999492016), which I mistakenly assumed was part of this task. Fortunately, I think it was faster this way.

* In a small enhancement from the mockups, I added mode icons to the "mode" radio buttons on the form, matching the style of the "line" radio buttons. This wasn't much effort since we already needed these icons elsewhere, with the only extra one being "Subway" (since it isn't a valid adjustment kind).

* Since, [as before](https://github.com/mbta/arrow/pull/811), this task doesn't specify how adjustment-less disruptions should be displayed on the calendar, I left the "(disruption 123)" titles as-is. These do now have color-coding from the `adjustment_kind` and ROW status, so there is more useful information at a glance.

### Screenshots

Filters | Form | Calendar
--- | --- | ---
Screen Shot 2021-12-20 at 11 23 20 AM | Screen Shot 2021-12-20 at 11 32 44 AM | Screen Shot 2021-12-20 at 11 35 08 AM

#### Reviewer Checklist


assets/src/components/DisruptionForm.tsx
```diff
@@ -3,19 +3,35 @@ import Select from "react-select"
...
+  "green_line_d",
+  "green_line_e",
+  "mattapan_line",
+  "orange_line",
+  "red_line",
+] as const
+
+type Mode = typeof modeAdjustmentKinds[number] | "subway"
+type SubwayLine = typeof subwayLineAdjustmentKinds[number]
+type AdjustmentKind = typeof modeAdjustmentKinds[number] | SubwayLine
```
losvedir, line 23: Hm, I'm not sure this very granular taxonomy is working for me. In particular, with the way "subway" kind of interacts oddly in here. It feels like "adjustment kinds" work two different ways in the form, but are being unified into a single type hierarchy.

I think this is most clear in the subtle distinction between `adjustmentKindForMode` and `adjustmentKindsForMode` and how the former says the adjustment kind for "subway" is `null`, while the latter says it's actually several different adjustment kinds.

I think these are the two different uses of this data:

* When a mode is selected, what `adjustment_kind` value should we send to the backend? For bus, commuter_rail, and silver_line, send those values. For subway, there's a finer grained line that should be selected, so it's null for now.
* When a mode is selected, what adjustments should be selectable below? Here, if subway is selected, we should filter for all the adjustments with an adjustment kind of one of the subways.

But I've been staring at and fiddling with this for a while and I'm not really seeing any different or better organization. Maybe `adjustmentKind` vs `adjustmentKinds` can be teased a bit further apart in the name at least?

losvedir (commented):
Pulled it down and it looks great! Couple questions but otherwise should shortly be ready to approve.

lib/arrow/adjustment.ex
```diff
@@ -63,15 +69,43 @@ defmodule Arrow.Adjustment do
...
+  @doc "Determines the kind of an adjustment."
+  @spec kind(t()) :: atom()
+  def kind(%__MODULE__{route_id: "Blue"}), do: :blue_line
+  def kind(%__MODULE__{route_id: "Orange"}), do: :orange_line
+  def kind(%__MODULE__{route_id: "Red"}), do: :red_line
+  def kind(%__MODULE__{route_id: "Mattapan"}), do: :mattapan_line
+  def kind(%__MODULE__{route_id: "Green-B"}), do: :green_line_b
+  def kind(%__MODULE__{route_id: "Green-C"}), do: :green_line_c
+  def kind(%__MODULE__{route_id: "Green-D"}), do: :green_line_d
+  def kind(%__MODULE__{route_id: "Green-E"}), do: :green_line_e
```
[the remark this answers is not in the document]
digitalcora: There is no route with ID `Green`, so we should never see it as the value of `route_id`.

lib/arrow/adjustment.ex
```diff
@@ -63,15 +69,43 @@ defmodule Arrow.Adjustment do
...
+  Builds an Ecto `where` fragment for selecting adjustments with the given kind. Assumes a named
+  binding `adjustments`. Bus adjustments are not yet supported in general, so we assume there are
+  none in the database.
+  """
+  @spec kind_is(atom()) :: %Ecto.Query.DynamicExpr{}
+  def kind_is(:blue_line), do: dynamic([adjustments: a], a.route_id == "Blue")
+  def kind_is(:orange_line), do: dynamic([adjustments: a], a.route_id == "Orange")
+  def kind_is(:red_line), do: dynamic([adjustments: a], a.route_id == "Red")
+  def kind_is(:mattapan_line), do: dynamic([adjustments: a], a.route_id == "Mattapan")
+  def kind_is(:green_line), do: dynamic(false)
```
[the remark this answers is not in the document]
digitalcora: I went with this because of the symmetry with `kind/1`: there's no such thing as an `Adjustment` whose kind is `:green_line`, so this condition should not match any records. The "green line expansion" (or "GLX" if you will) feels specific to the filters on the index, though the boundaries aren't super distinct right now since the only user of this code is the filters. I could also go either way.

assets/src/components/DisruptionForm.tsx
digitalcora: This is something like the third or fourth iteration of fiddling with this for me, so it's pretty well-fiddled 😆

> When a mode is selected, what `adjustment_kind` value should we send to the backend? For bus, commuter_rail, and silver_line, send those values. For subway, there's a finer grained line that should be selected, so it's null for now.

Correct, and this is the one use of `adjustmentKindForMode`. For this one I think the name basically make sense as-is because it's asking "given this mode, what adjustment kind does it correspond to?" And returning `null` for subway is appropriate, because there is no adjustment kind for that mode — it's "just" a mode, not a mode-that-is-also-a-kind.

> When a mode is selected, what adjustments should be selectable below? Here, if subway is selected, we should filter for all the adjustments with an adjustment kind of one of the subways.

Correct, and this is the one use of `adjustmentKindsForMode`. I think this is the shakier one of the two and I'm open to renaming it, though it sort of makes sense, as it's asking "given this mode, what _kinds of adjustments_ correspond to it?" Which is subtly different from the above question, but I'm not sure how to make that clear via the wording. It's the distinction between "`adjustment_kind`" and "`Adjustment` kinds".

Since there is only one use, I'm thinking maybe I could just extract a bit more of the filtering logic and avoid this naming confusion entirely:

```ts
  const adjustmentMatchesMode = ({ kind }: Adjustment, mode: Mode): boolean =>
    // ...  

  const adjustmentSelectOptions = useMemo(() => {
    return allAdjustments
      .filter((adjustment) => adjustmentMatchesMode(adjustment, mode))
      .map(adjustmentSelectOption)
  }, [allAdjustments, mode])
```

Let me know if you think that's a productive direction.

digitalcora: Re: the type definitions, do you think it would be better if instead of defining the "adjustment kinds that are modes" and "adjustment kinds that are subway lines", and then building the `AdjustmentKind` type out of that, we first defined the `AdjustmentKind` type and then extracted the others from it? I chose the current approach because it avoids duplicating any values, but just for general DRY reasons — it would still be type-safe the other way around.

losvedir: I'd say my gut feeling is that I like the way your code sample there is looking. And in your last question, I lean towards "or just defined them without any reference to it".

However, given that you've already fiddled with this area quite a bit, I don't want to drive by with a suggestion which may not work in practice. Since this is the last outstanding question, I'll go ahead and give an approval, and maybe you can try out the changes here. But if it's not working feel free to just revert them and merge as-is.

losvedir (approved):
🚀

PR #819, opened 2021-12-22
refactor: remove `source_label` table sort

digitalcora:
**Asana:** extra task — noticed this while implementing #818

The ability to explicitly select this sort was removed in #770, but it was still supported and used as the default sort, despite adjustment labels no longer appearing in the table. This changes the default sort to ascending by start date, the only remaining sort other than ID. It's a somewhat arbitrary choice, but we can always change it to ID if feedback indicates that's a more useful default.

We agreed this should be updated somehow during a [Slack conversation](https://mbta.slack.com/archives/CF9QKK2AF/p1632510493307200) a while back, but I can't find any Asana tasks that might have been created from it. The same conversation also mentions making the search feature work with descriptions — for that, see #820.

#### Reviewer Checklist


losvedir (approved):
Thanks!

PR #822, opened 2022-01-03
feat: sort adjustment icons/events consistently

digitalcora:
**Asana:** extra task — noticed this while implementing #818

Following several changes to how we display disruption adjustments, the sorting of adjustment icons and events was a bit obscure and inconsistent:

* In the table view, icons were sorted by adjustment label (a property which is no longer actually displayed in the table)
* In the calendar view, events were also sorted by adjustment label, though with no regard for mode/line
* On the detail page, icons were sorted by the order the adjustments happened to be returned from the database (no order was specified)

This change aligns these places on a single "adjustment kind order", which is the same one established by the filter toggles. Calendar events are additionally grouped by ROW approval status, though we could change this if it's preferred to group approved and unapproved adjustments of the same line/mode together — see screenshots below for how it currently looks.

Since this order is now defined in a single place, this also makes it easier to change in the future without forgetting something.

Table | Detail | Calendar
--- | --- | ---
Screen Shot 2022-01-03 at 1 29 44 PM | Screen Shot 2022-01-03 at 1 29 54 PM | Screen Shot 2022-01-03 at 1 30 25 PM


lib/arrow_web/views/disruption_view/calendar.ex
```diff
@@ -57,7 +57,11 @@ defmodule ArrowWeb.DisruptionView.Calendar do
         start: event_start,
         # end date is treated as exclusive
         end: Date.add(event_end, 1),
-        url: Routes.disruption_path(Endpoint, :show, disruption_id)
+        url: Routes.disruption_path(Endpoint, :show, disruption_id),
+        extendedProps: %{
```
losvedir, line 61: Does `extendedProps` mean something to FullCalendar? The `eventOrder` doesn't mention anything about it. Does it like check in the object's keys and then in a special `extendedProps` field after that?

losvedir (approved):
Thanks! 🎉

lib/arrow_web/views/disruption_view/calendar.ex
digitalcora: `extendedProps` is a designated container for "user data". If you specify a property in `eventOrder` that isn't a direct property of the event, it falls back to looking up the same property on `extendedProps`. Unfortunately it doesn't seem like you can do e.g. `extendedProps.kindOrder` to be unambiguous, it only takes "plain" names.

PR #823, opened 2022-01-07
feat: Add "notes" to disruptions

losvedir:
#### Summary of changes
**Asana Ticket:** [🏹 Notes can be added to a disruption](https://app.asana.com/0/584764604969369/1201330656931525)

This adds the concept of "notes" to the app. Notes can be added from a form on the show page (if the user is an admin), or while creating or updating a disruption.

This PR can be reviewed commit-by-commit. The commits are:

1. Set up the data model. Notes belong to a disruption (not a revision).
2. Add notes from a form on the show page.
3. A refactor of `Disruption.create` and `Disruption.update` to switch to using `Ecto.Multi` in preparation for the next commit which adds notes. I went back and forth here and tried a few different approaches and liked this one the best. The main consideration is that the Ecto docs and the book *Programming Ecto* (which I bought and read while working on this PR......) recommend *not* including associations in a changeset on the parent, if you're only appending a new one, and not managing the collection as a whole. Given that, that points to either a transaction or a Multi, and I found the Multi pretty ergonomic in juggling the various parts of the transaction.
4. Add notes via disruption creation or update. This uses the Multi set up in the previous commit. I thought it was important to maintain the note contents on the page in the event that there was a validation failure on submit. I'm not entirely happy with the extra stuff I had to do to flow that through. I feel like ideally maybe it would be part of the changeset, but given the above point that children shouldn't be a part of the parent changeset in append-only scenarios I didn't want to do that. *Programming Ecto* gave the example of using an "embedded" schema (not backed by the DB) to be a virtual changeset container, which then handled the multiple underlying changesets, but this felt a little too complicated.

For now, I punted on the JS interaction to expose the note form on the show page. I think this can be cleanly and easily done by upgrading our `phoenix_live_view` to `0.17` which includes a `JS` module for this sort of thing. But the PR was already dragging and large, so I carved it off to another ticket.

#### Reviewer Checklist

### DIFF
--- a/assets/src/components/DisruptionForm.tsx
+++ b/assets/src/components/DisruptionForm.tsx
@@ -149,6 +149,7 @@ interface DisruptionFormProps {
   allAdjustments: Adjustment[]
   disruptionRevision: DisruptionRevision
   iconPaths: { [icon: string]: string }
+  noteBody: string
 }
 
 /**
@@ -172,12 +173,14 @@ const DisruptionForm = ({
     tripShortNames: initialTripShortNames,
   },
   iconPaths,
+  noteBody: initialNoteBody,
 }: DisruptionFormProps) => {
   const [isRowApproved, setIsRowApproved] = useState(initialRowApproved)
   const [description, setDescription] = useState(initialDescription)
   const [adjustmentKind, setAdjustmentKind] = useState(initialAdjustmentKind)
   const [hasAdjustments, setHasAdjustments] = useState(adjustmentKind === null)
   const [adjustments, setAdjustments] = useState(initialAdjustments)
+  const [noteBody, setNoteBody] = useState(initialNoteBody)
 
   const [mode, setMode] = useState<Mode | null>(
     initialMode(adjustments, adjustmentKind)
@@ -550,6 +553,20 @@ const DisruptionForm = ({
             </button>
           </div>
         )}
+      </fieldset>
+
+      <fieldset>
+        <legend>
+          notes{" "}
+          <span className="m-disruption-form__legend-optional">optional</span>
+        </legend>
+        <textarea
+          className="w-100"
+          name="revision[note_body]"
+          value={noteBody}
+          onChange={(event) => setNoteBody(event.target.value)}
+          aria-label="note"
+        />
       </fieldset>
     </>
   )
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -10,12 +10,14 @@ defmodule Arrow.Disruption do
   import Ecto.Query
 
   alias Arrow.{DisruptionRevision, Repo}
+  alias Arrow.Disruption.Note
   alias Ecto.Changeset
 
   @type id :: integer
   @type t :: %__MODULE__{
           id: id,
           published_revision: DisruptionRevision.t() | Ecto.Association.NotLoaded.t(),
+          notes: [Note.t()] | Ecto.Association.NotLoaded.t(),
           last_published_at: DateTime.t() | nil,
           inserted_at: DateTime.t(),
           updated_at: DateTime.t()
@@ -24,6 +26,7 @@ defmodule Arrow.Disruption do
   schema "disruptions" do
     belongs_to :published_revision, DisruptionRevision
     has_many :revisions, DisruptionRevision
+    has_many :notes, Note
 
     field(:last_published_at, :utc_datetime)
 
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -12,6 +12,7 @@ defmodule Arrow.Disruption do
   alias Arrow.Disruption.Note
   alias Arrow.{DisruptionRevision, Repo}
   alias Ecto.Changeset
+  alias Ecto.Multi
 
   @type id :: integer
   @type t :: %__MODULE__{
@@ -57,30 +58,35 @@ defmodule Arrow.Disruption do
 
   @doc "Creates a new disruption, with its first revision having the given attributes."
   @spec create(map) ::
-          {:ok, DisruptionRevision.t()} | {:error, Changeset.t(DisruptionRevision.t())}
+          {:ok, %{disruption: t(), revision: DisruptionRevision.t()}}
+          | {:error, :revision, Changeset.t(DisruptionRevision.t()), map()}
   def create(attrs) do
-    Repo.transaction(fn ->
-      %{id: id} = Repo.insert!(%__MODULE__{})
-
-      case DisruptionRevision.new(disruption_id: id)
-           |> DisruptionRevision.changeset(attrs)
-           |> Repo.insert() do
-        {:ok, revision} -> revision
-        {:error, changeset} -> Repo.rollback(changeset)
-      end
+    Multi.new()
+    |> Multi.insert(:disruption, %__MODULE__{})
+    |> Multi.insert(:revision, fn %{disruption: %{id: id}} ->
+      DisruptionRevision.new(disruption_id: id)
+      |> DisruptionRevision.changeset(attrs)
     end)
+    |> Repo.transaction()
   end
 
   @doc "Creates a new revision, with given attributes, of the given disruption ID."
   @spec update(id, map) ::
-          {:ok, DisruptionRevision.t()} | {:error, Changeset.t(DisruptionRevision.t())}
+          {:ok, %{revision: DisruptionRevision.t()}}
+          | {:error, :revision, Changeset.t(DisruptionRevision.t()), map()}
   def update(id, attrs) do
+    Multi.new()
+    |> Multi.insert(:revision, update_revision_changeset(id, attrs))
+    |> Repo.transaction()
+  end
+
+  @spec update_revision_changeset(id, map) :: Changeset.t(DisruptionRevision.t())
+  defp update_revision_changeset(id, attrs) do
     id
     |> latest_revision_id()
     |> DisruptionRevision.get!()
     |> DisruptionRevision.clone()
     |> DisruptionRevision.changeset(attrs)
-    |> Repo.insert()
   end
 
   @doc "Creates a new revision of the given disruption ID with `is_active` set to false."
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -9,8 +9,8 @@ defmodule Arrow.Disruption do
   use Ecto.Schema
   import Ecto.Query
 
-  alias Arrow.{DisruptionRevision, Repo}
   alias Arrow.Disruption.Note
+  alias Arrow.{DisruptionRevision, Repo}
   alias Ecto.Changeset
 
   @type id :: integer
@@ -52,6 +52,7 @@ defmodule Arrow.Disruption do
       ]
     )
     |> Repo.one!()
+    |> Repo.preload([:notes])
   end
 
   @doc "Creates a new disruption, with its first revision having the given attributes."
@@ -150,5 +151,15 @@ defmodule Arrow.Disruption do
       preload: [revisions: {r, ^DisruptionRevision.associations()}]
     )
     |> Arrow.Repo.all()
+  end
+
+  @doc """
+  Inserts a new note associated with the given disruption_id.
+  """
+  @spec add_note(id, String.t(), map) :: {:ok, Note.t()} | {:error, Changeset.t(Note.t())}
+  def add_note(disruption_id, author, params) do
+    disruption_id
+    |> Note.changeset(author, params)
+    |> Repo.insert()
   end
 end
--- a/lib/arrow/disruption.ex
+++ b/lib/arrow/disruption.ex
@@ -57,37 +57,54 @@ defmodule Arrow.Disruption do
   end
 
   @doc "Creates a new disruption, with its first revision having the given attributes."
-  @spec create(map) ::
-          {:ok, %{disruption: t(), revision: DisruptionRevision.t()}}
+  @spec create(String.t(), map) ::
+          {:ok, %{disruption: t(), revision: DisruptionRevision.t(), note: Note.t() | nil}}
           | {:error, :revision, Changeset.t(DisruptionRevision.t()), map()}
-  def create(attrs) do
+  def create(author_id, attrs) do
     Multi.new()
     |> Multi.insert(:disruption, %__MODULE__{})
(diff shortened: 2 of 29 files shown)


lib/arrow_web/templates/disruption/_note.html.heex
```diff
@@ -0,0 +1,11 @@
...
+  <div class="m-note__header">
+    <div class="m-note__author">
+      <%= @note.author %>
+    </div>
+    <div class="m-note__timestamp">
+      <%= date(@note.inserted_at) %>
+    </div>
+  </div>
+  <%= @note.body %>
+</div>
```
digitalcora, line 11: suggestion: Can we use more semantic HTML here instead of it being entirely divs? Off-hand, something like

```html
<article class="mb-4">
  <header class="m-note__header">
    <h4 class="m-note__author"><%= @note.author %></h4>
    <div class="m-note__timestamp"><%= date(@note.inserted_at) %>
  </header>

  <%= @note.body %>
</article>
```

* I use `h4` here arbitrarily but we should use a heading level that's globally appropriate for the disruption page.
* I think `article` makes sense as the root element since the definition of this is broad enough to include comments (user comments are explicitly called out on MDN as something that could be an "article"), but if that feels incorrect, we could use a plain `<section>`.

lib/arrow_web/views/disruption_view.ex
```diff
@@ -85,4 +85,8 @@ defmodule ArrowWeb.DisruptionView do
   defp mark_as_approved_or_pending(false) do
     "mark as approved"
   end
+
+  def date(dt) do
+    DateTime.shift_zone!(dt, "America/New_York") |> Calendar.strftime("%m/%d/%y")
+  end
```
digitalcora, line 91: thought: It's confusing to have both this and `format_date` in the same module (I was originally going to suggest renaming this function to `format_date` to clarify its purpose, before realizing that name was already taken). Since it appears they serve very similar purposes, can we consolidate them?

lib/arrow_web/views/error_helpers.ex
```diff
@@ -14,6 +14,14 @@ defmodule ArrowWeb.ErrorHelpers do
     end)
   end
 
+  @spec changeset_errors(Ecto.Changeset.t()) :: [String.t()]
+  def changeset_errors(changeset) do
```
digitalcora, line 18: nitpick: This should have a `@doc`.

thought: I wonder if the name could be improved. `changeset_errors` is not really informative as to the "shape" of what it returns — you might think it just does `changeset.errors`, but it's quite a bit more than that.

lib/arrow/disruption.ex
```diff
@@ -57,30 +58,35 @@ defmodule Arrow.Disruption do
 
   @doc "Creates a new disruption, with its first revision having the given attributes."
   @spec create(map) ::
-          {:ok, DisruptionRevision.t()} | {:error, Changeset.t(DisruptionRevision.t())}
+          {:ok, %{disruption: t(), revision: DisruptionRevision.t()}}
+          | {:error, :revision, Changeset.t(DisruptionRevision.t()), map()}
```
digitalcora, line 62: question: Couldn't this technically also return `{:error, :disruption, ..., ...}`? Before we were using `Repo.insert!` for the disruption, so if there was an error it would throw.

lib/arrow/disruption.ex
```diff
@@ -57,30 +58,35 @@ defmodule Arrow.Disruption do
...
+          {:ok, %{revision: DisruptionRevision.t()}}
+          | {:error, :revision, Changeset.t(DisruptionRevision.t()), map()}
   def update(id, attrs) do
+    Multi.new()
+    |> Multi.insert(:revision, update_revision_changeset(id, attrs))
+    |> Repo.transaction()
+  end
+
+  @spec update_revision_changeset(id, map) :: Changeset.t(DisruptionRevision.t())
+  defp update_revision_changeset(id, attrs) do
```
digitalcora, line 97: thought: I can't help reading this as "give me a changeset to update a revision" (which makes no sense, revisions are immutable). Could there be a more intention-clarifying name? Is there a need to extract this function at all, since it only has one use?

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -54,12 +54,12 @@ defmodule ArrowWeb.DisruptionController do
   @spec create(Conn.t(), Conn.params()) :: Conn.t()
   def create(conn, %{"revision" => attrs}) do
     case Disruption.create(attrs) do
-      {:ok, %{disruption_id: id}} ->
+      {:ok, %{revision: %{disruption_id: id}}} ->
```
digitalcora, line 61: suggestion: Since the multi now makes the created disruption itself accessible, this could be `{:ok, %{disruption: %{id: id}}}`

assets/src/components/DisruptionForm.tsx
```diff
@@ -551,6 +554,20 @@ const DisruptionForm = ({
...
+      <fieldset>
+        <legend>
+          notes{" "}
+          <span className="m-disruption-form__legend-optional">optional</span>
+        </legend>
+        <textarea
+          className="w-100"
+          name="note[body]"
+          value={noteBody}
+          onChange={(evt) => setNoteBody(evt.target.value)}
```
digitalcora, line 567: suggestion: This could be the full `event` — we have plenty of space on the line.

lib/arrow/disruption.ex
```diff
@@ -57,29 +57,42 @@ defmodule Arrow.Disruption do
...
     |> Repo.transaction()
   end
 
+  @spec maybe_add_note(Multi.t(), String.t(), map | nil) :: Multi.t()
+  defp maybe_add_note(multi, _author, nil) do
+    Multi.put(multi, :note, nil)
+  end
+
+  defp maybe_add_note(multi, author, note_params) do
+    Multi.insert(multi, :note, fn %{revision: %{disruption_id: id}} ->
```
digitalcora, line 91: suggestion: This could be `fn %{disruption: %{id: id}} ->`

digitalcora (commented):
Definitely liking the `Multi` approach overall.

> I thought it was important to maintain the note contents on the page in the event that there was a validation failure on submit. I'm not entirely happy with the extra stuff I had to do to flow that through. I feel like ideally maybe it would be part of the changeset, but given the above point that children shouldn't be a part of the parent changeset in append-only scenarios I didn't want to do that.

The stuff where we are manually extracting `note_body` from params and passing it around as its own thing also gives me serious pause — it feels very "fragile", in that if we want notes to work even slightly differently in the future, there's a lot that would have to change.

I'm not sure I follow your reasoning about the parent changeset. In the case where we're adding a note at the same time as a create/update, that _isn't_ an "append-only scenario".

> _Programming Ecto_ gave the example of using an "embedded" schema (not backed by the DB) to be a virtual changeset container, which then handled the multiple underlying changesets, but this felt a little too complicated.

I think this is where I'd like to go in the ideal case. With the addition of this feature, a "form submission" is now a distinct thing from "creating a revision" (where before those were equivalent concepts), so I think it _should_ be promoted to a first-class concept that encapsulates both creating a revision and optionally creating a note.

As a simpler solution, have you thought about adding `note_body` as a `virtual` field on disruption revisions? This is a bit architecturally muddled as notes don't have much to do with revisions, but it could allow integrating notes into the current architecture without major changes and without having to extract and handle the note body as a special case (except when needed to actually create the note).

lib/arrow_web/views/disruption_view.ex
losvedir: Yeah, good point. I'll add a clause for datetimes to `format_date` and just use the existing format. I'll bring it up when Jessie does her design review.

losvedir:
I'm not going to have time to address this for a bit, just to clarify:

> I'm not sure I follow your reasoning about the parent changeset. In the case where we're adding a note at the same time as a create/update, that isn't an "append-only scenario".

What I meant was that the parent changeset, the disruption, only needs to deal with a single note at a time. It doesn't manage the whole collection of notes at once the way we do with, say, the "days of week" relationship.

The only way to work with associations directly on a parent changeset is when it knows about the whole set. This would mean pre-loading the existing notes, so that Ecto can diff the IDs and delete, update, or create notes as necessary.

Since that's all a lot of unnecessary work, I opted not to manage the notes relationship that way. That's the approach recommended in *Programming Ecto* and mentioned in the docs [here](https://hexdocs.pm/ecto/Ecto.Changeset.html#put_assoc/4-example-adding-a-comment-to-a-post).

That's all I was trying to get at: going with `Multi` and having a separate `build_assoc` insert, vs trying to add a new association to `disruptions` (or revisions) like we have with `days_of_week`, `exceptions`, etc.

Is that clear? Or did I misunderstand your comment?

digitalcora:
All clear to me now, thanks!

lib/arrow/disruption.ex
losvedir: Ah, well if a tree falls in the forest...

I went back and forth on this. I actually started with including it as part of the return type, but then that made me want to handle it in the controller, which made me realize that I couldn't actually induce that particular failure, so I wouldn't be able to test that branch. So in the end I opted to document what the current value actually is, where `{:error, :disruption, ..., ...}` is not possible.

lib/arrow/disruption.ex
losvedir: No, I suppose it doesn't need to be extracted.

assets/src/components/DisruptionForm.tsx
losvedir: I dunno, letters - like syllables - add up over time...

lib/arrow/disruption.ex
losvedir: Actually, in the `update` case, there won't be a `:disruption` key.

losvedir:
> As a simpler solution, have you thought about adding note_body as a virtual field on disruption revisions? This is a bit architecturally muddled as notes don't have much to do with revisions, but it could allow integrating notes into the current system without major changes and without having to extract and handle the note body as a special case 

Just pushed up a commit which tries this. Take a look and let me know what you think.

I like that I'm not passing around `note_body` as a distinct value all over the place now, so that's a major improvement. However, it is "architecturally muddled", ha, as you put it, and it couples notes and revisions.

I could go either way on it.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -47,37 +50,51 @@ defmodule ArrowWeb.DisruptionController do
...
-        |> put_flash(:errors, {"Disruption could not be created:", errors(changeset)})
-        |> render("new.html", adjustments: Adjustment.all(), changeset: changeset)
+        |> put_flash(
+          :errors,
+          {"Disruption could not be created:", ErrorHelpers.changeset_error_messages(changeset)}
+        )
+        |> render("new.html",
+          adjustments: Adjustment.all(),
+          changeset: changeset
+        )
```
digitalcora, line 75: nitpick: This line is now unchanged, so it shouldn't need to be reformatted.

lib/arrow_web/controllers/disruption_controller.ex
```diff
@@ -47,37 +50,51 @@ defmodule ArrowWeb.DisruptionController do
...
-        |> put_flash(:errors, {"Disruption could not be updated:", errors(changeset)})
-        |> render("edit.html", adjustments: Adjustment.all(), changeset: changeset, id: id)
+        |> put_flash(
+          :errors,
+          {"Disruption could not be updated:", ErrorHelpers.changeset_error_messages(changeset)}
+        )
+        |> render("edit.html",
+          adjustments: Adjustment.all(),
+          changeset: changeset,
+          id: id
```
digitalcora, line 96: nitpick x2 combo: This line is now unchanged, so it shouldn't need to be reformatted.

lib/arrow/disruption.ex
```diff
@@ -49,36 +53,59 @@ defmodule Arrow.Disruption do
...
+  def update(id, author_id, attrs) do
+    Multi.new()
+    |> Multi.insert(:revision, fn _ ->
+      id
+      |> latest_revision_id()
+      |> DisruptionRevision.get!()
+      |> DisruptionRevision.clone()
+      |> DisruptionRevision.changeset(attrs)
+    end)
+    |> maybe_add_note(author_id, note_params(attrs))
```
digitalcora, line 87: thought: Technically I don't think we should need to separately extract the note params like this. If we allow the revision changeset to accept and cast the `note_body`, then it should be present on the virtual field within the `:revision` multi result. Something like this should work:

```elixir
Multi.new()
|> Multi.insert(:revision, ...)
|> Multi.merge(&maybe_add_note(&1, author_id))

# ...

defp maybe_add_note(%{revision: %{note_body: nil}}, _), do: Multi.new()

defp maybe_add_note(%{revision: %{disruption_id: id, note_body: body}}, author_id) do
  Multi.new()
  |> Multi.insert(:note, fn _ -> Note.changeset(id, author_id, %{body: body}) end)
end
```

I'll leave it to you whether you like the trade-offs here better. Just seemed a bit odd to me to have the virtual field but never actually access/assign it. (is it even needed, in this version?)

digitalcora (approved):
Few non-blocking things. Seems ready to go!

lib/arrow/disruption_revision.ex
```diff
@@ -35,6 +36,7 @@ defmodule Arrow.DisruptionRevision do
     field(:row_approved, :boolean, default: true)
     field(:description, :string)
     field(:adjustment_kind, Ecto.Enum, values: Adjustment.kinds())
+    field(:note_body, :string, virtual: true)
```
digitalcora, line 39: suggestion: Add a comment here explaining what this is for and where it's used?

lib/arrow/disruption.ex
losvedir: Ah, I do like that approach.

> Just seemed a bit odd to me to have the virtual field but never actually access/assign it. (is it even needed, in this version?)

Yes, it's used in this version after a validation error to pass the note-in-progress to the frontend in the revision changeset, rather than having to pass it out of band like in the previous version. I changed the form params to `revision[note_body]` and updated the revision's `cast`, so it _is_ getting assigned, just not explicitly.

losvedir: Actually, after implementing it this way, I think I don't like all the Multi juggling as much, so I'm going to go with the current approach.

PR #825, opened 2022-01-12
draft: file uploads and attachments

digitalcora:
**Asana:** [🏹 Design doc for file attachment](https://app.asana.com/0/584764604969369/1201613967362755)

[Quick link to the rendered version](https://github.com/mbta/arrow/blob/e3c10e79c4d3ec4b1f36d8773269f5d6ffafcb38/docs/attachments.md) (v3).

### DIFF
--- a/docs/attachments.md
+++ b/docs/attachments.md
@@ -63,9 +63,17 @@ like there are any good pure-Elixir options for this.
 
 ### Data Model
 
-Since the "types" of attachments we want to support have a lot of behavior in
+Since the "kinds" of attachments we want to support have a lot of behavior in
 common, and easily extending the system to support new types is desired, we can
-use a single Ecto schema for all of them. Below is a sketch of this approach.
+use a single Ecto schema for all of them. This would need only an enum field to
+distinguish between different kinds of attachments.
+
+### Code Sketch
+
+Below is a sketch of the `Attachment` schema module, which also includes the
+logic that would be needed to handle the file storage itself. This specific
+module structure and API is not prescriptive; it's only here to prove the basic
+elements are feasible and give an idea of how much work will be involved.
 
 In this sketch, full storage paths for an attachment are generated when needed,
 and only the original filename is stored in our database — this allows handling
@@ -146,7 +154,7 @@ defmodule Arrow.Disruption.Attachment do
 
   defp path(%__MODULE__{}, variant \\ :original) do
     # Generate a storage path for the file. This could be in the form:
-    #   /<disruption ID>/<variant>/<filename>
+    #   /<env-specific prefix>/<disruption ID>/<variant>/<filename>
     # To be able to use the disruption ID, a new disruption would need to have
     # already been inserted before `Attachment.insert` is called (probably in a
     # multi of its own, to which the Attachment multi is appended).
@@ -167,6 +175,41 @@ defmodule Arrow.Disruption.Attachment do
   defp convert(path, :preview) do
     # Use Mogrify to open the file path, convert it, and save it under a new
     # temp file path, which we return
+  end
+end
+```
+
+Currently the specs for this feature only include being able to upload files
+using a single static file input on the disruption show page, which is not part
+of the disruption edit form. The template code for this could look like:
+
+```eex
+<%= form_for @conn, Routes.attachment_path(@conn, :create, @id),
+      [method: "post", multipart: true], fn f -> %>
+  <%= file_input f, :upload %>
+  <%= hidden_input f, :kind, "miscellaneous" %>
+  <%= submit "upload", class: "btn btn-primary" %>
+<% end %>
+```
+
+With a corresponding AttachmentController having code like this (assuming the
+attachment route is nested under the disruption route):
+
+```elixir
+def create(conn, %{"disruption_id" => id} = params) do
+  result =
+    %Attachment{disruption_id: id}
+    |> Attachment.changeset(params)
+    |> Attachment.insert()
+    |> Attachment.cleanup()
+
+  case result do
+    {:ok, _} ->
+      redirect(conn, to: Routes.disruption_path(conn, :show, id))
+
+    {:error, _operation, _result, _changes} ->
+      # same redirect as above but show an appropriate error flash based on the
+      # operation and/or result
   end
 end
 ```
--- a/docs/attachments.md
+++ b/docs/attachments.md
@@ -29,138 +29,183 @@ This doc outlines how file attachments to disruptions will be handled. 📎
 
 ## Proposal
 
-[Waffle] appears to be the most widely used off-the-shelf solution for handling
-file uploads and downloads in Elixir. On investigation of this library and its
-[Ecto support module][Waffle.Ecto], it appears to offer everything we need for
-this feature, and also checks off some nice-to-haves.
+Since `Plug.Upload` and `ExAws.S3` already handle most of the "hard parts" of
+accepting file uploads, storing them in S3, and generating URLs for them, it's
+an entirely reasonable option for us to glue them together ourselves, without a
+third-party library. Libraries are, however, explored in the **Alternatives**.
 
-The docs have a [complete example module][example] which includes dynamic
-storage paths, validation, and producing multiple "versions" using ImageMagick,
-which might be useful to refer to. This does not include Ecto integration.
+### Relevant APIs
 
-[Waffle]: https://hexdocs.pm/waffle/Waffle.html
-[Waffle.Ecto]: https://hexdocs.pm/waffle_ecto/Waffle.Ecto.html
-[example]: https://hexdocs.pm/waffle/s3.html
+* When a file is uploaded from a `Phoenix.HTML.Form.file_input/3`, Phoenix
+  automatically receives and stores the file in a temporary location on disk
+  that is deleted when the request process dies, and generates a `Plug.Upload`
+  struct for it in the request's `params`. This has `filename`, `content_type`,
+  and `path` fields.
 
-### Terminology
+* `ExAws.S3.put_object/4` allows uploading a binary to a specified path.
+  Initially this might be the simplest way to handle uploads, but if we need to
+  account for very large files, there is also `ExAws.S3.upload/4` which allows
+  streaming data directly from disk (instead of reading the whole file into a
+  binary in memory first).
 
-**"Definition module"** is a module which uses `Waffle.Definition` and
-conceptually represents a "kind of upload" (an "avatar" is the example used in
-the docs). Such modules gain an API to `store`, `delete`, and retrieve the `url`
-of a file, and can override several callbacks to control how those things happen
-or add validations, transformations, or "versions" (variants, e.g. thumbnails).
+* `ExAws.S3.delete_object/3` can delete the objects we `put`.
 
-Filename is the primary identifier used by Waffle. **"Scope"** refers to any
-Elixir term passed, together with a filename, to one of the definition module's
-functions. The filename-plus-scope then flows through to the callbacks. For
-example, if an Ecto "user" struct were used as the scope, this would allow
-storing avatars under different paths per user by incorporating the user ID,
-instead of all avatars occupying a single global namespace.
+* `ExAws.S3.presigned_url/5` generates a pre-signed URL for an object (enabling
+  it to be accessed without authentication for a limited time).
 
-### Storage
+### Variants
 
-Waffle has a first-party S3 storage adapter which uses ExAws for configuration,
-the same AWS library we already use. Storage can be configured per environment,
-allowing us to use local storage for testing and S3 storage when deployed:
-
-```elixir
-# config/config.exs
-config :waffle, storage: Waffle.Storage.Local
-
-# config/prod.exs
-config :waffle, storage: Waffle.Storage.S3, bucket: {:system, "S3_BUCKET"}
-```
-
-Most storage parameters (path, ACLs, HTTP headers) can be set either globally in
-the app config or per-attachment using callbacks.
+For generating "variants" of image attachments, such as thumbnails, we can use
+the [Mogrify](https://hexdocs.pm/mogrify/) library, a convenient wrapper around
+the ImageMagick command-line tool. This would require adding a new system-level
+dependency both for developers and in the production images, but it doesn't seem
+like there are any good pure-Elixir options for this.
 
 ### Data Model
 
 Since the "types" of attachments we want to support have a lot of behavior in
 common, and easily extending the system to support new types is desired, we can
-use a single Ecto schema and definition module for all of them. In principle it
-even appears possible for both of these to be the same module — the docs use
-separate modules since their example of a "user" who has an "avatar" involves
-two separate concepts.
+use a single Ecto schema for all of them. Below is a sketch of this approach.
 
-Here is what a combined approach might look like:
+In this sketch, full storage paths for an attachment are generated when needed,
+and only the original filename is stored in our database — this allows handling
+variants without much extra effort, since the desired variant can be part of the
+call to generate the path (this mirrors how Waffle works).
+
+Comments tagged with `[v]` indicate logic that would only be needed once we
+implement support for image variants.
 
 ```elixir
 defmodule Arrow.Disruption.Attachment do
   use Ecto.Schema
-  use Waffle.Definition
-  use Waffle.Ecto.Definition
-  use Waffle.Ecto.Schema
 
   @kinds ~w(miscellaneous cr_schedule shuttle_diagram signage_plan)
-  @versions ~w(original preview)a
+
+  # Module that implements a common storage adapter behaviour, using `ExAws.S3`
+  # in production and the local file system in dev/test. The callbacks could be
+  # `put`, `delete`, and `url`, each accepting a file path.
+  @storage Application.compile_env(:arrow, :attachment_storage)
 
   schema "disruption_attachments" do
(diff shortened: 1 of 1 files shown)


docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+## Drawbacks
+
+* With the `Waffle.Ecto` integration, `cast_attachments` immediately stores the
+  file data to S3 if it is valid. If something goes wrong later in the process
+  and the Attachment record is not persisted, there is now a "disconnected"
+  file in S3, and we need to account for this and do the cleanup ourselves.
+  This is in contrast to alternatives (see below), in which casting/validation
+  is not necessarily coupled to uploading.
+
+* The general quality of Waffle's code, APIs, and documentation seems a bit
```
losvedir, line 151: I really appreciate your brief audit of the code here! From the outside it seems like Waffle ticks all the boxes, but it's good to be aware of the potentially shaky foundation.

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+doesn't seem out of the question for us to write all the glue code ourselves.
+
+Below is a sketch of this approach, taking cues from Waffle. For error handling,
+`insert` and `delete` could wrap their operations in `Multi` steps; the module
+could then provide a `cleanup` function that accepts and "passes through" a
+transaction result and, if the result was failure, uses the Multi data to "roll
+back" the upload steps by deleting the objects from S3.
+
+In some ways, this approach appears cleaner than either Waffle or Capsule,
+though of course high-level sketches always _appear_ clean. In any case it is
```
losvedir, line 194: Ha, very true. Though conversely, using libraries like Waffle can often appear clean at first, but end up being a maintenance burden down the road if the author abandons it, or if we want to do something outside of what it provides. Tricky balance.

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+an intriguing option.
+
+```elixir
+defmodule Arrow.Disruption.Attachment do
+  use Ecto.Schema
+
+  @kinds ~w(miscellaneous cr_schedule shuttle_diagram signage_plan)
```
losvedir, line 201: A nice benefit of having all the kinds of attachments specified here, is I think the various forms on the page would just have to specify the kind as a hidden parameter, but would otherwise use the same controller endpoint.

I don't think this would complicate things too much on the display. As far a I can tell, we'd just filter all the attachments based on `kinds` to display the right ones in the right place on the show page.

losvedir (commented):
This was a great write-up. I think I went into it expecting that we'd just use Waffle, but I'm actually pretty open to the idea of just implementing ourselves the bits that we need. Especially since resizing and other variants aren't really on the near term roadmap, and since this approach _can_ handle them, if we need.

If we do go with implementing ourselves, I wonder about doing it with a slightly different structure. Drawing inspiration from contexts and Ecto, I'm thinking of putting the `@storage` adapter in the context along with the integration functions like `insert`, with pure side of things in the module. The `changeset` could go there, and maybe the various `insert` and delete` functions could return an `ExAws` request.

Though maybe because of the S3 interaction it won't work out quite so neatly that way. It also points to either further complicating `Arrow.Disruption` (and so maybe as a pre-req we should finally move the schema stuff into `Arrow.Disruption.Disruption`), or thinking about another `Arrow.Attachments` context.

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+This doc outlines how file attachments to disruptions will be handled. 📎
+
+
+## Requirements
+
+* Attachments are stored in AWS S3 (at least in production)
+* Attachments on different disruptions can have the same filename
+* Attachments can be deleted/"unattached" from their disruption
+* Disruptions can have multiple attachment "kinds" (signage plan, shuttle
+  diagram, etc.) and attachments can have different validations per kind
```
bklebe, line 12: I'm curious, what sorts of validations are we imagining here, and how do kinds relate to them? The file is going to be a binary upload, so I assume we're not talking about validating its structure, maybe just its size and format and things like that?

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
+# Attachments RFC
+
+This doc outlines how file attachments to disruptions will be handled. 📎
+
+
+## Requirements
+
+* Attachments are stored in AWS S3 (at least in production)
+* Attachments on different disruptions can have the same filename
+* Attachments can be deleted/"unattached" from their disruption
```
bklebe, line 10: Does this mean deleting it from the persistence layer or just removing the association?

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+conceptually represents a "kind of upload" (an "avatar" is the example used in
+the docs). Such modules gain an API to `store`, `delete`, and retrieve the `url`
+of a file, and can override several callbacks to control how those things happen
+or add validations, transformations, or "versions" (variants, e.g. thumbnails).
+
+Filename is the primary identifier used by Waffle. **"Scope"** refers to any
+Elixir term passed, together with a filename, to one of the definition module's
+functions. The filename-plus-scope then flows through to the callbacks. For
+example, if an Ecto "user" struct were used as the scope, this would allow
+storing avatars under different paths per user by incorporating the user ID,
```
bklebe, line 57: Does this mean the user ID would be leaked as part of the S3 path? Do we consider user IDs sensitive currently?

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+
+[Waffle]: https://hexdocs.pm/waffle/Waffle.html
+[Waffle.Ecto]: https://hexdocs.pm/waffle_ecto/Waffle.Ecto.html
+[example]: https://hexdocs.pm/waffle/s3.html
+
+### Terminology
+
+**"Definition module"** is a module which uses `Waffle.Definition` and
+conceptually represents a "kind of upload" (an "avatar" is the example used in
+the docs). Such modules gain an API to `store`, `delete`, and retrieve the `url`
```
bklebe, line 49: One question I had from looking at Waffle myself was whether it makes it easy to get the object key and the other metadata needed to retrieve something from S3 directly ourselves should we find the need to do that in the future. I didn't see anything about this in the docs but it seems like a weird oversight. Did you find anything about this?

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+two separate concepts.
+
+Here is what a combined approach might look like:
+
+```elixir
+defmodule Arrow.Disruption.Attachment do
+  use Ecto.Schema
+  use Waffle.Definition
+  use Waffle.Ecto.Definition
+  use Waffle.Ecto.Schema
```
bklebe, line 93: This is a lot of `__using__`s for what would appear at first glance to be a fairly simple library. `use Waffle.Ecto.Schema` looks like it just adds an import, while `use Waffle.Ecto.Definition` is more complicated: 

https://github.com/elixir-waffle/waffle_ecto/blob/4d7c153351072542283e446d0ffcf76ccdfd41cd/lib/waffle_ecto/schema.ex#L34-L38

https://github.com/elixir-waffle/waffle_ecto/blob/4d7c153351072542283e446d0ffcf76ccdfd41cd/lib/waffle_ecto/definition.ex#L41-L80

`use Waffle.Definition` calls a bunch of other `__usings__`, which makes me sort of uncomfortable in terms of trying to figure out what's going on: https://github.com/elixir-waffle/waffle/blob/108559e055e68def6f273fd6c71db41ecd82808b/lib/waffle/definition.ex#L24-L33

It doesn't actually look like any of these is that complicated of itself, but I don't love the amount of indirection happening here for what could otherwise be a few composable functions, given the underlying operation (upload something to S3 and update the database) is fairly simple.

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+    "attachments/#{id}/#{version}"
+  end
+
+  # example of validation per kind of attachment
+  def validate({%{file_name: name}, %__MODULE__{kind: "cr_schedule"}}) do
+    case name |> Path.extname() |> String.downcase() in ~w(pdf xlsx) do
+      true -> :ok
+      false -> {:error, "must be a PDF or Excel sheet"}
+    end
+  end
```
bklebe, line 122: Ah, I see this somewhat answers my question about what validations entail.

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+  use Waffle.Ecto.Definition
+  use Waffle.Ecto.Schema
+
+  @kinds ~w(miscellaneous cr_schedule shuttle_diagram signage_plan)
+  @versions ~w(original preview)a
+
+  schema "disruption_attachments" do
+    belongs_to :disruption, Arrow.Disruption
+
+    field :kind, :string
```
bklebe, line 101: I realize this is just a partial example, but any chance we could make this an enum or a lookup table or something else we could enforce at the level of the DB schema when we implement it, instead of a string?

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+
+  def transform(:preview, _), do: :skip
+end
+```
+
+Note using a record ID as part of the storage path requires that the record be
+persisted before the attachment is casted/saved. In the above example the
+disruption ID is used rather than the attachment ID, so this would only come
+into play when a disruption is newly created. The recommended solution is to
+use `Multi`, which does mean we can't use `cast_assoc` directly.
```
bklebe, line 139: Yes, thinking about this, I feel like I'd rather have things stored in S3 under an ID generated by the database before uploading, and rollback if the upload fails, than the other way around.

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+## Drawbacks
+
+* With the `Waffle.Ecto` integration, `cast_attachments` immediately stores the
+  file data to S3 if it is valid. If something goes wrong later in the process
+  and the Attachment record is not persisted, there is now a "disconnected"
+  file in S3, and we need to account for this and do the cleanup ourselves.
```
bklebe, line 147: 😱 yikes! Definitely think we want a stronger guarantee (i.e. atomicity) about the validity of these associations than this would provide out of the box.

docs/attachments.md
bklebe: I concur with the assessment of Waffle's code quality. I think I would feel better if the design of Waffle wasn't so dependent on all those macros, and if it provided an easier interface to the backing store (S3). 

Before I heard about Waffle I assumed we were just going to build this ourselves, because it seems fairly straightforward as a feature. I checked and Waffle is a little under 1200 lines of code not including tests. I think we could probably write something for our specific situation that isn't quite as general in much fewer lines, and I think it would be easier to test and I would have more confidence in its long-term maintainability. It's important to have a lot of control over our persistence layer.

docs/attachments.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+## Alternatives
+
+### Capsule
+
+[Capsule](https://github.com/elixir-capsule/capsule) appears to be the only
+other file upload library on Hex with significant use (1k recent downloads,
+versus Waffle's 60k). Its author describes it as "experimental and still in
+active development ... Use at your own risk." Despite this, it seems to have a
```
bklebe, line 170: If this means that it could fully break at any time and it's not expected to be stable, that seems like a fairly strong strike against using it, despite the superior architecture. It also looks like it hasn't seen any new commits in over a year.

bklebe (commented):
Count me in for attempting to do this ourselves, provided we can find a reasonable MVP. I expect that the first run won't have ImageMagick or anything fancy but the basic problem of uploading something to S3 and adding a relation specifying its location seems like something we're more than capable of tackling.

docs/attachments.md
digitalcora: The task for deleting files says this much:

> There is a "delete" link next to each miscellaneous file, which if clicked disassociates the file from the disruption and deletes it in S3.

So, yes, deleting it from everywhere.

docs/attachments.md
digitalcora: This is an example involving "avatars", which is the Waffle documentation's preferred example. In our case we would not be incorporating user IDs in the paths, as files are associated with disruptions, not users.

docs/attachments.md
digitalcora: Waffle doesn't have a convenience function for this, but all you need to retrieve the file from S3 is the bucket and object path (and the permissions to do so). We know the bucket from our own configuration, we know the filename since that's what is stored in the Waffle-serialized field, and we know the rest of the object path from calling `storage_dir` with the filename and the appropriate scope. So assuming we had some `attachment` and wanted to download its file, this would be:

```elixir
prefix = Attachment.storage_dir(:original, {attachment.file, attachment})
object = Path.join(prefix, attachment.file.file_name)
ExAws.S3.download_file(bucket, object, "some/dir")
```

docs/attachments.md
digitalcora: I have no disagreement that it's an unnecessarily convoluted library. I alluded to this in the Drawbacks section.

docs/attachments.md
digitalcora: It's a lot of friction to change the possible values for a database-level enum — adding a new value requires a migration, and _removing_ a value is harder (see here: ["Existing values cannot be removed from an enum type"](https://www.postgresql.org/docs/current/datatype-enum.html)). What we should do is use `Ecto.Enum`, and in fact we already do for `DisruptionRevision.adjustment_kind`, but I didn't think to do it here since it wasn't a very important part of the example.

bklebe: Oh nice, `Ecto.Enum` slipped my mind.

docs/attachments.md
bklebe: Agreed.

docs/attachments.md
bklebe: Good to know it's that easy.

digitalcora:
@losvedir:
> I think I went into it expecting that we'd just use Waffle, but I'm actually pretty open to the idea of just implementing ourselves the bits that we need.

@bklebe:
> Before I heard about Waffle I assumed we were just going to build this ourselves, because it seems fairly straightforward as a feature. I checked and Waffle is a little under 1200 lines of code not including tests. I think we could probably write something for our specific situation that isn't quite as general in much fewer lines, and I think it would be easier to test and I would have more confidence in its long-term maintainability.

So, the current proposal as written is to go with Waffle, but it seems like roll-our-own is a strong contender (I have no strong opinion but I do agree with the assessment of its viability and advantages). Could you both react to this comment with 👀 if you think I should revise the proposal to roll-our-own, or 👍 if you think we should accept it as-is?

digitalcora:
This is now revised with the main proposal being to roll-our-own.

Also, quick response to a part of @losvedir's main comment I didn't address before:

> If we do go with implementing ourselves, I wonder about doing it with a slightly different structure. Drawing inspiration from contexts and Ecto, I'm thinking of putting the `@storage` adapter in the context along with the integration functions like `insert`, with pure side of things in the module. The `changeset` could go there, and maybe the various `insert` and `delete` functions could return an `ExAws` request.
>
> Though maybe because of the S3 interaction it won't work out quite so neatly that way. It also points to either further complicating `Arrow.Disruption` (and so maybe as a pre-req we should finally move the schema stuff into `Arrow.Disruption.Disruption`), or thinking about another `Arrow.Attachments` context.

This sounds broadly reasonable — I don't think I have strong opinions on the module structure of this feature. The example I've written in the design doc has everything in one module because I think it's easier to understand at a glance that way, and we know we'll at least need a schema module somewhere. And if we start out writing it this way, it should be amenable to refactoring without necessarily boxing us into any corners.

docs/attachments.md
```diff
@@ -0,0 +1,222 @@
...
+    # ...
+  end
+
+  def url(%__MODULE__{}, variant \\ :original) do
+    # Call `path/2` and pass the result to `@storage.url`.
+  end
+
+  defp path(%__MODULE__{}, variant \\ :original) do
+    # Generate a storage path for the file. This could be in the form:
+    #   /<disruption ID>/<variant>/<filename>
```
losvedir, line 149: Just to be precise, I think this should be `/<arrow-env>/<disruption ID>/<variant>/<filename>`.

Regarding the bullet point in the asana ticket for the path:

> What path in S3 should be used? Should we use the provided file name, or a hash of some sort? How do we prevent collisions?

what about the collisions question? That is, what happens if someone uploads a new file to the disruption with an existing filename? I believe with this path process, it will overwrite the old one, and I think we'll then have two attachment records with the same backing file.

I think we could go two different ways:
* Hash the file or generate a random value, and use that in the path, so that two files with the same name will be added and kept distinct.
* Treat uploading a new file with the same name as a desire to replace the old one.

Let's ask product/design about that. But I think we have to do something here, because the current behavior is kind of a weird mix of both of those: the disruption will appear to have two files attached, both with the same name, and the old one will have been secretly updated to the new one.

losvedir: Or third approach, validation error.

digitalcora: My assumption was that the validations should prevent uploading a file with the same name as another. This would be straightforward to enforce with a database uniqueness constraint on `(disruption_id, filename)`, and Ecto has built-in support for converting constraint errors into changeset errors.

digitalcora:
> Can you add a brief section including what some of the view and the controller work might look like? I think it's pretty straightforward and have had an approach in mind, but just realized maybe we should get that into the design doc just to make sure we're all on the same page.

Done and tagged as v3. Also addressed your comment about the path.

digitalcora:
Closing as final! Whoever implements the attachment tasks can use this as a guide.

PR #829, opened 2022-01-26
fix: Use email address for Cognito ID

losvedir:
#### Summary of changes
**Asana Ticket:** n/a

When testing out the Notes feature on arrow-dev and finally exercising the real Cognito flow, it turns out that the "user ID" was the Cognito internal ID, which for ActiveDirectory currently has the form `ActiveDirectory_MBTA\[username]`. In the mock-ups and generally throughout the app, we'll want to tag actions with the email address.

I opted to change the username to the email address rather than keeping it with the cognito pool user ID and adding email as a field for simplicity. Since we're just annotating a field, rather than maintaining a full users table (with all the pros & cons that would entail), the ID itself is fairly meaningless while the email address is useful if you want to know how to contact someone who left a comment or (in the future) made a change to the disruption. While you can generally infer the email from the cognito ID currently, that ID format itself has changed 3 times since we've used Cognito, and may in the future as well. And non-Active Directory cognito accounts, like in Alerts-UI, just have a UUID in that field.

This branch is on dev now.

Screen Shot 2022-01-26 at 8 42 54 AM

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/auth_controller.ex
+++ b/lib/arrow_web/controllers/auth_controller.ex
@@ -4,7 +4,7 @@ defmodule ArrowWeb.AuthController do
 
   @spec callback(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def callback(%{assigns: %{ueberauth_auth: auth}} = conn, _params) do
-    username = auth.uid
+    username = auth.info.email
     expiration = auth.credentials.expires_at
     credentials = conn.assigns.ueberauth_auth.credentials
 
--- a/test/arrow_web/controllers/auth_controller_test.exs
+++ b/test/arrow_web/controllers/auth_controller_test.exs
@@ -6,10 +6,13 @@ defmodule ArrowWeb.Controllers.AuthControllerTest do
       current_time = System.system_time(:second)
 
       auth = %Ueberauth.Auth{
-        uid: "foo@mbta.com",
+        uid: "ActiveDirectory_MBTA\\foo",
         credentials: %Ueberauth.Auth.Credentials{
           expires_at: current_time + 1_000,
           other: %{groups: ["test1"]}
+        },
+        info: %Ueberauth.Auth.Info{
+          email: "foo@mbta.com"
         }
       }
 


bklebe:
I'm not sure how I feel about this with regards to data integrity, usability, and maintainability. This is not resilient if somebody's name (and thus email) changes. Granted, neither is the ActiveDirectory path. So I agree that we shouldn't be showing users the ActiveDirectory path, but we should probably have a robust way of tying notes back to authors internally. Problems like this suggest to me that it's time for us to have a Users table, or, since we are currently using Cognito as our source of user information, to store the user key and query Cognito for user information at the application level when we need to display it.

bklebe (commented):
edit: reposted as a review

I'm not sure how I feel about this with regards to data integrity, usability, and maintainability. This is not resilient if somebody's name (and thus email) changes. Granted, neither is the ActiveDirectory path. So I agree that we shouldn't be showing users the ActiveDirectory path, but we should probably have a robust way of tying notes back to authors internally, so that we can be sure to display correct user information in our UI. Problems like this suggest to me that it's time for us to have a Users table, or, since we are currently using Cognito as our source of truth, to store the user key and query Cognito for user information at the application level when we need to display it.

losvedir:
I don't think the juice of a users table is worth the squeeze. I get that it may be formally correct in some relational sense, but what is the actual, user-facing problem we're solving here? If comments are annotated with an email address, that should be enough for an interested party to reach out to whoever left the recent comment, or (upcoming) to ask why a change was made.

Regardless, it's not *really* even technically feasible anyway. There is no "user key" that we can use here, at least that would result in better user-facing behavior. If we stored `ActiveDirectory_MBTA\gdurazo` and tried to use that to get at the email address, then that would fail when the ADFS format changes (that is the 3rd form of the key that I've seen so far). Not to mention, we've switched out the entire Cognito pool in prod once before already.

The most resilient information we have is the user's email address (which I haven't seen change), so we could potentially key our users table with that, but then at that point... what's the point?

bklebe:
> I get that it may be formally correct in some relational sense, but what is the actual, user-facing problem we're solving here?

My first-order concerns are name changes and data deletion, not relational correctness (thought there is clearly a missing dependency here).  Hopefully if somebody's name (and thus email) changes, IT will forward any messages that get sent to their old email and not reassign it to someone else, but I don't know what their policies are here. And it would still be bad to display outdated information, for a number of reasons. If somebody leaves the T, we have no robust way of eliminating their email from the system if it's been changed, or replacing it with something that says they can't be contacted anymore. Some people may also consider their old name to be private. Above all I think it defies user expectations to have notes (and disruption changes) associated with the email address they had when they made a change instead of their identity as a user.

> Regardless, it's not really even technically feasible anyway. There is no "user key" that we can use here, at least that would result in better user-facing behavior.

This seems like a much bigger problem. If we don't have a durable primary key for users, how do we reassign everyone to the correct authorization group in Cognito when it changes?

losvedir:
This is a bit of a sticky wicket (I hope I'm using that right. I don't know cricket, but I like that sound of that phrase!). I think to do it "right" as you bring up, is quite a lift: we'd have to work with the infrastructure team to try to get our ActiveDirectory - Cognito integration updated to expose some sort of employee-persistent identifier distinct from email address. No one particularly likes working on that integration because it's in the hands of ITD to some extent, and no one (us or them) totally understands how it all works. See, for example, [this ongoing conversation](https://mbta.slack.com/archives/G01E809HUF2/p1643915063353229) in our infrastructure channel about ADFS / Cognito sign out behavior which is being weird right now. Assuming we could successfully figure out how to expose that information, something of the app, the AWS Cognito client settings, and/or ueberauth_cognito would need to be updated to surface it. And then we'd need an internal users table to tie that ID to an email address and that table would have to be kept in sync maybe through a periodic polling process. (Or maybe we do without the table, just associate the note with new ID directly, and fetch the email address as needed when displaying on the page, with some sort of passthrough caching mechanism).

All that to say, I think doing it "right" is off the table, at this point, especially in light of the decreased prioritization of Arrow.

So, given that, let's get the easy concerns out of the way. :-) 

> Hopefully if somebody's name (and thus email) changes, IT will forward any messages that get sent to their old email and not reassign it to someone else, but I don't know what their policies are here. And it would still be bad to display outdated information, for a number of reasons. If somebody leaves the T, we have no robust way of eliminating their email from the system if it's been changed, or replacing it with something that says they can't be contacted anymore.

Yes, this is a subpar UX, though that's not a dealbreaker per se. Given an email address and a timestamp, that should be enough information for someone to get ahold of whoever they need to, whether by emailing it directly, or asking around to see what the deal is.

> This seems like a much bigger problem. If we don't have a durable primary key for users, how do we reassign everyone to the correct authorization group in Cognito when it changes?

Heh, in the past we... haven't. I went in and manually applied new groups to those who needed them or asked for them. I think the canonical answer here, if there are too many users for that approach, is a lambda that fires and copies users between pools when they log in.

> Some people may also consider their old name to be private.

I take the privacy concern seriously, and I think this is the biggest issue. I think the question is whether it's enough to torpedo notes entirely or if we can move forward as-is here, with tickets in the backlog for the "right" approach (which, admittedly, we may never get to or not for a while). 

To try to contextualize the problem, I'm thinking of this feature of leaving notes about a disruption as replacing emails between people or maybe a slack conversation. In the former case, I think if someone changes their AD information, past emails won't necessarily update (especially salutations or closings in the body), and would be exposed in a FOIA request. In the latter, changing their AD information won't *automatically* update their Slack account or the history there. The user can go into Slack and update their information, and that would retroactively apply, but to me that's similar to someone sending us a request to update their information in Arrow, which we could do via ssh'ing to the bastion host, connecting via the credentials in DevOps 1password, and running an ad-hoc DB query, which of course we'd do.

So, I guess, I see it as no worse than status quo, with some benefits to Arrow. But obviously everyone thinks about the pros and cons differently, and I appreciate you raising this issue which honestly hadn't occurred to me. I'm still inclined towards merging this as-is rather than scrapping notes and revision authorship tagging, but we could also escalate this question to Jessie to make the final call, if you disagree with that.

bklebe:
First off, I'm sorry for not surfacing this sooner. Like you said it's too late to do this right for this cycle, and it's clear I should have identified these problems at refinement. It wasn't until I reviewed this PR that I fully understood what was going on with Cognito and Active Directory in Arrow, and that's on me. I assumed we already had a UUID for users.

I would prefer for Jessie to make the final call here. Ultimately it's no worse to be storing AD emails than AD usernames with regards to the factors I've identified, but I'm not comfortable doing either of them this way.

> The user can go into Slack and update their information, and that would retroactively apply, but to me that's similar to someone sending us a request to update their information in Arrow, which we could do via ssh'ing to the bastion host, connecting via the credentials in DevOps 1password, and running an ad-hoc DB query, which of course we'd do.

I think this process is a bug. We don't want to be SSHing into prod and manually running queries against the prod database. We also don't have the query written to do this. At minimum, we should document the process, write the query, and test it if we decide to move forward with these features as-is.

losvedir:
Per [this slack discussion](https://mbta.slack.com/archives/CF9QKK2AF/p1644339487973409) we've decided to just hide the notes feature for now. Closing in favor of https://github.com/mbta/arrow/pull/834 .

PR #885, opened 2022-09-19
💳 Update Arrow to pull secrets into environment variables

bfauble:
#### Summary of changes
**Asana Ticket:** [💳 Update Arrow to pull secrets into environment variables](https://app.asana.com/0/584764604969369/1202095236113170/f)

* remove `SecretsProvider`
* grab env vars directly via System


paulswartz (changes requested):
I'm glad we're doing this, but the implementation isn't quite right.

`config/config.exs` and the files it imports (`dev.exs` `test.exs` `prod.exs`) are read at compile-time. This means that they'll be read when the Docker container is built, not when the container is started in ECS.

The way around this is to use [`config/runtime.exs`](https://elixir-lang.org/getting-started/mix-otp/config-and-releases.html#configuration) for configuration which is read out of the environment. This file is run at startup, and allows pulling in configuration from the runtime environment. For an example, see Trike's [`runtime.exs`](https://github.com/mbta/trike/blob/main/config/runtime.exs).

paulswartz (commented):
Can you also do a test deploy to make sure it still loads?

paulswartz (approved):
:cake: assuming that it can successfully be deployed to `dev`.

config/runtime.exs
```diff
@@ -0,0 +1,11 @@
+import Config
+
+if config_env() == :prod do
+  config :arrow, ArrowWeb.Endpoint, secret_key_base: System.fetch_env!("SECRET_KEY_BASE")
+
+  config :ueberauth, Ueberauth,
+    providers: [
+      cognito:
+        {Ueberauth.Strategy.Cognito, [client_secret: System.get_env("COGNITO_CLIENT_SECRET")]}
```
paulswartz, line 9: question: should this use fetch_env! as well?

mix.exs
```diff
@@ -104,13 +103,4 @@ defmodule Arrow.MixProject do
       ]
     ]
   end
-
-  defp releases do
-    [
-      arrow: [
-        include_executables_for: [:unix],
```
paulswartz, line 111: Suggestion: since we aren't building for Windows, you can keep this line so that we only build the Linux executables.

paulswartz:
Looks good; can you deploy this to dev?

paulswartz (approved):

PR #949, opened 2023-10-23
Chore: Convert workflow to v2

krisrjohnson21:
#### Summary of changes
This PR updates the deploy workflow to use v2 of the `deploy-ecs` workflow. This updated workflow assumes an IAM role instead of relying on long-lasting AWS keys for an IAM user

Note the run failure, which appears to have been happening for a while and is not due to these changes.


paulswartz (approved):
:cake: on the infrastructure changes

PR #950, opened 2023-11-03
wip: logout with ueberauth_oidc

paulswartz:
#### Summary of changes
Support logging out based on the `Ueberauth.Strategy.OIDC` strategy.

#### Reviewer Checklist


test/arrow_web/try_api_token_auth/keycloak_test.exs
```diff
@@ -0,0 +1,179 @@
...
+      assert log =~ "{:error, :missing_user}"
+    end
+  end
+
+  defp auth_token_for(email) do
+    token = Arrow.AuthToken.get_or_create_token_for_user(email)
+    Arrow.Repo.get_by(Arrow.AuthToken, token: token)
+  end
+
+  # describe "call/2" do
```
nlwstein, line 111: Should we delete this commented tests or do you think they will be useful in the future?

lib/arrow_web/try_api_token_auth/keycloak.ex
```diff
@@ -0,0 +1,104 @@
...
+
+  def sign_in(conn, auth_token) do
+    with {:ok, user_id} <- lookup_user_id(auth_token.username),
+         {:ok, roles} <- lookup_user_roles(user_id) do
+      conn
+      |> Guardian.Plug.sign_in(
+        ArrowWeb.AuthManager,
+        auth_token.username,
+        %{roles: roles},
+        ttl: {0, :second}
```
nlwstein, line 16: Does this provide a session that never expires?

nlwstein (approved):
Overall, this looks good to me, just a couple non-blocking q's 😄

test/arrow_web/try_api_token_auth/keycloak_test.exs
paulswartz: oop, no; those are a left over when when I duplicated the existing tests.

lib/arrow_web/try_api_token_auth/keycloak.ex
paulswartz: The opposite; it expires immediately so that if it does happen to make it outside of the API (which it shouldn't) it won't be usable.

paulswartz: It should be a session that expires immediately, but I will double-check that's how Guardian treats it (or if there's another way to implement this).

The issue I ran into (and I think this is true of the existing implementation) is that the API returned a session cookie that you can use instead of the API key, and I didn't think that was the normal approach for an API. We disable that now, but also set a 0-second TTL to ensure that if the token did leak out, it wouldn't be usable.

PR #960, opened 2024-01-12
tests: add test for updating a disruption

bklebe:
tests: add test for updating a disruption


meagharty (changes requested):
This integration test passes locally for me but the commit needs a `mix format`

meagharty (approved):

PR #966, opened 2024-02-29
feat: move `logout_url` to the session

paulswartz:
Meets our best practices of not putting the URL in the Guardian claim, as we can run into errors where the cookie gets too large.

s/o to @firestack who pointed this out.


firestack (approved):
Oops, didn't realize I was the reviewer on this for a moment, this looks about like what we did on Skate!

I don't have any other contextual info on if there is any other place in Arrow that needs to be updated or any tests that should be written, but from what I can see and remember from #950 , LGTM.

PR #967, opened 2024-04-10
feat: follow new recommendation for idle/authentication timeouts

paulswartz:
#### Summary of changes
- Idle timeout: 30 minutes
- Maximum session length: 12 hours

This aligns Arrow with the NIST SP 800-63 guidelines.

#### Reviewer Checklist


lib/arrow_web/auth_manager.ex
```diff
@@ -14,4 +22,19 @@ defmodule ArrowWeb.AuthManager do
   end
 
   def resource_from_claims(_), do: {:error, :invalid_claims}
+
+  @impl true
+  def verify_claims(%{"iat" => iat, "auth_time" => auth_time} = claims, _opts) do
```
skyqrose, line 27: Where do the `iat` and `auth_time` claims come from? I couldn't find any reference to them in [`ueberauth_oidcc`](https://gitlab.com/search?search=auth_time&project_id=52390830&search_code=true&repository_ref=main)

lib/arrow_web/controllers/auth_controller.ex
```diff
@@ -66,14 +74,16 @@ defmodule ArrowWeb.AuthController do
...
     |> put_session(:logout_url, logout_url)
     |> Guardian.Plug.sign_in(
       ArrowWeb.AuthManager,
       username,
       %{
+        auth_time: auth_time,
         roles: roles
       },
-      ttl: {expiration - current_time, :seconds}
+      ttl: {1, :minute}
```
skyqrose, line 86: A TTL of one minute? Could you set this token to expire when the iat expires or the auth_time expires, whichever comes first? Or would that prevent revoking access to accounts that are deactivated in EntraID?

skyqrose: Oh, is this so that you reset the 15 minute idle timeout (by having to refresh the token and get a new iat claim) on the next request after the 1m is up?

paulswartz: Exactly; the 1 minute makes sure this token is valid long enough for the next request to be made, at which point the token will be refreshed. It would also work to use the idle timeout here, but then this module also needs to reference the idle timeout value so it seemed cleaner to avoid that.

lib/arrow_web/auth_manager.ex
paulswartz: `auth_time` is coming from EntraID (via `AuthController`), and `iat` is from Guardian.

paulswartz:
@cmaddox5 of course!

cmaddox5:
Actually, this was done in #1028 so we should be able to close this. Any ttl tweaks could be done separate from this branch.

PR #971, opened 2024-05-31
deps: upgrade phoenix to 1.7 with deps

meagharty:
#### Summary of changes
A guide with similar results: https://gist.github.com/chrismccord/00a6ea2a96bc57df0cce526bd20af8a7. This PR doesn't include the optional changes mentioned.

Smoking test this locally and on dev looks good. I tried various arrow pages and they all appear to work. Tests, credo, dialyzer are working too. It's deployed to dev for testing.

```
Upgraded:
  bunt 0.2.0 => 1.0.0 (major)
  castore 1.0.4 => 1.0.7
  cowboy 2.10.0 => 2.12.0
  cowlib 2.12.1 => 2.13.0
  credo 1.6.4 => 1.7.6
  file_system 0.2.10 => 1.0.0 (major)
  phoenix 1.6.16 => 1.7.12
  phoenix_html 3.2.0 => 3.3.4
  phoenix_live_reload 1.3.3 => 1.5.3
  phoenix_live_view 0.16.4 => 0.20.14 (minor)
  phoenix_template 1.0.3 => 1.0.4
  plug 1.15.2 => 1.16.0
  plug_cowboy 2.6.1 => 2.7.1
  plug_crypto 1.2.5 => 2.1.0 (major)
  react_phoenix 1.2.0 => 1.3.1
```

#### Reviewer Checklist
- [ ] Meets ticket's acceptance criteria
- [ ] Any new or changed functions have typespecs
- [ ] Tests were added for any new functionality (don't just rely on Codecov)
- [ ] This branch was deployed to the staging environment and is currently running with no unexpected increase in warnings, and no errors or crashes.

PR #975, opened 2024-06-04
fix: address credo complaint

nlwstein:
#### Summary of changes

Potential resolution to credo complaint in https://github.com/mbta/arrow/pull/971

#### Reviewer Checklist

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -42,7 +42,7 @@ defmodule Arrow.MixProject do
   # Specifies your project dependencies.
   #
   # Type `mix help deps` for examples and options.
-  defp deps do
+  def deps do
     [
       {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
       {:dialyxir, "~> 1.1", only: [:dev], runtime: false},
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -42,7 +42,7 @@ defmodule Arrow.MixProject do
   # Specifies your project dependencies.
   #
   # Type `mix help deps` for examples and options.
-  defp deps do
```
meagharty, line 45: I noticed this too!

This private function is called above on line 13 in `def project`, it looked like it was done that way in other mbta repos too. I'm not sure if it's an outdated convention but I don't feel strongly about it (changing it or not).

meagharty: Should have asked: why did you decide to change it?

meagharty: Oh, if it removes a compilation warning then sounds okay to :shipit:

nlwstein:
> I had a question re: `defp/def deps` but this looks good to me, thank you!

I'll let you merge or copy / paste whatever you want from this, unless you want to treat it like a normal PR. Let me know if you want me to merge etc.

PR #984, opened 2024-06-11
feat: 🏹  Implement Shuttle Route KML Save to S3

nlwstein:
#### Summary of changes
**Asana Ticket:** [🏹  Implement Shuttle Route KML Save to S3](https://app.asana.com/0/584764604969369/1207423464713208/f)

* Uploaded shape files are now stored in S3.
* Uploaded shape files can be downloaded securely from their record in the dashboard. 
* Uploaded shape files are deleted from S3 upon deletion of the record. 
* Uploaded shape files cannot squash existing files. This deviates slightly from the ticket description - I feel that it is simpler to just tell the user that the file already exists, since I would imagine these are only being managed from this UI going forward. If we need the ability to squash as opposed to the user simply renaming the file, I think we should automatically detect the existing record and prompt them to remove that as well to keep a 1:1 relationship between file in s3 and record in arrow. 

#### Reviewer Checklist


config/config.exs
```diff
@@ -34,7 +34,11 @@ config :arrow,
   },
   time_zone: "America/New_York",
   ex_aws_requester: {Fake.ExAws, :admin_group_request},
-  http_client: HTTPoison
+  http_client: HTTPoison,
+  shape_storage_enabled?: false,
+  shape_storage_prefix_env: "arrow-dev-local",
+  shape_storage_bucket: "mbta-gtfs-s3-sandbox",
```
nlwstein, line 40: This should be `meta-arrow`, but I do not have write access at least with my credentials.

config/config.exs
```diff
@@ -34,7 +34,11 @@ config :arrow,
   },
   time_zone: "America/New_York",
   ex_aws_requester: {Fake.ExAws, :admin_group_request},
-  http_client: HTTPoison
+  http_client: HTTPoison,
+  shape_storage_enabled?: false,
```
nlwstein, line 38: Disabled by default, and just stores `disabled` as the value for all of the relevant s3 columns.

config/config.exs
```diff
@@ -34,7 +34,11 @@ config :arrow,
...
+  shape_storage_prefix_env: "arrow-dev-local",
```
nlwstein, line 39: The storage path is like this: `$PREFIX/$PREFIX_ENV/the_file.kml`. In prod, the `prefix_env` is set via the existing AWS env var.

lib/arrow_web/controllers/layout_html/_flash.html.heex
```diff
@@ -2,6 +2,10 @@
   <aside role="alert" class="alert alert-info"><%= message %></aside>
 <% end %>
 
+<%= with message when is_binary(message) <- Phoenix.Flash.get(@flash, :error) do %>
```
nlwstein, line 5: Added this to display a simple error flash - it looks like the existing error flash was designed to show a whole collection of errors (form validation).

lib/arrow_web/controllers/shape_controller.ex
```diff
@@ -46,6 +54,17 @@ defmodule ArrowWeb.ShapeController do
...
   end
 
+  def download(conn, %{"id" => id}) do
+    shape = Shuttle.get_shape!(id)
+
+    conn
+    |> Plug.Conn.resp(:found, "")
+    |> Plug.Conn.put_resp_header(
+      "location",
+      "https://#{shape.bucket}.s3.amazonaws.com/#{shape.path}"
```
nlwstein, line 64: This is why I store the bucket and path - it may be a pre-optimization but should help if we ever have to rename these things vs. string replacing a URL column.

mix.exs
```diff
@@ -51,6 +51,8 @@ defmodule Arrow.MixProject do
       {:ex_aws_rds, "~> 2.0"},
       {:ex_aws_secretsmanager, "~> 2.0"},
       {:ex_aws, "~> 2.1"},
+      {:ex_aws_s3, "~> 2.1"},
+      {:sweet_xml, "~> 0.7.4"},
```
nlwstein, line 55: `sweet_xml` is soft-required by `ex_aws_s3`.

test/arrow/shuttle_test.exs
```diff
@@ -20,11 +20,24 @@ defmodule Arrow.ShuttleTest do
       assert Shuttle.get_shape!(shape.id) == shape
     end
 
-    test "create_shape/1 with valid data creates a shape" do
-      valid_attrs = %{name: "some name"}
+    test "create_shape/1 with valid data creates a shape and uploads to s3" do
```
nlwstein, line 23: Updated our "good-path" test to also actually upload to S3. 

My one concern with this is storage usage over time without a cleanup routine. I think the cleanest implementation would be to update the policy to whack anything with `arrow/test-runner/*` older than n days, vs cleaning it up in the test, because runs that fail may leave detritus on the bucket.

nlwstein:
@meagharty There are a few outstanding issues here that aren't really related to the core feature but may block it's acceptance into the repo until they are complete: 

1. S3 bucket `mbta-arrow`'s permissions don't seem to be right. (Also, my tester S3 bucket `mbta-gtfs-s3-sandbox`'s permissions are fine for writing but it does not have public read and thus the `/shape/{id}/download` route isn't working quite right, but we shouldn't fix this...and just update the correct bucket!) 
2. The tests fail on GitHub because the AWS user has less permissions than I do to interact with S3. 
3. We should also add a policy for truncating the test files uploaded (and possibly the dev env files uploaded) to avoid a ton of useless storage accumulation on s3. 

Otherwise, I think the feature as-is can be reviewed while we deal with these issues 😄 If you disagree I can take it back as a draft while I resolve these things.

test/arrow/shuttle_test.exs
```diff
@@ -20,11 +20,24 @@ defmodule Arrow.ShuttleTest do
...
+      uuid = Ecto.UUID.generate()
+      prefix = "arrow/test-runner/#{uuid}/"
+      Application.put_env(:arrow, :shape_storage_prefix, prefix)
+
+      valid_attrs = %{
+        "name" => "some name",
+        "path" => "some/path/to/sample.kml",
+        "prefix" => prefix,
+        "bucket" => Application.get_env(:arrow, :shape_storage_bucket),
+        "filename" => %Plug.Upload{filename: "sample.kml", path: "test_files/sample.kml"}
```
meagharty, line 35: **not blocking**: we could resolve this in a future PR since the full featureset is expected to extract individual shapes from a KML file (not part of this work)

But: I was talking with @bfauble and we might want this filename to be saved as `#{name}.kml`. To make the names easier to reconcile between gtfs_creator and arrow.

I am pro-keeping the KML format around since it's industry common.

nlwstein: @meagharty This is a great idea and would also mean that we could drop the columns eventually - since the file path / name / etc. would be inherent. 

I guess an alternative would be a "smart" connection between the two systems where it would pull some metadata JSON payload, which might be more extensible long term BUT I think starting with the names makes a lot of sense. 

It would be an easy change here, too!

nlwstein: I also feel like this could be a follow up ticket, just to avoid this getting stuck in review as long as you're cool with that.

lib/arrow/shuttle.ex
```diff
@@ -68,6 +136,22 @@ defmodule Arrow.Shuttle do
 
   """
   def update_shape(%Shape{} = shape, attrs) do
+    if Map.has_key?(attrs, "filename") do
+      case upload_shape(attrs["filename"]) do
+        {:ok, new_attrs} ->
+          # Delete old file:
+          {:ok, _} = delete_shape_file(shape)
```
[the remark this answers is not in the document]
nlwstein: The purpose for this is to whack the file (original name is retained, as-is) so that if it were to be uploaded elsewhere there wouldn't be a collision. So, this isn't technically an "overwrite" scenario 😄

config/prod.exs
```diff
@@ -9,6 +9,11 @@ import Config
 # manifest is generated by the `mix phx.digest` task,
 # which you should run after static files are built and
 # before starting your production server.
+
+config :arrow,
+  shape_storage_enabled?: true,
+  shape_storage_prefix_env: System.get_env("AWS_SECRET_PREFIX") || "arrow-prod-unknown"
```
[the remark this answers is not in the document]
nlwstein: Should be all set, thank you!

lib/arrow/shuttle.ex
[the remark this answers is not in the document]
nlwstein: > oh do you mean it accounts for a name change as well as an overwrite?

Essentially, yes. The filename in s3 matches the uploaded filename. 

We could use a UUID (maybe the primary key of the table) or something instead of the original name, if that would be better? 

I was thinking the names would typically be self-explanatory and it might be useful to keep them (so if someone inspects the s3 bucket, it's obvious what the kmls are for without opening them) and that is why it is implemented this way, with the delete operation ensuring that the file is removed on update when the name changes.

test/arrow/shuttle_test.exs
meagharty: This test doesn't upload to S3 anymore, it uses a mock.

test/arrow/shuttle_test.exs
meagharty: > the full featureset is expected to extract individual shapes from a KML file (not part of this PR)

We name the kml file after the provided (by the user) shape name with more recent changes.

lib/arrow/shuttle.ex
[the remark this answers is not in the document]
meagharty: The current changes (with #991) allow deletion of a shape but it doesn't handle update.

https://github.com/mbta/arrow/pull/984/commits/695d41df795977726ef3009791ac305cec1247bf
```
fix: update code to use processed content instead of reading from a user-inputted file

fix: remove editing of existing shape to avoid feature collision
```

config/config.exs
meagharty: I'm seeing access denied if I switch this to `mbta-arrow` so I'm looking into this.

config/config.exs
meagharty: This was updated in #991

config/config.exs
meagharty: This was addressed in #991

lib/arrow_web/controllers/shape_html/index.html.heex
```diff
@@ -9,11 +9,12 @@
...
+  <:col :let={shape} label="File"><%= shape.path %></:col>
   <:action :let={shape}>
-    <div class="sr-only">
-      <.link navigate={~p"/shapes/#{shape}"}>Show</.link>
-    </div>
-    <.link navigate={~p"/shapes/#{shape}/edit"}>Edit</.link>
+    <.link navigate={~p"/shapes/#{shape}"}>Show</.link>
+  </:action>
+  <:action :let={shape}>
+    <.link navigate={~p"/shapes/#{shape}/download"}>Download</.link>
```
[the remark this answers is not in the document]
meagharty: Yeah I removed "edit" ability when merging in the shape upload and processing due to conflicts. The original mocks don't call for editing post bulk creation flow.

fix: update code to use processed content instead of reading from a file
fix: remove editing of existing shape to avoid feature collision
[695d41d](https://github.com/mbta/arrow/pull/984/commits/695d41df795977726ef3009791ac305cec1247bf)

lib/arrow_web/controllers/shape_html/show.html.heex
```diff
@@ -1,15 +1,11 @@
 <.header>
   Shape <%= @shape.id %>
   <:subtitle>This is a shape record from your database.</:subtitle>
-  <:actions>
```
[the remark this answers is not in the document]
meagharty: This was intentional to not include in this PR. I expected we would re-visit the requirements for edit with the team after getting out the base functionality with S3. That discussion would include support for overwriting a shape file/name if it already exists as well; currently it requires an explicit Arrow deletion to replace a shape.

PR #994, opened 2024-07-18
feat: shuttle stop creation

Whoops:
#### Summary of changes
**Asana Tickets:** 
[🏹 Implement "Create Shuttle Stops" page/form](https://app.asana.com/0/1204473309455397/1207472819136800/f)
[🏹 Implement "Edit Shuttle Stop" page](https://app.asana.com/0/1204473309455397/1207472819136806/f)

#### Reviewer Checklist


priv/repo/structure.sql
```diff
@@ -375,18 +375,18 @@ ALTER SEQUENCE public.shapes_id_seq OWNED BY public.shapes.id;
 
 
 --
--- Name: shuttle_stops; Type: TABLE; Schema: public; Owner: -
+-- Name: stops; Type: TABLE; Schema: public; Owner: -
 --
 
-CREATE TABLE public.shuttle_stops (
+CREATE TABLE public.stops (
```
Whoops, line 381: I didn't actually change the name of the table. It looks like it got changed after the schema was dumped?

Whoops:
image

fsaid90:
@Whoops just a couple of nitpicky comments:

- Would be nice to rename "Stop Latitude" and "Stop Longitude" to "Latitude" and "Longitude"
- A little more spacing between the last textbox and the _Save/Cancel_ buttons so we have equivalent spacing
- Seems like there's a little too much spacing in general between fields + fields/their titles - maybe we can make it a little more compact? Not too much - something like 10-20% decrease. If this matches other pages then you can definitely just ignore this comment 😃 

_Looks great though!!_

meagharty (approved):
I tested this out locally and it looks good to me! 🎊

lib/arrow/stops.ex
```diff
@@ -0,0 +1,104 @@
...
+  ## Examples
+
+      iex> create_stop(%{field: value})
+      {:ok, %Stop{}}
+
+      iex> create_stop(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def create_stop(attrs \\ %{}) do
```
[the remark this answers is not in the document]
meagharty: Ah, this is because we don't call this function without any attrs. Removing the default value `\\ %{}` marks this line as covered.

The generators generate the function signature like this though. We could add a test to test/arrow/stops_test.exs like this since the schema has required fields:
```
    test "create_stop/1 with no data returns error changeset" do
      assert {:error, %Ecto.Changeset{}} = Stops.create_stop()
    end
```

Whoops:
@fsaid90 Here's what the page looks like now. I may have actually gone in the opposite direction, making things bigger, but it's now using the same as the existing Arrow content. Feel free to bully me and tell me I'm wrong :-)

image

lib/arrow_web/controllers/stop_html/stop_form.html.heex
```diff
@@ -1,34 +1,43 @@
...
   <.input field={f[:on_street]} type="text" label="Street" />
   <.input field={f[:at_street]} type="text" label="Cross street" />
   <:actions>
-    <input type="submit" value="Save Shuttle Stop" class="w-60 btn btn-primary" />
-    <a href={~p"/stops"} class="w-60 btn btn-outline-primary" onclick="return confirm('Are you sure you want to cancel? All changes will be lost!')">Cancel</a>
+    <div class="w-25 mr-2">
+      <.button type="submit" class="btn-primary w-100">Save Shuttle Stop</.button>
+    </div>
+    <div class="w-25 mr-2">
+      <.link_button href={~p"/stops"} class="btn-outline-primary w-100" onclick="return confirm('Are you sure you want to cancel? All changes will be lost!')">Cancel</.link_button>
```
meagharty, line 40: Can you use `data-confirm` for this instead?

lib/arrow_web/components/core_components.ex
```diff
@@ -228,20 +227,24 @@ defmodule ArrowWeb.CoreComponents do
...
-    <button
-      type={@type}
-      class={[
-        "phx-submit-loading:opacity-75 rounded-lg bg-zinc-900 hover:bg-zinc-700 py-2 px-3",
-        "text-sm font-semibold leading-6 text-white active:text-white/80",
-        @class
-      ]}
-      {@rest}
-    >
+    <button type={@type} class={["btn", @class]} {@rest}>
```
meagharty, line 230: not blocking: did you consider making this button a `btn-primary` by default?

It seems like we could keep phx-submit-loading:opacity-75 but I don't know if that would work (if it was setup with a liveview implentation) with btn-primary anyway.

meagharty, line 230: These updates to align with the styling on other pages look good to me.

I don't know if we need to be strict about using bootstrap classes or tailwind utility classes in a general sense but in this context I support the changes to match existing components.

I might have tried to re-create the bootstrap styling (i.e. for btn-primary) with tailwind utility classes instead (since those classes are easier to mix and match, or override) but that would be take more time to do and be less exact.

lib/arrow_web/controllers/stop_html/stop_form.html.heex
Whoops: Testing it: Yes. I'm not sure what's providing this functionality, though. Our javascript doesn't seem to read this attribute.

lib/arrow_web/components/core_components.ex
Whoops: I thought about it, but I didn't come up with a solution I love. I don't think I'd want to hardcode a list of "include `btn-primary` unless this class is present" classes, and making the default for `@class` to be `["button-primary"]` unless it's overridden has the (maybe) surprising behavior that if you include any classes (e.x. `w-100`) you'd have to also include `btn-primary`.

I do think this would be a better default if you have a brainstorm on a good way of doing it.

lib/arrow/stops.ex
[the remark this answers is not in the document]
Whoops: I don't see a compelling reason for having default argument that'll just return an error, so I removed it.

lib/arrow_web/components/core_components.ex
Whoops: > I might have tried to re-create the bootstrap styling (i.e. for btn-primary) with tailwind utility classes instead (since those classes are easier to mix and match, or override) but that would take more time to do and be less exact.

Given the discussion in Arrow it sounds like we'll have a polish phase. I think we'll have some decisions to make then around how we want to handle this kind of thing. The annoying thing about the front/backend split we have is that we can't just make a button component that's the same in both. Maybe WebComponents? But it kind of feels like overkill for a simple button. Maybe this is a case for redoing the React functionality in LiveView, or maybe I'm just overthinking it.

lib/arrow_web/controllers/stop_html/stop_form.html.heex
meagharty: It's from liveview's javascript.
https://hexdocs.pm/phoenix_live_view/0.20.14/Phoenix.Component.html#link/1-data-attributes

meagharty (approved):

PR #996, opened 2024-07-25
🏹  Implement Shuttle Route KML Visualization 

bfauble:
#### Summary of changes
**Asana Ticket:** [🏹  Implement Shuttle Route KML Visualization ](https://app.asana.com/0/584764604969369/1207339672568994/f)

image
image
image
image
image

### DIFF
--- a/lib/arrow_web/controllers/stop_html.ex
+++ b/lib/arrow_web/controllers/stop_html.ex
@@ -10,4 +10,10 @@ defmodule ArrowWeb.StopView do
   attr :action, :string, required: true
 
   def stop_form(assigns)
+
+  def format_timestamp(%DateTime{} = dt) do
+    dt
+    |> DateTime.shift_zone!("America/New_York")
+    |> Calendar.strftime("%Y-%m-%d %I:%M %p")
+  end
 end
--- a/lib/arrow_web/controllers/stop_html/index.html.heex
+++ b/lib/arrow_web/controllers/stop_html/index.html.heex
@@ -22,6 +22,7 @@
   <:col :let={stop} label="Municipality"><%= stop.municipality %></:col>
   <:col :let={stop} label="On street"><%= stop.on_street %></:col>
   <:col :let={stop} label="At street"><%= stop.at_street %></:col>
+  <:col :let={stop} label="Last updated"><%= format_timestamp(stop.updated_at) %></:col>
   <:action :let={stop}>
     <div class="sr-only">
       <.link navigate={~p"/stops/#{stop}"}>Show</.link>
--- a/test/arrow_web/controllers/stop_html_test.exs
+++ b/test/arrow_web/controllers/stop_html_test.exs
@@ -1,0 +1,12 @@
+defmodule ArrowWeb.StopViewTest do
+  @moduledoc false
+  use ExUnit.Case
+  alias ArrowWeb.StopView
+
+  describe "format_timestamp" do
+    test "formats a timestamp into Eastern time" do
+      utc = DateTime.from_naive!(~N[2024-07-24T09:30:00], "Etc/UTC")
+      assert StopView.format_timestamp(utc) =~ "2024-07-24 05:30 AM"
+    end
+  end
+end


assets/src/components/ShapeViewMap.tsx
```diff
@@ -1,18 +1,135 @@
...
+  `<div class="legend-square color-${color}"></div> ${name}`
+
+const generatePolyline = (shape: Shape, index: number) => {
+  const colorValue = COLORS[index]
+  const color = `#${colorValue}`
+  const start = shape.coordinates[0]
+  const end = shape.coordinates.slice(-1)[0]
+  const key = crypto.randomUUID()
+
+  return [
```
jzimbel-mbta, line 47: Is there a reason why the returned JSX expression needs to be wrapped in an array?

If not, could you remove the wrapping array and turn this function into a component (capitalize the name, convert it to a noun, and have it take its arguments as a props object) since it returns JSX?

If you need the component to only re-render when its props change, you can [wrap it in `memo`](https://react.dev/reference/react/memo#memo).

assets/src/components/ShapeViewMap.tsx
```diff
@@ -1,18 +1,135 @@
...
+      ,
+    </LayersControl.Overlay>,
+  ]
+}
+
+const getMapBounds = (shapes: Shape[]) => {
+  const shapeLats: number[] = []
+  const shapeLongs: number[] = []
+  shapes.map((shape: Shape) =>
+    shape.coordinates.map((coordinate) => {
```
jzimbel-mbta, line 86: Could you change these `.map`s to `.forEach`? The iteration's return value is unused in both cases.

assets/src/components/ShapeViewMap.tsx
```diff
@@ -1,18 +1,135 @@
...
+  return [
+    [Math.max(...shapeLats), Math.max(...shapeLongs)],
+    [Math.min(...shapeLats), Math.min(...shapeLongs)],
+  ] as LatLngBoundsExpression
+}
+
+const ShapeViewMap = ({ shapes }: ShapeViewMapProps) => {
+  const polyLines = useMemo(() => {
+    if (shapes && shapes.length > 0) {
+      const lines = shapes.map((shape, index) => generatePolyline(shape, index))
```
jzimbel-mbta, line 101: Do you get any console complaints about this? I think `lines` currently has a type like `[ReactNode][]` (An array of 1-element tuples of JSX expressions), so I could see it complaining about a missing `key` prop.

lib/arrow_web/controllers/shape_html.ex
```diff
@@ -1,5 +1,7 @@
 defmodule ArrowWeb.ShapeView do
   use ArrowWeb, :html
+  alias Arrow.Shuttle.ShapesUpload
+  require Logger
```
jzimbel-mbta, line 4: I think this can be removed.

(Looks like the repo is missing a CI step to run credo and/or dialyzer, I forget which tool usually points out unused imports)

lib/arrow_web/controllers/shape_html.ex
```diff
@@ -10,4 +12,30 @@ defmodule ArrowWeb.ShapeView do
...
+
+  defp map_coordinates(coordinates) do
+    Enum.map(coordinates, &process_coordinate_pair/1)
+  end
+
+  defp process_coordinate_pair(coordinate_pair) do
+    coordinate_pair
+    |> String.split(",")
+    |> Enum.map(&String.to_float/1)
+    |> Enum.reverse()
```
jzimbel-mbta, line 39: Why does the pair need to be reversed? Is the `coordinate_pair` string formatted like "long,lat" for some reason?

jzimbel-mbta (commented):
A handful of minor things, otherwise looks good.

lib/arrow_web/controllers/shape_html.ex
jzimbel-mbta: Weird!

jzimbel-mbta: Do you know where the coordinate pair string originates from? Wondering if this is something that's within our ability to fix.

If not, maybe just leave a comment above this line explaining why it needs to be reversed.

assets/src/components/ShapeViewMap.tsx
```diff
@@ -1,18 +1,141 @@
...
 
-const ShapeViewMap = () => {
+const generateNameField = (name: string, color: string) =>
+  `<div class="legend-square color-${color}"></div> ${name}`
+
+const PolyLines = ({ shapes }: { shapes: Shape[] }) =>
+  shapes.map((shape: Shape, index: number) => {
+    const key = crypto.randomUUID()
+    return <PolyLine shape={shape} index={index} key={key} keyPrefix={key} />
+  })
```
jzimbel-mbta, line 44: Just leaving a note here:
The randomly-generated `key`s should be fine in this case since we only expect this component to re-render (in other words, this function to run) when `shapes` changes, and when `shapes` changes it will be a completely different object since it describes a different route pattern.

So it's fine that all of this component's children will remount on re-render—we don't expect there to ever be partial changes to the `shapes` object.

jzimbel-mbta (approved):
Good to go!

(Unless you think there's a way to fix the cause of the "Long,Lat" strings. Then it's good to go after you make that fix.)

PR #1002, opened 2024-08-23
🏹 Fetch and parse s3 file related to shape record to view on map for show pages

bfauble:
#### Summary of changes
**Asana Ticket:** [🏹 Link/Merge Shape and ShapeUpload Structs](https://app.asana.com/0/584764604969369/1208027410019787/f)

Since coordinates are not persisted in the db, fetch the s3 file associated with the shape record to be able to visualize the shape on the map when viewing a specific shape.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN mix local.hex --force && \
 
 RUN curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem \
   -o /root/aws-cert-bundle.pem
-RUN echo "51b107da46717aed974d97464b63f7357b220fe8737969db1492d1cae74b3947 /root/aws-cert-bundle.pem" | sha256sum -c -
+RUN echo "d72191eaa5d48fe2b6e044a0ae0b0e9f35e325b34b1ecab6ffe11d490d5cdb8f /root/aws-cert-bundle.pem" | sha256sum -c -
 
 # Instructions from:
 # https://github.com/nodesource/distributions#debian-versions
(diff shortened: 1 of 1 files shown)


lib/arrow/shuttle.ex
```diff
@@ -41,6 +42,39 @@ defmodule Arrow.Shuttle do
...
+
+      iex> get_shapes_upload(%Shape{})
+      %ShapesUpload{}
+  """
+  def get_shapes_upload(%Shape{} = shape) do
+    with true <- Application.get_env(:arrow, :shape_storage_enabled?),
+         {:ok, %{body: shapes_kml}} <- get_shape_file(shape),
+         {:ok, parsed_shapes} <- ShapesUpload.parse_kml(shapes_kml),
+         {:ok, shapes} <- ShapesUpload.shapes_from_kml(parsed_shapes) do
+      ShapesUpload.changeset(%ShapesUpload{}, %{filename: shape.name, shapes: shapes})
```
meagharty, line 58: Makes sense to re-use the code for parsing the kml.

When I was working on the "Upload" modules, I was referring to the arrow user's initial file upload, not the final upload to s3. But "shape-uploads" was also used in the S3 work as the `shape_storage_prefix`.

`ShapesUpload` isn't necessarily one to one with the s3 entity since we currently upload a s3 file per shape (no multi-shape files, so from S3 `shapes` will always be a list with one shape but from an arrow user, possibly a list with multiple shapes). That said I think this interpretation works!

a nitty request though: `ShapeUpload`'s [moduledoc](https://github.com/mbta/arrow/blob/d28f34ba7de6bc54d57fe9f8acb6c3715cc8f73e/lib/arrow/shuttle/shape_upload.ex#L2) is `schema for selecting shuttle shapes as an embedded schema` where selecting refers to the user flow on `/shapes_upload`. Can we remove "selecting" from the moduledoc?

meagharty (changes requested):

meagharty (approved):

PR #1003, opened 2024-08-23
🏹 Append -S in Arrow shape save

Whoops:
#### Summary of changes
**Asana Ticket:** [🏹 Append -S in Arrow shape save](https://app.asana.com/0/1204473309455397/1208045514739578/f)

#### Reviewer Checklist

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -77,7 +77,7 @@ config :arrow, dev_routes: true
 # Set prefix env for s3 uploads
 config :arrow,
   shape_storage_enabled?: true,
-  shape_storage_prefix_env: "local/"
+  shape_storage_prefix_env: "dev/local/"
 
 # Do not include metadata nor timestamps in development logs
 config :logger, :console, format: "[$level] $message\n"
--- a/lib/arrow/shuttle/shape.ex
+++ b/lib/arrow/shuttle/shape.ex
@@ -23,8 +23,33 @@ defmodule Arrow.Shuttle.Shape do
   @doc false
   def changeset(shape, attrs) do
     shape
-    |> cast(attrs, [:name, :path, :bucket, :prefix])
+    |> cast(attrs, [:name, :path, :bucket, :prefix], empty_values: ["-S" | empty_values()])
     |> validate_required([:name, :path, :bucket, :prefix])
     |> unique_constraint(:name)
+    |> validate_change(:name, fn :name, name ->
+      if String.ends_with?(name, "-S") do
+        []
+      else
+        [{:name, "must end with -S"}]
+      end
+    end)
+  end
+
+  def validate_and_enforce_name(attrs) do
+    changeset =
+      %__MODULE__{}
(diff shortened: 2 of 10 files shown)


config/dev.exs
```diff
@@ -77,7 +77,7 @@ config :arrow, dev_routes: true
 # Set prefix env for s3 uploads
 config :arrow,
   shape_storage_enabled?: true,
-  shape_storage_prefix_env: "local/"
+  shape_storage_prefix_env: "dev/local/"
```
Whoops, line 80: Our users only have access to the `dev` prefix of that bucket

lib/arrow/shuttle/shape.ex
```diff
@@ -23,8 +23,33 @@ defmodule Arrow.Shuttle.Shape do
   @doc false
   def changeset(shape, attrs) do
     shape
-    |> cast(attrs, [:name, :path, :bucket, :prefix])
+    |> cast(attrs, [:name, :path, :bucket, :prefix], empty_values: ["-S"] ++ empty_values())
```
jzimbel-mbta, line 26: Super nit: You can use `[head | tail]` syntax for a simple 1-element prepend.
```suggestion
    |> cast(attrs, [:name, :path, :bucket, :prefix], empty_values: ["-S" | empty_values()])
```

lib/arrow/shuttle/shape.ex
```diff
@@ -23,8 +23,33 @@ defmodule Arrow.Shuttle.Shape do
...
     |> validate_required([:name, :path, :bucket, :prefix])
     |> unique_constraint(:name)
+    |> validate_change(:name, fn :name, name ->
+      if String.ends_with?(name, "-S") do
```
jzimbel-mbta, line 30: Just double-checking: We don't need to also check that there's something preceding the "-S" at this point, because it would have been cast to `nil` on line 26 if there wasn't? (Due to the `"-S"` added to `empty_values`)

lib/arrow/shuttle/shape.ex
```diff
@@ -23,8 +23,33 @@ defmodule Arrow.Shuttle.Shape do
...
+  end
+
+  def validate_and_enforce_name(attrs) do
+    changeset =
+      %__MODULE__{}
+      |> cast(attrs, [:name])
+      |> validate_required([:name])
+
+    cond do
+      !changeset.valid? ->
```
jzimbel-mbta, line 45: nit: Assuming `changeset.valid?` always has a boolean value, prefer using `not`
```suggestion
      not changeset.valid? ->
```

jzimbel-mbta (approved):
Looks good! Just some extremely minor optional suggestions.

lib/arrow/shuttle/shape.ex
Whoops: Yep exactly. We could maybe make the error message better, where the error message if the user enters "-S" for the name will be "can't be blank" but I think that's fine.

PR #1006, opened 2024-09-04
feat: Create tables for GTFS-static archive import

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [Define DB table schemata for GTFS data](https://app.asana.com/0/584764604969369/1208141467238946/f)
(first subtask of [[extra] 🏹 Implement /import-GTFS Arrow endpoint](https://app.asana.com/0/584764604969369/1207472819136817/f))

This is split into several migrations, partly to keep file lengths reasonable and partly to make it clear which tables reference/depend on the existence of other tables.

<details>
<summary><h4>Output of <code>mix ecto.migrate --log-migrations-sql</code></h4></summary>

```sql
11:27:58.114 [info] == Running 20240826153959 Arrow.Repo.Migrations.CreateGtfsTablesPart1.change/0 forward

11:27:58.117 [info] create table gtfs_agencies

11:27:58.137 [debug] QUERY OK db=18.8ms
CREATE TABLE "gtfs_agencies" ("id" integer, "name" varchar(255) NOT NULL, "url" varchar(255) NOT NULL, "timezone" varchar(255) NOT NULL, "lang" varchar(255), "phone" varchar(255), PRIMARY KEY ("id")) []

11:27:58.137 [info] create table gtfs_checkpoints

11:27:58.140 [debug] QUERY OK db=3.3ms
CREATE TABLE "gtfs_checkpoints" ("id" varchar(255), "name" varchar(255) NOT NULL, PRIMARY KEY ("id")) []

11:27:58.140 [info] create table gtfs_levels

11:27:58.142 [debug] QUERY OK db=2.1ms
CREATE TABLE "gtfs_levels" ("id" varchar(255), "index" float NOT NULL, "name" varchar(255), PRIMARY KEY ("id")) []

11:27:58.142 [info] create table gtfs_lines

11:27:58.146 [debug] QUERY OK db=3.5ms
CREATE TABLE "gtfs_lines" ("id" varchar(255), "short_name" varchar(255) NOT NULL, "long_name" varchar(255) NOT NULL, "desc" varchar(255) NOT NULL, "url" varchar(255), "color" varchar(255) NOT NULL, "text_color" varchar(255) NOT NULL, "sort_order" integer NOT NULL, PRIMARY KEY ("id")) []

11:27:58.147 [info] == Migrated 20240826153959 in 0.0s

11:27:58.167 [info] == Running 20240826154124 Arrow.Repo.Migrations.CreateGtfsTablesPart2.change/0 forward

[...]

PR #1013, opened 2024-09-19
feat: GTFS feed import

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [ [extra] 🏹 Implement /import-GTFS Arrow endpoint](https://app.asana.com/0/584764604969369/1207472819136817/f)

This adds the rest of the code for GTFS import, started in #1006.

I had intended to split the remaining work into more than 1 PR, but hopefully it will be easy enough to review it commit-by-commit. Happy to break it apart if that would be preferable to the reviewer.

## Test-running the import
With Arrow running locally and a GTFS archive saved locally, you can run the import with the following:
```sh
# Change to 'validate' to hit the validate endpoints instead
ACTION=import
ARCHIVE_PATH="$HOME/Downloads/MBTA_GTFS.zip"

# Enqueue the import job
job_id=$(
  curl -X POST localhost:4000/api/gtfs/$ACTION \
    --header "x-api-key: $ARROW_API_KEY" \
    --header "Content-Type: application/zip" \
    --data-binary "@$ARCHIVE_PATH" \
  | jq .id
)

# Check status of the job until it's something other than "executing"
while true; do
  result=$(
    curl -s -X GET localhost:4000/api/gtfs/$ACTION/${job_id}/status \
      --header "x-api-key: $ARROW_API_KEY" \
    | jq .status
  )

  echo $result
  if [[ $result != "\"executing\"" ]]; then break; else sleep 1; fi
done
```
(It assumes the import job is successfully enqueued in the first request, so you'll need to adjust things a bit if you're testing import of an invalid archive)

#### Reviewer Checklist


Whoops:
Drive by commentary based on a skimming of the implementation:

- On directions, how would you feel about a quick GTFS PR to clean up directions that don't have routes in the feed? I'm pretty confident you'd just have to remove them [here](https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/truncate/main.py#L119), right after we remove routes that aren't scheduled.
- Ecto has `insert_all/3` that I believe will outperform doing multiple `insert`s, since it can load everything up, then check the constraints and update indexes instead of having to do it multiple times for individual insters. It takes maps instead of changesets, but I think we're relying on the database for validations, so I think that's a worthwhile tradeoff.
- If we really wanted to golf this, there are other things we could do like dropping constraints and re-adding them
 after import (SQL Server actually has a way of disabling constraints during a transaction and rechecking them at the end, but sadly Postgres does not), but I think increasing the timeout is fine. This'll only run on deploys and we already wait ~15 minutes for Alerts UI to validate. We can always try to bring it down later if it becomes annoying.

jzimbel-mbta:
@Whoops Re: directions.txt rows referencing nonexistent routes,

Based on what I see [here](https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/truncate/main.py#L133-L137), it seems like at least for shuttle routes we make an exception and keep rows in directions.txt that reference nonexistent `Shuttle-*` routes. Is there a reason for that?

(I'm not sure why directions referencing bus route 602 doesn't get removed by this line though, it definitely should be)

Whoops:
> Based on what I see [here](https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/truncate/main.py#L133-L137), it seems like at least for shuttle routes we make an exception and keep rows in directions.txt that reference nonexistent `Shuttle-*` routes. Is there a reason for that?

Unless I'm misreading something that doesn't protect `Shuttle-*` just `Shuttle-Generic` (it does the same for routes just above). I'm not sure why, but since the route is kept, it shouldn't be causing this issue. On the flip side, we're already truncating out these shapes I'm not sure why they are ending up in the feed. I'm between tickets at the moment, I'll probably drop in a ticket and look into this today.

lib/arrow_web/controllers/api/gtfs_import_controller.ex
```diff
@@ -0,0 +1,30 @@
...
+  use ArrowWeb, :controller
+  import Ecto.Query
+
+  # TODO: Authorization
+
+  def import(conn, %{"archive" => %Plug.Upload{} = upload}) do
+    version_query =
+      from info in Arrow.Gtfs.FeedInfo, where: info.id == "mbta-ma-us", select: info.version
+
+    version = Arrow.Repo.one!(version_query)
```
meagharty, line 11: Note from testing this out:

I changed this to be 
```suggestion
    version = Arrow.Repo.one(version_query)
```

so the initial import can occur (before there's anything in the feed info table).

jzimbel-mbta: Oop thanks! I had noticed and fixed this yesterday but just got the commit out for it.

lib/arrow/gtfs.ex
```diff
@@ -0,0 +1,151 @@
...
+          Repo.rollback(:dry_run_success)
+        end
+      end,
+      timeout: @import_timeout_ms
+    )
+  end
+
+  defp truncate_all do
+    tables = Enum.map_join(importable_schemas(), ", ", & &1.__schema__(:source))
+    Repo.query!("TRUNCATE #{tables}")
```
Whoops, line 82: TIL you can truncate multiple tables in a single statement

lib/arrow/gtfs/import_helper.ex
```diff
@@ -0,0 +1,115 @@
...
+      iex> values = List.duplicate(row, 6_000)
+      iex> chunked = chunk_values(values)
+      iex> Enum.count(chunked)
+      3
+      iex> length(Enum.at(chunked, 0))
+      2520
+  """
+  @spec chunk_values(Enumerable.t(map)) :: list(list(map))
+  def chunk_values(values) do
+    if Enum.empty?(values) do
```
Whoops, line 98: This is cool. I probably would have just picked an arbitrary number sufficiently small to not cause issues :-)

lib/arrow/gtfs/route.ex
```diff
@@ -0,0 +1,65 @@
...
+  Represents a row from routes.txt.
+
+  Changeset is intended for use only in CSV imports--
+  table contents should be considered read-only otherwise.
+  """
+  use Arrow.Gtfs.Schema
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{
+          id: String.t(),
```
Whoops, line 12: Just sharing, not recommending anything be rewritten: https://hexdocs.pm/typed_ecto_schema/TypedEctoSchema.html

lib/arrow/gtfs/types/enum.ex
```diff
@@ -0,0 +1,195 @@
+# credo:disable-for-this-file
+defmodule Arrow.Gtfs.Types.Enum do
```
Whoops, line 2: Can you clarify for me what this is doing for us vs using Ecto.Enum directly?

lib/arrow/gtfs/stop_time.ex
```diff
@@ -0,0 +1,76 @@
...
+  Changeset is intended for use only in CSV imports--
+  table contents should be considered read-only otherwise.
+  """
+  use Arrow.Gtfs.Schema
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{
+          trip: Arrow.Gtfs.Trip.t() | Ecto.Association.NotLoaded.t(),
+          stop_sequence: integer,
+          arrival_time: Time.t(),
```
Whoops, line 14: Shouldn't this be a String now that depature and arrival are being imported as is? I think that would also allow dropping the entire `Arrow.Gtfs.Types.Time` module.

lib/arrow_web/controllers/api/gtfs_import_controller.ex
```diff
@@ -0,0 +1,30 @@
+defmodule ArrowWeb.API.GtfsImportController do
+  use ArrowWeb, :controller
+  import Ecto.Query
+
+  # TODO: Authorization
```
Whoops, line 5: I think authorization is being covered in the router (requires an admin API key)

lib/arrow_web/controllers/api/gtfs_import_controller.ex
```diff
@@ -0,0 +1,30 @@
...
+        json(conn, %{old_version: version, new_version: new_version})
+
+      :unchanged ->
+        conn
+        |> put_status(412)
+        |> text("Feed version \"#{version}\" already imported")
+
+      :error ->
+        conn
+        |> put_status(500)
```
Whoops, line 26: Nit: this is a bit cheaty, but I'd be inclined to return a 400, the idea being most of the time it'll fail due to an issue with the GTFS feed (bad upload rather than a server-side error). An obvious exception would errors like Arrow being unable to connect to the database, but I don't think we need to put a ton of effort into covering those. Devs will be using this endpoint and will hopefully be able to tell the difference.

Also, I think getting the database error in the message somewhere would be beneficial so GTFS creator can report on why it failed, not just that it did.

priv/repo/migrations/20240826153959_create_gtfs_tables_part1.exs
```diff
@@ -8,8 +8,99 @@ defmodule Arrow.Repo.Migrations.CreateGtfsTablesPart1 do
...
+    ###########################################
+    # Tables enumerating integer-coded fields #
+    ###########################################
+    create_and_populate_enum_table("gtfs_service_exception_types", %{
+      1 => "Added",
+      2 => "Removed"
+    })
+
+    create_and_populate_enum_table("gtfs_route_types", [
+      "Light Rail",
```
Whoops, line 24: I wonder if this is overkill. Ecto.Enum would already not let us provide a value outside this set. If we wanted to enforce it on a database level, too, a check constraint would let us limit these to pre-defined values. I also think the usage of Ecto.Enum in the schemas mean we never actually join on these tables, despite populating them and using them for FKs.

Whoops (commented):
Very non-blocking commentary: if we went all-in on CSV import, I think that would allow for the removal of a lot of code. But of course that conflicts with what we just discussed around services.

lib/arrow/gtfs.ex
```diff
@@ -0,0 +1,151 @@
...
+  end
+
+  defp import_transaction(unzip, dry_run?) do
+    Repo.transaction(
+      fn ->
+        truncate_all()
+        import_all(unzip)
+
+        if dry_run? do
+          Repo.query!("SET CONSTRAINTS ALL IMMEDIATE")
```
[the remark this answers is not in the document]
jzimbel-mbta: Yes—there's still one deferred constraint that we can't un-defer since it's a circular reference: `gtfs_route_patterns.representative_trip_id`.
`gtfs_trips.route_pattern_id` makes it a circular reference.

The deferred FK is set up [here](https://github.com/mbta/arrow/blob/jz-import-gtfs-endpoint--ecto-schemas/priv/repo/migrations/20240826154213_create_gtfs_tables_part5.exs#L30).

lib/arrow/gtfs.ex
jzimbel-mbta: So glad it's supported! Otherwise blowing away all of the old data would have been much more of an ordeal.

(Postgres is even nice enough to yell at you if you try to truncate table B referenced by table A, without _also_ truncating table A in the same statement!)

lib/arrow/gtfs/types/enum.ex
jzimbel-mbta: Yeah sorry, this was a bit of an exasperated moment for me.

The default `Ecto.Enum` is set up as follows:
- load: field is expected to be a string or integer, gets converted to atom
- dump: field is expected to be an atom, gets converted to string or integer
- **cast: :warning: field is expected to be a string :warning:, gets converted to atom**

The problem for our use case is in that last part. For us, the field's value comes in from CSV as a string-formatted integer, but `Ecto.Enum` only accepts strings.

My change makes it so that a string-encoded integer is properly cast to its corresponding atom value.

For example, in `Arrow.Gtfs.CalendarDate`:

```ex
field :exception_type, Arrow.Gtfs.Types.Enum, values: [added: 1, removed: 2]
```

In calendar_dates.txt we see something like:
```csv
...,exception_type,...
...,1,...
```

With `Ecto.Enum`, we'd fail to convert `%{"exception_type" => "1"}` during casting.
With the custom enum type, it converts properly to `%{exception_type: :added}`.

---

<details>
<summary><h3>Diff between original <code>Ecto.Enum</code> and my tweaked version</h3></summary>

```diff
-    on_cast = Map.new(mappings, fn {key, _} -> {Atom.to_string(key), key} end)
+    on_cast =
+      case type do
+        :integer ->
+          mappings
+          |> Enum.flat_map(fn {key, val} ->
+            [{Atom.to_string(key), key}, {Integer.to_string(val), key}]
+          end)
+          |> Map.new()
+
+        _ ->
+          Map.new(mappings, fn {key, _} -> {Atom.to_string(key), key} end)
+      end
```

</details>

I'd love to find a way to adjust `Ecto.Enum`'s behavior around this without copy-pasting the whole thing, but I couldn't find one. 🤷‍♂️

lib/arrow_web/controllers/api/gtfs_import_controller.ex
jzimbel-mbta: Yeah agreed, 500 isn't the right response type here.

I had been meaning to look into other status codes, as well as how to more clearly distinguish between "input feed was bad" and "Arrow broke".

I'll make some changes, thanks for raising.

lib/arrow/gtfs/types/enum.ex
Whoops: Can we go the other direction and do it when cast it as it comes out of the CSV? With the exceptions of the tables being loaded via COPY there's a fair amount of preprocessing that happens already, and the COPY tables should get cast to integers by the database.

My concern is that this is a lot of code to import that'll very easily drift out of date as we update Ecto.

Another option might be to just accept them as Strings in the database. With the allowable values being controlled by `Ecto.Enum`, I don't think that'd open us up to data corruption, and we don't need to do math on these values, in fact `Enum` is going to hide the underlying values from us by converting them to atoms.

jzimbel-mbta: I was wrong about this—`Ecto.Enum`'s `cast` will properly convert integers to atoms, so all I need to do is convert the enum fields from numeric strings to integers before they get passed through `Ecto.Changeset.cast`.

I thought I'd checked this before making my complicated solution, but I guess not 🤦‍♂️

priv/repo/migrations/20240826153959_create_gtfs_tables_part1.exs
jzimbel-mbta: Yeah, check constraint should work just as well without creating unnecessary tables. I'll make that change.

config/config.exs
```diff
@@ -38,7 +38,11 @@ config :arrow,
   shape_storage_enabled?: false,
   shape_storage_bucket: "mbta-arrow",
   shape_storage_prefix: "shape-uploads/",
-  shape_storage_request_fn: {ExAws, :request}
+  shape_storage_request_fn: {ExAws, :request},
+  gtfs_archive_storage_enabled?: false,
+  gtfs_archive_storage_bucket: "mbta-arrow",
+  gtfs_archive_storage_prefix: "gtfs-archive-uploads/",
```
Whoops, line 44: Why is the archive getting loaded into S3?

Whoops: I think I answered my own question. Instead of starting the job directly, the job is inserted into the database and picked up later, so we can't just hand it the iodata that's already in memory, and need a temporary place to store the zip file until Oban picks it up, is that correct?

Whoops (approved):
image

PR #1024, opened 2024-10-24
🏹 Upgrade Elixir version

Whoops:
#### Summary of changes
**Asana Ticket:** [🏹 Upgrade Elixir version](https://app.asana.com/0/1204137169527258/1208367835950806/f)

#### Reviewer Checklist


config/runtime.exs
```diff
@@ -56,8 +56,7 @@ if config_env() == :prod do
       root_source_code_path: File.cwd!(),
       tags: %{
         env: sentry_env
-      },
-      included_environments: [sentry_env]
```
Whoops, line 60: https://hexdocs.pm/sentry/upgrade-10-x.html#stop-using-included_environments

mix.exs
```diff
@@ -60,11 +60,11 @@ defmodule Arrow.MixProject do
       {:gettext, "~> 0.11"},
       {:guardian, "~> 2.0"},
       {:hackney, "~> 1.9"},
-      {:httpoison, "~> 1.6"},
-      {:ja_serializer, github: "mbta/ja_serializer", branch: "master"},
+      {:httpoison, "~> 2.2"},
+      {:ja_serializer, "~> 0.18.0"},
```
Whoops, line 64: Our version of ja_serializer has diverged from upstream, but also not been touched in 3 years. Best I could tell Arrow does not appear to rely on any MBTA specific behaviors.

lib/arrow_web/controllers/shape_controller.ex
```diff
@@ -51,11 +51,6 @@ defmodule ArrowWeb.ShapeController do
         conn
         |> put_flash(:errors, reason)
         |> render(:new_bulk, errors: reason, shapes_upload: reset_upload)
-
-      error ->
```
Whoops, line 55: Dialyzer flags this clause as unreachable, and I agree. `parse_kml_from_file` and `shapes_from_kml` will always return `{:ok, _}` or `{:error, _}`, and `changeset` will always return a changeset, so will always match.

mix.exs
```diff
@@ -14,7 +14,7 @@ defmodule Arrow.MixProject do
       releases: releases(),
       dialyzer: [
         plt_add_apps: [:mix],
-        plt_add_deps: :transitive,
+        plt_add_deps: :app_tree,
```
Whoops, line 17: https://hexdocs.pm/dialyxir/changelog.html#changed-3

mix.exs
```diff
@@ -14,11 +14,10 @@ defmodule Arrow.MixProject do
...
         flags: [
           :unmatched_returns
-        ],
-        ignore_warnings: ".dialyzer.ignore-warnings"
```
Whoops, line 21: Dialyzer complained about `could not read file stats ".dialyzer_ignore.exs": no such file or directory` as long as this was present. I didn't figure out why, but we also aren't ignoring warnings (.dialyzer.ignore-warnings is blank) so I removed it.

Dockerfile
```diff
@@ -39,6 +39,7 @@ COPY config/config.exs config/
 COPY config/prod.exs config/
 
 RUN mix deps.compile
+RUN mix sentry.package_source_code
```
Whoops, line 42: https://hexdocs.pm/sentry/upgrade-10-x.html#what-do-i-have-to-do

test/arrow_web/controllers/stop_controller_test.exs
```diff
@@ -40,7 +40,7 @@ defmodule ArrowWeb.StopControllerTest do
       conn = post(conn, ~p"/stops", stop: invalid_attrs)
       assert redirected_to(conn) == ~p"/stops/new"
 
-      assert get_flash(conn, :errors) ==
+      assert conn.assigns.flash["errors"] ==
```
meagharty, line 43:
```suggestion
      assert Phoenix.Flash.get(conn.assigns.flash, :errors) ==
```

mix.exs
meagharty: I'm in support of this change since our version hasn't been maintained.

Was this upgrade of `ja_serializer` required as part of upgrading the elixir version though?

Can we split this change into it's own separate PR? If we notice any unexpected changes (like with the gtfs_creator build when parsing API results, performance issues with endpoints, whatever), it'd be easier to isolate that way.

meagharty (changes requested):

mix.exs
Whoops: Looks like it can be. Some of these depredations were forced since we compile with `--warnings-as-errors` in the test phase, but while keeping `ja_serializer` generates warnings during tests, it seems to compile happily. I've split it into #1026

lib/arrow/disruption_revision.ex
```diff
@@ -285,4 +285,12 @@ defmodule Arrow.DisruptionRevision do
         changeset
     end
   end
+
+  defp date_range(start_date, end_date) do
```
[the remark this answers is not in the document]
meagharty: I think the code below is correct but what you meant in this comment was `start_date > end_date` for step of -1 (so it enumerates backwards in time from start_date to end_date)

meagharty (approved):
🎉

PR #1030, opened 2024-10-31
tweak: Drop external foreign key constraints during GTFS import

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** No ticket, followup to GTFS import feature

In order for `TRUNCATE all_the_gtfs_tables` to be possible, the tables need to have no dependent foreign key constraints originating from outside the set of tables being truncated.

This achieves that by temporarily dropping such constraints, and then re-adding them after the GTFS tables have been repopulated.

### Implementation notes
I wanted to make sure any Arrow -> GTFS foreign key constraints added in the future would be handled automatically by this logic.

I felt it would be too cumbersome for developers to have to remember to update code in two different places only when adding foreign keys satisfying these specific conditions.

As a result the code relies on some introspection tools provided by Postgres and is a bit heavy on custom queries, but I think it's still relatively straightforward, and worth it for the ease of use it provides.

#### Reviewer Checklist


lib/arrow/foreign_key_constraint.ex
jzimbel-mbta, line 1: Would this be better located within `repo/`, as `Arrow.Repo.ForeignKeyConstraint`?
It feels a little weird to have it right alongside other more normal schema modules.

Whoops: I think that would be fine. I don't necessarily think there's anything wrong with where it is, but it also makes sense as a sort of "utility module" specific to the repo.

Whoops (changes requested):
I don't have a problem with the implementation as written, but how hard would it be to write tests for this functionality? If it's not to much effort, I think it'd be worthwhile to have.

jzimbel-mbta:
> I don't have a problem with the implementation as written, but how hard would it be to write tests for this functionality? If it's not to much effort, I think it'd be worthwhile to have.

@Whoops Sounds good, would testing the functions in `ForeignKeyConstraint` be enough?
Or would you also like to see tests on `Arrow.Gtfs`?

The latter is a bit more challenging in its current state, since it's hardcoded to act on the GTFS tables specifically, but I should be able to tweak it to be more testable.

Whoops:
I think `ForeignKeyConstraint` is sufficient. If you did try to cover `GTFS` I'd suggest doing integration tests that import a feed and check on the state of the world afterward, but that may be more work than it's worth.

Whoops (approved):

PR #1031, opened 2024-11-01
fix: edit stop broken after dependency update

Whoops:
#### Summary of changes
**Asana Ticket:** [🐛🏹 Fix Stop ID Validation in Stop Form](https://app.asana.com/0/1204137169527258/1208671244010889/f)

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/live/stop_live/stop_live.ex
+++ b/lib/arrow_web/live/stop_live/stop_live.ex
@@ -103,14 +103,13 @@ defmodule ArrowWeb.StopViewLive do
   def mount(_params, session, socket) do
     logout_url = session["logout_url"]
     form = to_form(Stops.change_stop(%Stop{}))
-    stops = Stops.list_stops()
 
     socket =
       socket
       |> assign(:form, form)
       |> assign(:form_action, "create")
       |> assign(:http_action, ~p"/stops")
-      |> assign(:stops, stops)
+      |> assign(:stop, %Stop{})
       |> assign(:title, "create shuttle stop")
       |> assign(:stop_map_props, %{})
       |> assign(:logout_url, logout_url)
@@ -130,7 +129,7 @@ defmodule ArrowWeb.StopViewLive do
   end
 
   def handle_event("validate", %{"stop" => stop_params}, socket) do
-    form = Stops.change_stop(%Stop{}, stop_params) |> to_form(action: :validate)
+    form = Stops.change_stop(socket.assigns.stop, stop_params) |> to_form(action: :validate)
 
     {:noreply,
      socket
--- a/lib/arrow_web/router.ex
+++ b/lib/arrow_web/router.ex
@@ -54,7 +54,7 @@ defmodule ArrowWeb.Router do
     live("/stops/new", StopViewLive, :new)
     live("/stops/:id/edit", StopViewLive, :edit)
     get("/stops", StopController, :index)
-    post("/stops/:id", StopController, :update)
+    put("/stops/:id", StopController, :update)
     post("/stops", StopController, :create)
     resources("/shapes", ShapeController, only: [:delete, :index, :show])
     get("/shapes_upload", ShapeController, :new)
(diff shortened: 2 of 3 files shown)


lib/arrow_web/live/stop_live/stop_live.ex
```diff
@@ -130,7 +129,7 @@ defmodule ArrowWeb.StopViewLive do
   end
 
   def handle_event("validate", %{"stop" => stop_params}, socket) do
-    form = Stops.change_stop(%Stop{}, stop_params) |> to_form(action: :validate)
+    form = Stops.change_stop(socket.assigns.stop, stop_params) |> to_form(action: :validate)
```
Whoops, line 132: The initial bug happened because [we use unsafe_validate_unique](https://github.com/mbta/arrow/blob/master/lib/arrow/shuttles/stop.ex#L67) to ensure we don't duplicate stop_ids and with us always comparing to an empty `%Stop{}`, it always thought we changed the `stop_id` so ran the validation. Building our changeset against the original stop prevents that.

I'm unsure why the original code was OK with it.

lib/arrow_web/live/stop_live/stop_live.ex
```diff
@@ -103,14 +103,13 @@ defmodule ArrowWeb.StopViewLive do
   def mount(_params, session, socket) do
     logout_url = session["logout_url"]
     form = to_form(Stops.change_stop(%Stop{}))
-    stops = Stops.list_stops()
```
Whoops, line 106: `stops` was never used, so I removed it. It is unrelated to the issue at hand, so I can put it back if we'd rather be conservative.

lib/arrow_web/router.ex
```diff
@@ -54,7 +54,7 @@ defmodule ArrowWeb.Router do
     live("/stops/new", StopViewLive, :new)
     live("/stops/:id/edit", StopViewLive, :edit)
     get("/stops", StopController, :index)
-    post("/stops/:id", StopController, :update)
+    put("/stops/:id", StopController, :update)
```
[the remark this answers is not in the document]
Whoops: I've tracked this down to the [call to to_form](https://github.com/mbta/arrow/blob/master/lib/arrow_web/live/stop_live/stop_live.ex#L87) generating a form structure with the method `options: [method: "put"],` when being called on a changeset based on an existing stop, and `options: [method: "post"]` when the changeset is based on an empty Stop, but best I can tell this is undocumented, and doesn't appear in the changelog.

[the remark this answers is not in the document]
Whoops: OK, I tracked this down. The [behavior comes from phoenix_html](https://github.com/phoenixframework/phoenix_ecto/blame/main/lib/phoenix_ecto/html.ex#L338), and it's been this way for a long time. The reason it was POST in the version of the code before is because while the form would initially be generated from the stop with and therefore have the `method` set to "put", the validate event when it fired was replacing the form with one based on a new stop (`form = Stops.change_stop(%Stop{}, stop_params) |> to_form(action: :validate)`), and so got replaced with a form with the `method: "post"`. This new version continues to use the existing stop, so the form remains "put"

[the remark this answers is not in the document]
lemald: Aha, that's another weird issue I ran into with the form that I had been wondering about!

lib/arrow_web/live/stop_live/stop_live.ex
lemald: I think it's clear enough to me that it's unused, plus pulling in all of the stops probably adds a smidgen of slowness to the initial component mount, so I'm fine with removing it.

lemald (approved):
LGTM - thanks for figuring this out, even if it's still a little mysterious exactly why it worked initially.

test/arrow_web/live/stop_live/stop_live_test.exs
```diff
@@ -122,7 +122,7 @@ defmodule ArrowWeb.StopLiveTest do
       conn = follow_trigger_action(form, conn)
       assert conn.method == "POST"
```
meagharty, line 123: I wonder why this conn.method is still `POST`? 🤔

The test fails if you change it

PR #1032, opened 2024-11-04
chore: Fix structure.sql to align with what current migrations produce

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** no ticket

A previous PR I posted had an erroneous structure.sql—I'm not sure how it happened, I must have run migrations on an outdated commit at some point.

This fixes the file so that it properly reflects what the current set of migrations produces.

#### Reviewer Checklist

### DIFF
--- a/priv/repo/structure.sql
+++ b/priv/repo/structure.sql
@@ -32,6 +32,45 @@ CREATE TYPE public.day_name AS ENUM (
 
 
 --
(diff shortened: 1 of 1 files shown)


lemald (approved):
Confirmed that if I check this branch out locally and run `mix ecto.reset` I get the same result, minus an inconsequential diff because I'm using Postgres.app instead of having it installed via Homebrew.

PR #1043, opened 2024-11-26
feat: add 

meagharty:
#### Summary of changes
**Asana Ticket:** [Implement "Create/Edit Shuttle Route Definition" - Time to Next Stop](https://app.asana.com/0/584764604969369/1208355803605134/f)

* Adds feature to Shuttle page to retrieve travel time estimates between shuttle stops
* Add OpenRouteServiceAPI module, a slightly modified version of [skate's](https://github.com/mbta/skate/blob/70c3a07da1f5c3499c48767cdab1a6ce3135b4c0/lib/skate/open_route_service_api.ex)

#### Reviewer Checklist


test/support/fixtures/shuttles_fixtures.ex
```diff
@@ -67,9 +69,37 @@ defmodule Arrow.ShuttlesFixtures do
   def unique_shuttle_route_destination,
     do: "some shuttle_route_destination#{System.unique_integer([:positive])}"
 
-  defp shuttle_routes do
+  defp shuttle_route_stops do
+    [stop1, stop2, stop3, stop4] = insert_list(4, :gtfs_stop)
```
meagharty, line 73: Noting this because it's not obvious from this PR: phoenix generators with Ecto produce this file with fixtures by default -- I like these as they call the Ecto.Changeset functions (through `create_*`, in this case on line 139). I believe that's the main difference between these and the manually created ExMachina test data in `Arrow.Factory`.

We are using ExMachina with options for Ecto where it'll do inserts for you. We have one set up for `Arrow.Shuttles.Stop` and one for `Arrow.Gtfs.Stop` [in Arrow.Factory](https://github.com/mbta/arrow/blob/36a52ed5dd4d0abdf50aa1c5020bee3d854855f0/test/support/factory.ex#L60-L93)

lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -358,6 +428,51 @@ defmodule ArrowWeb.ShuttleViewLive do
     }
   end
 
+  @spec get_stop_travel_times(list({:ok, any()})) ::
+          {:ok, list(number())} | {:error, any()}
+  defp get_stop_travel_times(stop_coordinates) do
+    stop_coordinates
```
rudiejd, line 434: what happens if one of the stops in the list that's passed is `{:error, _}` ? should we have a default function clause with an error?

README.md
```diff
@@ -13,6 +13,9 @@
 - PostgreSQL 15
   - Can be run as a container via docker-compose by running `docker-compose up -d` in this repository (see `docker-compose.yml`)
   - Using Homebrew: `brew install postgresql@15`
+- OpenRouteService
+  - Can use https://api.openrouteservice.org/ with an API key requested through their console
```
meagharty, line 17: I hit some issues locally running ORS in docker-compose (it works but after a while it has issues). I expect I need to fine-tune my colima setup for docker-compose.

lib/arrow/open_route_service_api/directions_request.ex
```diff
@@ -0,0 +1,50 @@
...
+  @derive Jason.Encoder
+
+  defmodule Options do
+    @moduledoc false
+    defmodule ProfileParams do
+      @moduledoc false
+      defmodule HgvRestrictions do
+        @moduledoc false
+        @type t :: %{length: float(), width: float(), height: float()}
+        defstruct [:length, :width, :height]
```
jzimbel-mbta, line 14: Does this need a `defstruct` if the struct is not used? (`bus_40ft/0` returns a bare map, not a struct, and `t()` specifies a bare map as well.)

lib/arrow/open_route_service_api/directions_request.ex
```diff
@@ -0,0 +1,50 @@
...
+        def bus_40ft,
+          do: %{
+            length: 12.192,
+            width: 3.2004,
+            height: 3.5052
+          }
+      end
+
+      @type t :: %{restrictions: HgvRestrictions.t()}
+      defstruct [:restrictions]
```
jzimbel-mbta, line 25: Same thing here, a struct is defined but not used.

lib/arrow/open_route_service_api/directions_request.ex
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    @type t :: %{profile_params: ProfileParams.t(), vehicle_type: String.t()}
+    defstruct [:profile_params, :vehicle_type]
```
jzimbel-mbta, line 29: Same thing here, a struct is defined but not used.

jzimbel-mbta:
Oops, I switched to the "GitHub Pull Requests" VSCode extension to continue my review and was accidentally hitting "add comment" instead of "start review". Didn't mean to submit each comment as a separate review! The rest will come in as a single review.

lib/arrow/open_route_service_api/directions_request.ex
meagharty: I changed bus_40ft to return the struct 👍 

I don't know if the original from skate used the struct somewhere else but maybe it was just overlooked.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -121,6 +122,7 @@ defmodule ArrowWeb.ShuttleViewLive do
   end
 
   attr :f, :any, required: true
+  attr :errors, :map, required: false, default: %{route_stops: %{}}
```
jzimbel-mbta, line 125: I think since `shuttle_form_stops_section` is rendered only within `shuttle_form`, the `errors` attr can be required at this point. In the same way that you can have a public function with a default argument, but private functions it passes that argument on to expect it to always be specified.

test/support/factories/open_route_service_factory.ex
```diff
@@ -0,0 +1,69 @@
+defmodule Arrow.OpenRouteServiceFactory do
+  @moduledoc """
+  Defines ExMachina factory functions for `Arrow.Factory` related to
+  open route service API return values
+  """
+
+  defmacro __using__(_opts) do
```
jzimbel-mbta, line 7: (Feel free to ignore if this module was just copy-pasted from Skate)

Is there any reason why these functions need to be defined in a `__using__` macro, which injects the code directly into the using module?
It seems like if these functions were defined normally within this module, a simple `import Arrow.OpenRouteServiceFactory` would accomplish the same thing as the current `use Arrow.OpenRouteServiceFactory`.
(This module would also need to explicitly `use ExMachina.Ecto, repo: Arrow.Repo`, which imo provides more clarity.)

jzimbel-mbta (approved):
LGTM! Nice and clean. My comments are just a handful of very minor, non-blocking nits.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
meagharty: This was a copy/pasta, will fix

test/support/factories/open_route_service_factory.ex
meagharty: Turns out `__using__`  is what is recommended in the docs for ExMachina for splitting factories into files: https://github.com/beam-community/ex_machina?tab=readme-ov-file#splitting-factories-into-separate-files

I did try it out initially:
The `use ExMachina.Ecto, repo: Arrow.Repo` doesn't play nice in both files (and presumably you also try to retain the "one file for all factories" approach.

```
    error: imported Arrow.OpenRouteServiceFactory.build/1 conflicts with local function
    │
  5 │   use ExMachina.Ecto, repo: Arrow.Repo
    │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    │
    └─ test/support/factory.ex:5: Arrow.Factory (module)
```

PR #1044, opened 2024-12-02
Drag and drop stops list

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 Implement "Create/Edit Shuttle Route Definition" - Reorder Stops Function](https://app.asana.com/0/584764604969369/1208403724207316/f)

The first commit, 84204a4, just adds `typescript-language-server` as a dev-only dependency, which I do to keep it installed across different runs of `npm install` / `npm ci` with how my editor is set up. It should be harmless but I can remove it and work around it some other way if people don't want it in our dependencies.

As for the actual feature implementation, I had some issues getting LiveView hooks to play along nicely with TypeScript - I really wanted some kind of type that's something along the line of "thing that will be used to build a `ViewHook`" but I don't think that exists and I didn't see an easy way to build it using TypeScript magic, so I ended up just using a type assertion (`as ViewHook`). I'd like to be able to avoid that kind of thing, though, so I'm open to suggestions on other approaches I could take.

On the backend, I set it up so that the ordering transposes stops and then does a pass to make sure that all of the `stop_sequence` values are in order, adding to them if necessary. Aside from that it doesn't currently enforce any niceties like starting from a particular value, but if we want something like that we should probably implement it in a more general way.

#### Reviewer Checklist

### DIFF
--- a/assets/src/app.tsx
+++ b/assets/src/app.tsx
@@ -4,7 +4,8 @@ import LiveReact from "./LiveReactPhoenix"
 import live_select from "live_select"
 // Establish Phoenix Socket and LiveView configuration.
 import { Socket } from "phoenix"
-import { LiveSocket } from "phoenix_live_view"
+import { LiveSocket, ViewHook } from "phoenix_live_view"
+import Sortable from "sortablejs"
 
 import ReactPhoenix from "./ReactPhoenix"
 import DisruptionCalendar from "./components/DisruptionCalendar"
@@ -21,8 +22,34 @@ declare global {
   }
 }
 
+const sortable = {
+  mounted() {
+    new Sortable(this.el, {
+      animation: 150,
+      handle: ".drag-handle",
+      draggable: ".item",
+      dragClass: "drag-item",
+      ghostClass: "drag-ghost",
+      forceFallback: true,
+      onEnd: (e) => {
+        const params = {
+          old: e.oldDraggableIndex,
+          new: e.newDraggableIndex,
+          ...e.item.dataset,
+          ...this.el.dataset,
+        }
+        this.pushEventTo(this.el, "reorder_stops", params)
+      },
+    })
+  },
+} as ViewHook
+
 // https://github.com/fidr/phoenix_live_react
-const hooks = { LiveReact, ...live_select }
+const hooks = {
+  LiveReact,
+  sortable,
+  ...live_select,
+}
 
 const csrfToken = document
   .querySelector("meta[name='csrf-token']")!
--- a/lib/arrow_web/live/shuttle_live/shuttle_live.ex
+++ b/lib/arrow_web/live/shuttle_live/shuttle_live.ex
@@ -112,60 +112,68 @@ defmodule ArrowWeb.ShuttleViewLive do
       </div>
       <hr />
       <h2>define stops</h2>
-      <.inputs_for :let={f_route} field={f[:routes]} as={:routes_with_stops}>
-        <h4>direction <%= input_value(f_route, :direction_id) %></h4>
-        <div class="container">
-          <.inputs_for :let={f_route_stop} field={f_route[:route_stops]}>
-            <div class="row">
-              <.input field={f_route_stop[:display_stop_id]} label="Stop ID" class="col-lg-7" />
-              <.input
-                field={f_route_stop[:time_to_next_stop]}
-                type="number"
-                label="Time to next stop"
-                class="col-lg-4"
-              />
-              <button
-                type="button"
-                name={input_name(f_route, :route_stops_drop) <> "[]"}
-                value={f_route_stop.index}
-                phx-click={JS.dispatch("change")}
-                class="col-lg-1"
-              >
-                <.icon name="hero-x-mark-solid" class="h-4 w-4" />
-              </button>
-              <input
-                value={f_route_stop.index}
-                type="hidden"
-                name={input_name(f_route, :route_stops_sort) <> "[]"}
-              />
-              <input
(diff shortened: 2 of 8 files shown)


assets/src/app.tsx
```diff
@@ -21,8 +22,34 @@ declare global {
...
+          old: e.oldDraggableIndex,
+          new: e.newDraggableIndex,
+          ...e.item.dataset,
+          ...this.el.dataset,
+        }
+        this.pushEventTo(this.el, "reorder_stops", params)
+      },
+    })
+  },
+} as ViewHook
```
Whoops, line 45: I don't think the assertion here is inherently bad, but here's how I got this to play with TypeScript in a personal project:
```typescript
class MediaButtonHook {
  mounted: (this: ViewHook) => void;

  constructor(game: Hooks) {
    this.mounted = function (this: ViewHook) {
...

class Hooks {
  MediaButton = new MediaButtonHook(this);
  ...
}
```

Whoops:
Feel free to ignore:

https://github.com/user-attachments/assets/b8753ece-966e-4313-9624-01eba799dbe3


The ghost that I'm dragging around along with the real one confuses my brain. If it's easy, it might be nice to remove the ghost and/or highlight the row that's currently being dragged?

lemald:
> Feel free to ignore:
> Screen.Recording.2024-12-03.at.12.19.38.PM.mov
> 
> The ghost that I'm dragging around along with the real one confuses my brain. If it's easy, it might be nice to remove the ghost and/or highlight the row that's currently being dragged?

This is a good callout, but unfortunately I spent a bit of time poking and I think that the class specified by `ghostClass` applies to both the item in the list and the the actual ghost that you're referring to (the [example here](https://sortablejs.github.io/Sortable/#simple-list) actually shows it well), and I'm not sure how much flexibility the underlying browser APIs give us in addressing that.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -380,6 +422,51 @@ defmodule ArrowWeb.ShuttleViewLive do
...
+    end
+  end
+
+  defp ensure_stop_sequence_order(route_stops) do
+    {new_route_stop_changes, _max_stop_sequence} =
+      Enum.reduce(route_stops, {[], 0}, fn route_stop, {route_stop_changes, prev_stop_sequence} ->
+        current_stop_sequence = route_stop.stop_sequence
+
+        if current_stop_sequence > prev_stop_sequence do
+          {route_stop_changes ++ [Arrow.Shuttles.RouteStop.changeset(route_stop, %{})],
```
Whoops, line 453: Every call to `RouteStop.changeset` generates 1-2 queries in order to populate `display_stop_id`. On my test shuttle, this means that the example shuttle I was testing with reordering a single stop generated ~10 queries.

Suggestion: Maybe that function could check if the Stop/GTFS stop association is already loaded, and if so, generate the display without querying the database.

Alternative suggestion: I wonder if you could do this by using `put_assoc` differently without having to build up a list of changesets. I'm thinking instead you could map over the `route_stops` and build up a list of maps with their ids and their index (to use as the stop_sequence), so then your calling `put_assoc` like
```elixir
      Ecto.Changeset.put_assoc(
          route_changeset,
          :route_stops,
          [%{id: 123, stop_sequence: 0}, %{id: 456, stop_sequence: 1}, ...]
        )
```

Whoops: Related to my second suggestion, more generally I think `ensure_stop_sequence` could be simpler by just using the index in the list as the stop sequence. That also ensure the `stops_sequence` values remain contiguous from 0...n, which I don't think this does.

Whoops (changes requested):
See my comment about using the index for `stop_sequence` (or somehow keeping the sequence 0..n. I also think the extra database queries are worth looking at, but can be talked out of dealing with them in this PR.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
lemald: Yeah, I think I may have somewhat overcomplicated the logic. FWIW as it stands it doesn't necessarily end up being a contiguous range from `0..n`, and just using the index would solve that.

> Every call to `RouteStop.changeset` generates 1-2 queries in order to populate `display_stop_id`. On my test shuttle, this means that the example shuttle I was testing with reordering a single stop generated ~10 queries.
>
> Suggestion: Maybe that function could check if the Stop/GTFS stop association is already loaded, and if so, generate the display without querying the database.

Good point, I'll look into implementing that.

> Alternative suggestion: I wonder if you could do this by using `put_assoc` differently without having to build up a list of changesets. I'm thinking instead you could map over the `route_stops` and build up a list of maps with their ids and their index (to use as the stop_sequence)...

So I considered this particular approach but the thing that made me hesitate is that [the `put_assoc` documentation](https://hexdocs.pm/ecto/Ecto.Changeset.html#put_assoc/4) mentions that this skips some validation and is mainly meant for things like tests, so I'd prefer not to use it (and rather go with the above suggestion of avoiding extraneous queries in the changeset function).

lemald: Okay, what do you think about the changes in b8aacfa? I ended up opting to specifically do the stop lookup only if we're changing `display_stop_id`, and otherwise trust that things are in a consistent state. I'm signing off for the day now, but if you like this approach I'll also see about adding some more tests directly to the changeset function tomorrow.

PR #1045, opened 2024-12-04
feat(shapes): switch map to MBTA tileset

rudiejd:
#### Summary of changes
**Asana Ticket:** [🏹 Implement Skate Tileset](https://app.asana.com/0/inbox/1208877353871379/1208652433539040/1208898327310338)

Change tileset from default OSM tileset to MBTA tileset used on Skate.
image

#### Reviewer Checklist


docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+# docker compose for running locally
+services:
+  postgres:
+    image: postgres:14.1-alpine
+    restart: always
+    environment:
```
rudiejd, line 6: I can remove if we would rather leave this out of version control, but I find a compose file useful for getting postgres up and running

assets/src/components/ShapeViewMap.tsx
```diff
@@ -137,8 +137,8 @@ const ShapeViewMap = ({ shapes }: ShapeViewMapProps) => {
     >
       {polyLines}
       <TileLayer
-        attribution='&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
-        url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
+        attribution='&copy; <a href="https://www.mass.gov/info-details/massgis-data-2021-aerial-imagery">MassGIS 2021</a>'
```
rudiejd, line 140: do we want this to be configurable? I couldn't grok any other examples of configurable values in the React part of the app, this was static before, so I just left it hard-coded. But I don't think that's ideal, since the s3 bucket could change

meagharty: I doubt we'll change the s3 bucket or tileset anytime soon.

Maybe if we hit another instance of changing this or we want to add options within the map view to change the tileset? It doesn't seem worth it to me to plumb that into this part of the app atm but it's up to you.

To your point, afaik we don't have any app-level [config/...](https://github.com/mbta/arrow/tree/master/config) plumbing yet for React (like skate does w/ [appData.ts](https://github.com/mbta/skate/blob/8d6a841d98813185982edd52b8ce5b583503c55c/assets/src/appData.ts) and this server-side [`<div id="app" />`](https://github.com/mbta/skate/blob/8d6a841d98813185982edd52b8ce5b583503c55c/lib/skate_web/controllers/page_html/index.html.heex#L8))

docker-compose.yml
meagharty: this is nice, thank you! makes sense to add it to version control to share with others

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+# docker compose for running locally
+services:
+  postgres:
+    image: postgres:14.1-alpine
```
meagharty, line 4: why `14.1`?

our prod database is running [15.4](https://github.com/mbta/devops/blob/d73667663604d51c5b4d606de59a098f79202025/terraform/aws-tid-main-prod/arrow.tf#L19) so we should use a version of 15

It looks like the README is out of date (says `postgresql@11`). Do you mind updating the README with `15` and adding this docker-compose file as an option for setting up PostgreSQL?

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=arrow_dev
+    ports:
+      - '5432:5432'
+    volumes: 
+      - db:/var/lib/postgresql/data
+volumes:
+  db:
+    driver: local
```
meagharty, line 17: nit: `local` seems to be the default for `driver`, do you specify it here for any reason in particular?

meagharty (changes requested):

docker-compose.yml
rudiejd: i think i just copied this from another project 😅 

i made those changes!

docker-compose.yml
rudiejd: yep, I think I just tend to forget that you can specify the volumes as simply

```
volume:
    foo:
```
without saying a driver, because it feels a little weird. but i think that is indeed the modern way to do it. I changed that

assets/src/components/ShapeViewMap.tsx
rudiejd: I think we can do that as part of https://app.asana.com/0/1203810889612417/1208652433539046/f

meagharty (approved):

PR #1046, opened 2024-12-05
Cognito cleanup

lemald:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹  Ensure Arrow isn't reliant on Cognito config](https://app.asana.com/0/584764604969369/1208882904451687/f)

I ended up managing to (I think!) get rid of references to Cognito entirely, and keep the local login flow working.

To get the fake Ueberauth strategy working, I consulted the Screenplay implementation of the refresh feature in mbta/screenplay#520, as well as Skate's implementation of passing the groups through from their fake strategy.

I have tested out the normal browser login flow locally, as well as doing an API call to my login server (albeit with changing the function clause so I can use an account with a username other than `gtfs_creator_ci@mbta.com`).

I have deployed this to arrow-dev and confirmed that the SSO flow is still working. I also did an API request against `arrow-dev` with `curl` using my account's API token and confirmed it was still working.

Probably the part I'm least familiar with about our whole login flow is the token refresh bit. The real prod code for this should be untouched, but for the fake strategy I'm sort of just mimicking what I saw elsewhere to prevent it from turning into a redirect loop, so that aspect would probably be worth reviewing particularly carefully.

#### Reviewer Checklist


test/arrow_web/controllers/auth_controller_test.exs
```diff
@@ -90,21 +34,11 @@ defmodule ArrowWeb.Controllers.AuthControllerTest do
...
-
-  describe "request" do
-    test "redirects to auth callback", %{conn: conn} do
-      conn = get(conn, Routes.auth_path(conn, :request, "cognito"))
-
-      response = response(conn, 302)
-
-      assert response =~ Routes.auth_path(conn, :callback, "cognito")
-    end
-  end
```
lemald, line 109: Note that our controller doesn't actually implement `request`, so this was essentially testing the fake strategy. The real Ueberauth flow won't actually redirect you straight back to the callback without going through the SSO provider first.

Whoops (approved):
Screenshot 2024-12-05 at 11 18 46 AM

Red is better than green :-)

PR #1048, opened 2024-12-09
feat(shapes): add toggle for base map layer (satellite vs street map)

rudiejd:
#### Summary of changes
**Asana Ticket:** [🏹 Implement MassGIS Aerial Imagery Tileset + Base Map Selection](https://app.asana.com/0/1208877354377530/1208652433539046)
Screenshot 2024-12-09 at 3 28 24 PM

#### Reviewer Checklist

### DIFF
--- a/assets/src/components/BaseMapLayerControl.tsx
+++ b/assets/src/components/BaseMapLayerControl.tsx
@@ -16,8 +16,8 @@ const BaseMapLayerControl = () => {
       </LayersControl.BaseLayer>
       <LayersControl.BaseLayer name="Satellite">
         <TileLayer
-          attribution='&copy; <a href="https://www.mass.gov/info-details/massgis-data-2021-aerial-imagery">MassGIS 2021</a>'
-          url="https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/orthos2021/MapServer/tile/{z}/{y}/{x}"
+          attribution='&copy; <a href="https://www.mass.gov/info-details/massgis-data-2023-aerial-imagery">MassGIS 2023</a>'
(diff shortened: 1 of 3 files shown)


assets/src/components/BaseMapLayerControl.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react"
+import { LayersControl, TileLayer } from "react-leaflet"
+
+const BaseMapLayerControl = () => {
```
rudiejd, line 4: I chose not to collapse the base map layer control when it's not moused-over because we weren't doing that for the other layer control. Let me know if you have another preference

assets/src/components/StopViewMap.tsx
```diff
@@ -11,15 +12,14 @@ type Stop = {
...
-})
-
 const StopViewMap = ({ stop }: { stop?: Stop }) => {
+  const markerIcon = icon({
+    iconUrl: "/images/marker-icon.png",
+    iconRetinaUrl: "/images/marker-icon-2x.png",
+    shadowUrl: "/images/marker-shadow.png",
+    iconSize: [25, 41],
+    iconAnchor: [12, 41],
+  })
```
jzimbel-mbta, line 22: tiny nit:
Any reason for moving `markerIcon` into the component's render? It shouldn't make much difference, but I could see the original author defining it at the module scope for performance reasons, since its value is constant and doesn't need to be recomputed each time `StopViewMap` re-renders.

(One "best of both worlds" option if you do prefer it scoped within the component that uses it, is to memoize it with `useMemo(() => icon(...), [])`)

jzimbel-mbta (approved):
Good to go! Just a non-blocking nit.

assets/src/components/StopViewMap.tsx
rudiejd: I think this was an accident, actually.... I was going to add an icon into the layer toggle and then ended up deciding against it. I'll fix this!

PR #1053, opened 2024-12-12
feat: Upload shuttle definition

cmaddox5:
#### Summary of changes
**Asana Ticket:** [🏹 Implement Shuttle Route Definition Spreadsheet Upload ](https://app.asana.com/0/584764604969369/1207757783807423/f)

This PR adds the ability to upload shuttle definition files. File must be an `xlsx` file with tabs named `Direction 0 STOPS` and `Direction 1 STOPS` (rest are ignored) and expects column B to contain the stop IDs. Once imported, validation runs so the user knows if the stop IDs in the spreadsheet are valid. Will also show an error `flash` if the file is invalid.

Pulled in the `xlsxir` library to help parse these files. Library works by standing up an ETS process for each tab. Once the `pid` is active, some helper functions like `Xlsxir.get_info/2` and `Xlsxir.get_list/1` can be used to get tab names and cell data. The key thing to remember is these `pid`s will be active until `Xlsxir.close/1` is called. I close the tabs we don't need ASAP and close the tabs we need after getting the data. 

#### Reviewer Checklist
- [x] Meets ticket's acceptance criteria
- [x] Tests were added for any new functionality (don't just rely on Codecov)
- [x] This branch was deployed to the staging environment and is currently running with no unexpected increase in warnings, and no errors or crashes.


lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -453,4 +502,138 @@ defmodule ArrowWeb.ShuttleViewLive do
 
     {:noreply, assign(socket, form: form)}
   end
+
+  defp handle_progress(:definition, entry, socket) do
+    socket = clear_flash(socket)
+
+    if entry.done? do
```
rudiejd, line 509: could this combination of `if` and `case` be rewritten as a `cond` statement to reduce the nesting?

lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -453,4 +502,138 @@ defmodule ArrowWeb.ShuttleViewLive do
...
+      |> append_if(
+        headers != ["Stop Name", "Stop ID", "Notes"],
+        "Invalid/missing headers"
+      )
+
+    if Enum.empty?(errors), do: :ok, else: {:errors, errors}
+  end
+
+  defp append_if(list, condition, item) do
+    if condition, do: list ++ [item], else: list
```
rudiejd, line 637: nit: is it worth rewriting this to preprend and reverse instead of append? as I understand, prepending has better performance in elixir since lists are singly-linked lists by default. Or just prepending without reversing if order does not matter.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -453,4 +502,138 @@ defmodule ArrowWeb.ShuttleViewLive do
...
+          )
+        end)
+
+      changeset = Ecto.Changeset.put_assoc(changeset, :routes, new_routes)
+
+      to_form(changeset)
+    end)
+  end
+
+  defp extract_stop_ids_from_upload(socket, entry) do
```
meagharty, line 545: Could we put this function (and other functions specific to xls parsing) outside of `shuttle_live.ex`?

rudiejd (commented):
Should we add type specs to the new functions? It's an item in the code review checklist, but I don't see many type specs in the surrounding code

lib/arrow_web/live/shuttle_live/shuttle_live.ex
cmaddox5: General `LiveView` code health question, are we ok with live component logic (`consume_uploaded_entry` in this case) living outside the component module? I'm thinking it's fine but want to make sure.

cmaddox5: Actually, scratch that. After moving the `with` into the new module, the `consume_upload_entry` could be moved back into `handle_progress/3`.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
cmaddox5: It's probably safe to say order doesn't matter here. Each error needs to be addressed individually so not totally necessary to display them in any particular order.

cmaddox5:
> Should we add type specs to the new functions? It's an item in the code review checklist, but I don't see many type specs in the surrounding code

In Elixir, it's generally only public functions that get specs. That said, I am moving around some functions into a different file which will requiring flipping a function to public. Will definitely add a spec for that one.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
cmaddox5: The issue with `cond` here is the need to pattern match on the error return. In a `cond`, scope is lost with each conditional. That means we would need to call the function in the `case` clause twice in order to make sure we handle both returns appropriately.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
rudiejd: gotcha. I think it would be better to prepend in that case, but I think it's non-blocking since i don't expect the number of rows in one of these files to ever be high cardinality

lib/arrow_web/live/shuttle_live/shuttle_live.ex
rudiejd: gotcha. yeah I think there would be a way to wrangle into a cond using multiple `match?` calls, but that would be less readable

rudiejd:
**Functional Testing**
- Verified that uploading the doc from Asana gave me the expected list of stops, presented in the correct directions
image
image

- Removed the stop ID column and verified that the correct error was shown
image
    - one minor thing is that when there's no stop ID column, you still get an error about a stop ID missing in row 1. but I don't think that's a blocker
    
 - Tried to insert a blank value for stop ID on a row and verified that I got an error
image


- Removed the stop name for a row. There was no validation error. Should we do anything with the stop names from the XLSX?

rudiejd (approved):

cmaddox5:
> one minor thing is that when there's no stop ID column, you still get an error about a stop ID missing in row 1. but I don't think that's a blocker

This should be fixed. I now check for missing columns first and fail early instead of checking it last and adding it to the end of the error list.

> Removed the stop name for a row. There was no validation error. Should we do anything with the stop names from the XLSX?

I don't think missing stop names is something we need to worry about but can double check. The user may want to know about that.

PR #1047, opened 2024-12-05
🏹 Implement "Create/Edit Shuttle Route Definition" Map View - Stop Display

Whoops:
#### Summary of changes
**Asana Ticket:** [🏹 Implement "Create/Edit Shuttle Route Definition" Map View - Stop Display](https://app.asana.com/0/1204473309455397/1208573748126858/f)

[Please include a brief description of what was changed]

#### Reviewer Checklist
- [ ] Meets ticket's acceptance criteria
- [ ] Any new or changed functions have typespecs
- [ ] Tests were added for any new functionality (don't just rely on Codecov)
- [ ] This branch was deployed to the staging environment and is currently running with no unexpected increase in warnings, and no errors or crashes.

PR #1054, opened 2024-12-12
feat(shuttle-stops): shuttle stops map view 

rudiejd:
#### Summary of changes
**Asana Ticket:** [🏹 Implement "Create Shuttle Stops" map view - display existing stops on map ](https://app.asana.com/0/584764604969369/1207472819136804)

- Start passing all shuttle stops and all GTFS stops to the "new stop" and "edit stop" pages
- in the leaflet map components, render a layer with differently colored markers for Arrow stops ("Existing Shuttle Stops") and GTFS stops ("Existing Bus Stops")
- Allow for toggling each layer

Implementation:
image

Compare to the mockup:
image

Discussion thread with product: https://mbta.slack.com/archives/C07639BHFKR/p1734107456821499

Notes:
- I'm fetching every stop and every GTFS stop. This is probably not tenable if the number of stops in Arrow is ever going to be significantly large. If we're concerned about performance, I could try paginating stops, but that would introduce a lot more complexity and probably increase the cycle time of the change
- Shoutout to @Whoops for the pin CSS / HTML, I was banging my head trying to remember how to to CSS stuff. It's stolen from his PR

#### Reviewer Checklist


rudiejd:
I'm not really sure how to add automated tests for a mostly visual/javascript feature like this, save for browser-based tests. Happy to add them if folks think it's important for this though

assets/src/components/stops/types.tsx
```diff
@@ -0,0 +1,15 @@
+type Stop = {
+  stop_name: string
+  stop_desc: string
+  stop_lat: number
+  stop_lon: number
```
rudiejd, line 5: I chose to make these different frontend types even though they have exactly the same data because I figured we might want different information from them in the future, and the schema for these types on the backend is not exactly the same

assets/src/components/stops/StopViewMap.tsx
```diff
@@ -0,0 +1,154 @@
...
+    >
+      {existingShuttleStops && (
+        <LayersControl.Overlay
+          name={generateLegend(
+            existingShuttleStopColor,
+            "Existing Shuttle Stops"
+          )}
+        >
+          <LayerGroup>
+            {existingShuttleStops?.map(
```
cmaddox5, line 75: nit: Shouldn't need to worry about `existingShuttleStops` being `null`-ish here since there is a check above. Same for line 95 below.

assets/src/components/stops/StopViewMap.tsx
```diff
@@ -0,0 +1,154 @@
...
+                  <Marker
+                    position={[s.lat, s.lon]}
+                    icon={genIcon(existingBusStopColor)}
+                  >
+                    <Popup>{s.name}</Popup>
+                  </Marker>
+                )
+            )}
+          </LayerGroup>
+        </LayersControl.Overlay>
```
cmaddox5, line 108: To reduce duplication, might be useful to extract this `LayersControl.Overlay` group into its own component that takes in the data it needs in `props`. Would make future changes to them a little easier.

assets/src/components/stops/StopViewMap.tsx
```diff
@@ -0,0 +1,154 @@
...
+import BaseMapLayerControl from "../BaseMapLayerControl"
+import { Stop, GtfsStop } from "./types"
+
+const defaultCenter: LatLngExpression = [42.360718, -71.05891]
+
+const newStopColor = "f24727"
+const existingShuttleStopColor = "414bb2"
+const existingBusStopColor = "fac711"
+
+const genIcon = (color: string) => {
```
cmaddox5, line 19: It looks like this function will eventually return the same thing as the function in #1047. Makes sense cause they both represent a map marker. Not sure the best way to coordinate this, but should we think about exporting the function and using it in both places?

assets/src/components/stops/types.tsx
cmaddox5: I think this is fine considering the property names are different.

assets/src/components/stops/StopViewMap.tsx
rudiejd: @Whoops do you mind if I move this into a common helper function? not sure whose PR will be merged first 😅

Whoops: I'm all for it! I expect there's a lot of common functionality that could be extracted in a factor, if not making a map component that's used across all pages. But that's probably another ticket :-)

assets/src/components/stops/StopViewMap.tsx
```diff
@@ -0,0 +1,193 @@
...
+  useMapEvents({
+    overlayremove: e => {
+      if (e.name.includes("Bus Stops")) {
+        setLayersEnabled({ ...layersEnabled, busStops: false })
+      }
+      else if (e.name.includes("Shuttle Stops")) {
+        setLayersEnabled({ ...layersEnabled, shuttleStops: false })
+      }
+      else if (e.name.includes("Selected")) {
+        setLayersEnabled({ ...layersEnabled, stop: false })
```
rudiejd, line 136: This is kind of janky. The idea here was to track the state of overlay-layers and change the bounds of the map based on all currently enabled and valid stops

Whoops: Responding to this comment having not read the code; my apologies if this is something already factored in...

I don't think you should consider anything except the currently selected stop when setting the bounds. Otherwise, when bus stops are enabled, you'll essentially end up looking at the greater Boston area when the intent of the feature is to give you an idea of what other stops are near the one you are about to create or are looking at.

rudiejd: gotcha, that will make this way easier then 😅  i will make that change

assets/src/components/stops/StopViewMap.tsx
rudiejd: upon further inspection, I made the zoom/bound-setting code way overly complicated, and after refactoring I didn't need the names of each overlay layer. so I pulled this out

assets/src/components/stops/StopViewMap.tsx
```diff
@@ -0,0 +1,57 @@
...
+interface StopViewMapProps {
+  stop?: Stop
+  existingShuttleStops?: Stop[]
+  existingBusStops?: GtfsStop[]
+}
+
+const MapUpdater = ({ stop }: { stop?: Stop }) => {
+  const map = useMap()
+  useEffect(() => {
+    const bounds = getMapBounds(
```
cmaddox5, line 20: I'm noticing a pretty big pause on dev when checking `Existing Bus Stop`. I think it's because the map is rendering markers outside the visible bounds. If you zoom out a ton, the page will eventually just stop working. Is it possible to prevent rendering outside the visible bounds in `react-leaflet`?

assets/src/components/stops/StopLayerControl.tsx
```diff
@@ -0,0 +1,124 @@
...
+    <LayersControl
+      position="bottomright"
+      key="layer-control-stops"
+      collapsed={false}
+    >
+      {existingShuttleStops && (
+        <LayersControl.Overlay name={shuttleLayerName}>
+          <LayerGroup>
+            {existingShuttleStops
+              .filter(
```
rudiejd, line 61: Discussion about this change: https://mbta.slack.com/archives/C07639BHFKR/p1734467848132149

I opted to only render stops within 1 radius of the stop you are creating / editing. This needs to be done in the client side code because in the case of a new stop, the latitude and longitude aren't propogated to the backend until after the user saves the stop, so they wouldn't be able to see the new stop on a map while editing it

cmaddox5 (approved):
Big performance improvement with the new filter logic. Very cool. Have a couple more small comments/suggestions.

config/runtime.exs
```diff
@@ -59,7 +59,12 @@ if config_env() == :prod do
       capture_log_messages: true
   end
 
-  config :arrow, ArrowWeb.Endpoint, secret_key_base: System.fetch_env!("SECRET_KEY_BASE")
+  config :arrow, ArrowWeb.Endpoint,
+    secret_key_base: System.fetch_env!("SECRET_KEY_BASE"),
+    check_origin: [
```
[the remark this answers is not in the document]
rudiejd: yeah, I can do that. I just had to cherry pick that commit for the map to display on dev

cmaddox5 (changes requested):
Sorry to retract approval, but I think we should actually reconsider how we load GTFS stops on page load. Going and getting all stops will be very taxing considering how many we have. Now that I look at dev again, you can actually see the effect of the performance issue on the map. It takes a second or more to load after the rest of the page.

assets/src/components/stops/StopLayerControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+  existingShuttleStops: Stop[] | undefined
+  existingBusStops: GtfsStop[] | undefined
+}) => {
+  return (
+    <LayersControl
+      position="bottomright"
+      key="layer-control-stops"
+      collapsed={false}
+    >
+      {existingShuttleStops && existingShuttleStops.length > 0 && (
```
cmaddox5, line 35:
```suggestion
      {existingShuttleStops?.length > 0 && (
```

assets/src/components/stops/StopLayerControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+        <LayersControl.Overlay name={shuttleLayerName}>
+          <LayerGroup>
+            {existingShuttleStops
+              .filter(
```
cmaddox5, line 39: Now that the server does this, shouldn't need to do it here.

assets/src/components/util/maps.tsx
```diff
@@ -0,0 +1,86 @@
...
+    html: `<div id="${stopId}"><span style="${markerHtmlStyles}"><span style="${innerCircleStyles}"></span></div>`,
+  })
+}
+
+// Generates a label used for the legend of a map
+const generateLegend = (color: string, name: string) =>
+  `<div class="legend-square" style="background-color: #${color}"></div> ${name}`
+
+// Calculates the distance in miles between two latitude / longitude pairs
+const haversineDistanceMiles = (
```
cmaddox5, line 59: Shouldn't need this anymore.

assets/tests/components/StopViewMap.test.tsx
```diff
@@ -1,13 +1,57 @@
...
+      .querySelectorAll("input.leaflet-control-layers-selector")
+      .forEach((e) => fireEvent.click(e))
+
+    // show stop that is less than 1 mile away from the selected stop
+    expect(container.querySelectorAll("#arrow-stop-1234").length > 1)
+
+    // don't show stop that is more than 1 mile away from the selected stop
+    expect(container.querySelectorAll("#gtfs-stop-5678").length === 0).toBe(
+      true
+    )
```
cmaddox5, line 55: I think the assertion above this one is good to keep, but now that the server does the distance work you don't need to test against it here.

config/runtime.exs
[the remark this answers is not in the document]
cmaddox5: Just a reminder to remove this commit before merging.

lib/arrow/gtfs/stop.ex
```diff
@@ -81,6 +85,43 @@ defmodule Arrow.Gtfs.Stop do
...
+
+  ## Examples
+      iex> Arrow.Gtfs.Stop.get_stops_within_mile("123", {42.3774, -72.1189})
+      [%Arrow.Gtfs.Stop, ...]
+
+      iex> Arrow.Gtfs.Stop.get_stops_within_mile(nil, {42.3774, -72.1189})
+      [%Arrow.Gtfs.Stop, ...]
+  """
+  @spec get_stops_within_mile(String.t() | nil, {float(), float()}) :: list(Arrow.Gtfs.Stop.t())
+  def get_stops_within_mile(arrow_stop_id, {lat, lon}) do
```
cmaddox5, line 113: If I'm reading this right, you should be able to combine both of these functions. The ID comparison would be harmless if it is `nil` because no ID will equal `nil` anyway and would be `true` for all rows.

lib/arrow/gtfs/stop.ex
```diff
@@ -81,6 +85,43 @@ defmodule Arrow.Gtfs.Stop do
...
+    from(s in Arrow.Gtfs.Stop,
+      where:
+        s.id != ^arrow_stop_id and
+          s.lat <= ^lat + @latitude_degrees_per_mile and
```
cmaddox5, line 117: Should we consider a new index for lat and long columns now? I don't think this table has any atm.

lib/arrow/stops.ex
```diff
@@ -124,6 +124,44 @@ defmodule Arrow.Stops do
...
+  ## Examples
+      iex> Arrow.Stops.get_stops_within_mile("123", {42.3774, -72.1189})
+      [%Arrow.Shuttles.Stop%{}, ...]
+
+      iex> Arrow.Stops.Stop.get_stops_within_mile(nil, {42.3774, -72.1189})
+      [%Arrow.Shuttles.Stop{}, ...]
+  """
+  @spec get_stops_within_mile(String.t() | nil, {float(), float()}) ::
+          list(Arrow.Shuttles.Stop.t())
+  def get_stops_within_mile(stop_id, {lat, lon}) do
```
cmaddox5, line 153: Can combine these, too.

lib/arrow/stops.ex
```diff
@@ -124,6 +124,44 @@ defmodule Arrow.Stops do
...
+    from(s in Stop,
+      where:
+        s.stop_id != ^stop_id and
+          s.stop_lat <= ^lat + @latitude_degrees_per_mile and
```
cmaddox5, line 157: Same question on indexes for this table.

lib/arrow_web/live/stop_live/stop_live.ex
```diff
@@ -109,17 +119,66 @@ defmodule ArrowWeb.StopViewLive do
       |> assign(:stop, %Stop{})
       |> assign(:title, "create shuttle stop")
       |> assign(:stop_map_props, %{})
+      |> assign(:existing_stops, nil)
+      |> assign(:existing_gtfs_stops, nil)
```
cmaddox5, line 123: suggestion: Default these to `[]` and you can remove a lot of `null` checks in the client code.

lib/arrow_web/live/stop_live/stop_live.ex
```diff
@@ -109,17 +119,66 @@ defmodule ArrowWeb.StopViewLive do
...
+        :math.cos(degrees_to_radians(other_lat)) *
+        :math.sin(delta_lon_rads / 2) *
+        :math.sin(delta_lon_rads / 2) +
+        :math.sin(delta_lat_rads / 2) *
+          :math.sin(delta_lat_rads / 2)
+
+    line = 2 * :math.atan2(:math.sqrt(arc), :math.sqrt(1 - arc))
+
+    earth_radius_miles * line
+  end
```
cmaddox5, line 150: These appear to be unused.

lib/arrow_web/live/stop_live/stop_live.ex
```diff
@@ -109,17 +119,66 @@ defmodule ArrowWeb.StopViewLive do
...
+        _ -> {nil, nil}
+      end
+
+    {existing_stops, existing_gtfs_stops} =
+      if is_float(lat) and is_float(lon) do
+        {Stops.get_stops_within_mile(stop_id, {lat, lon}),
+         GtfsStop.get_stops_within_mile(stop_id, {lat, lon})}
+      else
+        {nil, nil}
+      end
```
cmaddox5, line 172: Could make this a bit more concise:
```suggestion
    {existing_stops, existing_gtfs_stops} =
      case stop_params do
        %{"stop_lat" => lat, "stop_lon" => lon, "id" => stop_id}
        when not is_nil(lat) and not is_nil(lon) ->
          lat_float = String.to_float(lat)
          lon_float = String.to_float(lon)

          {Stops.get_stops_within_mile(stop_id, {lat_float, lon_float}),
           GtfsStop.get_stops_within_mile(stop_id, {lat_float, lon_float})}

        _ ->
          {nil, nil}
      end
```

test/arrow/stops_test.exs
```diff
@@ -180,5 +180,27 @@ defmodule Arrow.StopsTest do
     test "get_stop_by_stop_id/1 returns nil when stop not found" do
       assert Stops.get_stop_by_stop_id("nonexistent") == nil
     end
+
+    test "get_stops_within_mile/2 returns stops roughly within one mile of a stop" do
```
cmaddox5, line 184: Can a similar test be added for GTFS Stops? Either that or find a way to let them share the same function and test the shared version.

assets/src/components/stops/StopLayerControl.tsx
rudiejd: The server actually does a more approximate version of this. if, for some reason, we wanted to do an exact 1 mile radius, we could keep it this way. But I can remove this for now

test/arrow/stops_test.exs
rudiejd: There's a test for gtfs_stops here: https://github.com/mbta/arrow/blob/afe44459b0434ea8a413ff7ffbfe46617c651ba9/test/arrow/gtfs/stop_test.exs#L6

cmaddox5: That gives a 404. Does it still need to be committed/pushed?

lib/arrow_web/live/stop_live/stop_live.ex
rudiejd: oops, forgot to remove those 😅

lib/arrow/gtfs/stop.ex
rudiejd: I actually was getting a runtime exception when I compared with `nil` in the Ecto code: https://github.com/elixir-ecto/ecto/blob/e644970b6b22369f2acec58ed5c598685488eee8/lib/ecto/query/builder/filter.ex#L191

I could handle this with branching logic rather than pattern matching on the function signature if you prefer that

test/arrow/stops_test.exs
rudiejd: oops, I forgot to `git add` that new file 😭 should be good now

lib/arrow/gtfs/stop.ex
cmaddox5: Hmm TIL! Haven't worked with Ecto a ton yet.

But the branching logic might be better just for the sake of keeping all the math in one place.

lib/arrow/gtfs/stop.ex
rudiejd: query plan before index
```
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on gtfs_stops  (cost=0.00..472.75 rows=2 width=279)
   Filter: ((lat >= '42.3774'::double precision) AND (lat <= '42.3784'::double precision) AND (lon >= '-71.1182'::double precision) AND (lon <= '-71.1082'::double precision) AND ((id)::text <> '1'::text))
(2 rows)
```

rudiejd: query plan after index
```
 Bitmap Heap Scan on gtfs_stops  (cost=4.98..12.48 rows=2 width=279)
   Recheck Cond: ((lat >= '42.3774'::double precision) AND (lat <= '42.3784'::double precision) AND (lon >= '-71.1182'::double precision) AND (lon <= '-71.1082'::double precision))
   Filter: ((id)::text <> '1'::text)
   ->  Bitmap Index Scan on gtfs_stops_lat_lon_vehicle_type_id_index  (cost=0.00..4.98 rows=2 width=0)
         Index Cond: ((lat >= '42.3774'::double precision) AND (lat <= '42.3784'::double precision) AND (lon >= '-71.1182'::double precision) AND (lon <= '-71.1082'::double precision))
```

reduces total cost from 472.75 to 12.48

lib/arrow/stops.ex
rudiejd: same problem as here: https://github.com/mbta/arrow/pull/1054#discussion_r1892591201

lib/arrow/gtfs/stop.ex
rudiejd: credo marks this is a refactoring opportunity when I try to branch the query, hah
```elixir
  def get_stops_within_mile(arrow_stop_id, {lat, lon}) do
    query =
      if is_nil(arrow_stop_id) do
        from(s in Arrow.Gtfs.Stop,
          where:
            s.lat <= ^lat + @latitude_degrees_per_mile and
              s.lat >= ^lat - @latitude_degrees_per_mile and
              s.lon <= ^lon + @longitude_degrees_per_mile and
              s.lon >= ^lon - @latitude_degrees_per_mile and
              s.vehicle_type == :bus
        )
      else
        from(s in Arrow.Gtfs.Stop,
          where:
            s.id != ^arrow_stop_id and
              s.lat <= ^lat + @latitude_degrees_per_mile and
              s.lat >= ^lat - @latitude_degrees_per_mile and
              s.lon <= ^lon + @longitude_degrees_per_mile and
              s.lon >= ^lon - @latitude_degrees_per_mile
        )
      end

    query |> Repo.all()
  end
```

```
[F] → Function is too complex (cyclomatic complexity is 10, max is 9).
┃       lib/arrow/gtfs/stop.ex:103:7 #(Arrow.Gtfs.Stop.get_stops_within_mile)
```

I think the cyclomatic complexity calculation is incorrect for macros

rudiejd: I found a way to rewrite this using dynamic query expressions 👍

assets/src/components/stops/types.tsx
```diff
@@ -0,0 +1,17 @@
+type Stop = {
+  stop_name?: string
+  stop_desc?: string
+  stop_lat?: number
+  stop_lon?: number
+  stop_id?: string
```
cmaddox5, line 6: I believe all of these are required fields so they shouldn't need to be nullable.

assets/src/components/util/maps.tsx
```diff
@@ -0,0 +1,58 @@
+import { divIcon, LatLngBoundsExpression, LatLngTuple } from "leaflet"
+
+// Generates minimum and maximum coordinate pairs given a list of coordinate pairs
+const getMapBounds = (
+  coords: LatLngTuple[]
+): LatLngBoundsExpression | undefined => {
+  if (coords.length === 0) {
+    return undefined
```
cmaddox5, line 8: nit: `null` feels like a more appropriate return here to me.

assets/src/components/stops/types.tsx
cmaddox5: Or is this specifically for `/new`?

lib/arrow/gtfs/stop.ex
```diff
@@ -81,6 +85,44 @@ defmodule Arrow.Gtfs.Stop do
...
+    conditions =
+      dynamic(
+        [s],
+        s.lat <= ^lat + @latitude_degrees_per_mile and
+          s.lat >= ^lat - @latitude_degrees_per_mile and
+          s.lon <= ^lon + @longitude_degrees_per_mile and
+          s.lon >= ^lon - @latitude_degrees_per_mile
+      )
+
+    conditions =
```
cmaddox5, line 113: When `arrow_stop_id` is `nil`, I believe `conditions` would then be set to `nil` and you would lose all the above conditions. You could solve it be doing something like:
```elixir
conditions = if is_nil(arrow_stop_id), do: conditions, else: dynamic([s], s.id != ^arrow_stop_id and ^conditions)
```

lib/arrow/stops.ex
```diff
@@ -124,6 +124,45 @@ defmodule Arrow.Stops do
...
+      dynamic(
+        [s],
+        s.stop_lat <= ^lat + @latitude_degrees_per_mile and
+          s.stop_lat >= ^lat - @latitude_degrees_per_mile and
+          s.stop_lon <= ^lon + @longitude_degrees_per_mile and
+          s.stop_lon >= ^lon - @latitude_degrees_per_mile
+      )
+
+    conditions =
+      if not is_nil(stop_id) do
```
cmaddox5, line 154: Same concern as above.

lib/arrow/gtfs/stop.ex
```diff
@@ -81,6 +85,43 @@ defmodule Arrow.Gtfs.Stop do
     |> assoc_constraint(:parent_station)
   end
 
+  @longitude_degrees_per_mile 1 / 54.6
+  @latitude_degrees_per_mile 1 / 69
+  def get_stops_within_mile(nil, {lat, lon}) do
+    from(s in Arrow.Gtfs.Stop,
+      where:
+        s.lat <= ^lat + @latitude_degrees_per_mile and
```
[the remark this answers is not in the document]
cmaddox5: I see `vehicle_type` was added to the index but am not seeing it in the query.

cmaddox5 (approved):
Looking good! Last things are pretty small. Thanks for bearing with me. Shouldn't need to see it on dev anymore so should be good to pull out the `Endpoint` change now.

assets/src/components/stops/types.tsx
rudiejd: yep, this is for `/new`

lib/arrow/gtfs/stop.ex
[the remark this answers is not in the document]
rudiejd: oops, I think i wiped it out when switching to use dynamic ecto expressions. bringing that back

PR #1059, opened 2024-12-17
fix: add handling for retrieve estimates on new shuttle creation

meagharty:
#### Summary of changes
**Asana Ticket:** [Follow up on "🏹 Implement "Create/Edit Shuttle Route Definition" - Time to Next Stop"](https://app.asana.com/0/584764604969369/1209003091082862/f)

* Add fetching of stop data (lat/long) when the initial shuttle hasn't been created yet
* Add explicit error handling for when there is only one stop defined for a direction

Part of this is similar to what the Map component does (Ecto.assoc_loaded? & calling `preload`) but if those values aren't set yet, then this uses a virtual field `display_stop` mirroring `display_stop_id`. This isn't the most elegant solution, opening this as a draft for now.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/shuttles.ex
+++ b/lib/arrow/shuttles.ex
@@ -369,11 +369,49 @@ defmodule Arrow.Shuttles do
 
   @spec get_stop_coordinates(RouteStop.t() | Stop.t() | GtfsStop.t()) ::
           {:ok, map()} | {:error, any}
+  def get_stop_coordinates(%RouteStop{display_stop: nil, display_stop_id: nil}) do
+    {:error, "Incomplete stop data, please check your input"}
+  end
+
+  def get_stop_coordinates(%RouteStop{id: nil, display_stop: nil, display_stop_id: id}) do
+    {:error, "Missing lat/lon data for stop #{id}"}
+  end
+
+  def get_stop_coordinates(%RouteStop{id: nil, display_stop: display_stop, display_stop_id: id}) do
+    if id,
+      do: get_stop_coordinates(display_stop),
+      else: {:error, "Missing id for stop"}
+  end
+
+  def get_stop_coordinates(%RouteStop{
+        display_stop: %Stop{} = display_stop,
+        display_stop_id: _display_stop_id
+      }) do
+    get_stop_coordinates(display_stop)
+  end
+
+  def get_stop_coordinates(%RouteStop{
+        display_stop: %GtfsStop{} = display_stop,
+        display_stop_id: _display_stop_id
+      }) do
+    get_stop_coordinates(display_stop)
+  end
+
   def get_stop_coordinates(%RouteStop{gtfs_stop: stop, stop: nil}) do
+    stop =
+      if Ecto.assoc_loaded?(stop),
+        do: stop,
+        else: Arrow.Repo.preload(stop, :gtfs_stop, force: true)
+
     get_stop_coordinates(stop)
   end
 
   def get_stop_coordinates(%RouteStop{gtfs_stop: nil, stop: stop}) do
+    stop =
+      if Ecto.assoc_loaded?(stop),
+        do: stop,
+        else: Arrow.Repo.preload(stop, :stop, force: true)
+
     get_stop_coordinates(stop)
   end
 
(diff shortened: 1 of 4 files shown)


lib/arrow/shuttles.ex
```diff
@@ -385,10 +395,20 @@ defmodule Arrow.Shuttles do
...
+  end
+
+  def get_stop_coordinates(%RouteStop{id: nil, display_stop: nil, display_stop_id: id}) do
     {:error, "Missing lat/lon data for stop #{id}"}
   end
 
+  def get_stop_coordinates(%RouteStop{id: nil, display_stop: stop, display_stop_id: id}) do
+    if id,
+      do: get_stop_coordinates(stop),
+      else: {:error, "Missing id for stop #{id}"}
```
meagharty, line 409: I should update this error message to remove `#{id}`, it's likely to be an empty string as far as existing code paths go

lib/arrow/shuttles.ex
```diff
@@ -385,10 +395,20 @@ defmodule Arrow.Shuttles do
...
+    {:error, "Incomplete stop data, please check your inputs"}
+  end
+
+  def get_stop_coordinates(%RouteStop{id: nil, display_stop: nil, display_stop_id: id}) do
     {:error, "Missing lat/lon data for stop #{id}"}
   end
 
+  def get_stop_coordinates(%RouteStop{id: nil, display_stop: stop, display_stop_id: id}) do
+    if id,
+      do: get_stop_coordinates(stop),
```
lemald, line 408: **question:** I may be missing something, but since `display_stop` is either a Arrow shuttle stop or a GTFS stop, but not a `RouteStop`, won't this fall through the other clauses of `get_stop_coordinates/1` and end up in the error state?

lib/arrow/shuttles/route_stop.ex
```diff
@@ -12,6 +12,7 @@ defmodule Arrow.Shuttles.RouteStop do
           stop_sequence: integer(),
           time_to_next_stop: float(),
           display_stop_id: String.t(),
+          display_stop: map(),
```
lemald, line 15: **nitpick**: I don't think it would help much in terms of providing better typechecking, but I think for documentation purposes at least this could be a bit more specific:
```suggestion
          display_stop: Arrow.Shuttles.Stop.t() | Arrow.Gtfs.Stop.t() | nil,
```

lib/arrow/shuttles.ex
meagharty: In this case, gtfs_stop and stop itself are not populated yet on the shuttle because the shuttle hasn't been created yet.

We call `get_stop_coordinates` again here, but now on `display_stop`. If there is a `display_stop_id`  and a `display_stop`, then it'll match the clauses on line 390 or 394 for %GtfsStop{} or %Stop{} on this call.

lib/arrow/shuttles/route_stop.ex
```diff
@@ -24,6 +25,7 @@ defmodule Arrow.Shuttles.RouteStop do
     field :stop_sequence, :integer
     field :time_to_next_stop, :decimal
     field :display_stop_id, :string, virtual: true
+    field :display_stop, :map, virtual: true
```
meagharty, line 28: I'm not sure we really need both `display_stop` and `display_stop_id` but it reduces the changes required across other pieces of work.

lib/arrow/shuttles.ex
lemald: Ahh, sorry, I hadn't expanded the view enough to see 390 and 394 since those aren't directly in the diff. That makes sense!

lib/arrow/shuttles/route_stop.ex
lemald: I think it's useful to have both because, in the case of an Arrow stop, the `display_stop_id` is the value from the  `stop_id` column, not the actual database primary key. We could definitely work around that with some helper function to get the display ID for a given `display_stop`, but I'm not sure how that would play along with form rendering. Anyhow, long story short I think it's fine to keep it this way.

lemald (approved):
There's the minor thing about the type on `display_stop`, but otherwise looks good to me! 👍

lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -374,6 +376,7 @@ defmodule ArrowWeb.ShuttleViewLive do
       |> assign(:form, form)
       |> assign(:form_action, "create")
       |> assign(:http_action, ~p"/shuttles")
+      |> assign(:shuttle, shuttle)
```
lemald, line 379: **question:** isn't this a duplicate of line 381?

lemald (approved):
👍

PR #1068, opened 2024-12-30
tweak: Remove depth option from heroicons

cmaddox5:
#### Summary of changes
While going through Dependabot PRs, I noticed that this dependency was being changed in every PR it opened. This is because `depth` is an option added in elixir 1.17 and dependabot runs with 1.14. Because it has no knowledge of the option, it drops it every time. I don't think having it is saving us much in clone time or memory so it should be fine to remove it. This will allow us to look at dependabot PRs without needing to run `mix deps.get` to get an up-to-date lock file.

#### Reviewer Checklist


lemald (approved):
Interesting catch. Just curious, do you know if there's a way to get dependabot to use a newer version of Elixir, or do we just have to wait?

cmaddox5:
The version number is unfortunately hardcoded into their [Dockerfile](https://github.com/dependabot/dependabot-core/blob/c3127ffb2d79b75abfdddb2ff7eb37c72e15f7e2/hex/Dockerfile#L22) so I think we just have to wait on them. They seem to be pretty behind in general but I do some some open PRs/issues that indicate they are aware of how behind they are.

PR #1069, opened 2024-12-31
dependabot: Ignore path dependencies

cmaddox5:
#### Summary of changes

Dependabot has issues with [path dependencies](https://github.com/mbta/arrow/actions/runs/12560209643/job/35017137053#step:3:172). A way I have solved this in the past is to remove them, but arrow actually uses some of them in React. I'm not 100% sure simply ignoring them will work so I want to try it out. We can't get npm PRs from Dependabot until we fix the issue.

I also added an ignore for `heroicons` because we use a specific tagged version of it. It is also causing Dependabot to fail for the `mix` eco.

#### Reviewer Checklist


lemald (approved):
Fingers crossed that this works! 🤞

PR #1082, opened 2025-01-09
New replacement service component

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 Implement "Create New Replacement Service Component" basic view](https://app.asana.com/0/584764604969369/1208885496650094/f)

This allows a user to click and start entering a replacement shuttle, including selecting the name of the shuttle from a dropdown of available shuttles, with auto-complete. Validation and save are explicitly out of scope and will go in a followup ticket / PR.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/shuttles.ex
+++ b/lib/arrow/shuttles.ex
@@ -466,8 +466,7 @@ defmodule Arrow.Shuttles do
 
   @spec stops_or_gtfs_stops_by_search_string(String.t()) :: [Stop.t() | GtfsStop.t()]
   def stops_or_gtfs_stops_by_search_string(string) do
-    # See https://github.blog/engineering/like-injection/
-    sanitized_string = Regex.replace(~r/([\%_])/, string, fn _, x -> "\\#{x}" end) <> "%"
+    sanitized_string = Arrow.Util.sanitized_string_for_sql_like(string)
 
     stops_query =
       from(s in Stop,
--- a/lib/arrow/shuttles.ex
+++ b/lib/arrow/shuttles.ex
@@ -492,4 +492,12 @@ defmodule Arrow.Shuttles do
 
     Enum.concat(matching_stops, matching_gtfs_stops)
   end
+
+  def shuttles_by_search_string(string) do
+    sanitized_string = Arrow.Util.sanitized_string_for_sql_like(string)
+
+    shuttles_query = from(s in Shuttle, where: ilike(s.shuttle_name, ^sanitized_string))
+
+    Repo.all(shuttles_query)
+  end
 end
--- a/lib/arrow_web/components/core_components.ex
+++ b/lib/arrow_web/components/core_components.ex
@@ -415,6 +415,7 @@ defmodule ArrowWeb.CoreComponents do
   attr :value_mapper, :any
   attr :allow_clear, :boolean
   attr :target, :any, default: nil
+  attr :update_min_len, :integer
 
   def live_select(%{field: %Phoenix.HTML.FormField{} = field} = assigns) do
     assigns =
@@ -481,7 +482,35 @@ defmodule ArrowWeb.CoreComponents do
       field={@field}
       stop_or_gtfs_stop={@stop_or_gtfs_stop}
       class={@class}
-      s
+    />
+    """
+  end
+
+  @doc """
+  LiveSelect-based input for shuttle ID with autocomplete
+  """
+
+  attr :id, :string
+  attr :field, :any, required: true, doc: "Field for `shuttle_id` value"
+
+  attr :shuttle, :any, required: true, doc: "Currently selected shuttle, if any"
+
+  attr :label, :string, default: "Stop ID"
+  attr :class, :string, default: nil
+
+  def shuttle_input(assigns) do
+    assigns =
+      assign_new(assigns, :id, fn %{field: field} ->
+        "#{field.form.name}_#{field.name}_shuttle_input_component"
+      end)
+
+    ~H"""
+    <.live_component
+      module={ArrowWeb.ShuttleInput}
+      id={@id}
+      field={@field}
+      shuttle={@shuttle}
+      class={@class}
     />
     """
   end
--- a/lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
+++ b/lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
@@ -135,6 +135,33 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
       <div :if={@adding_new_service?} class="border-2 border-dashed border-primary p-2">
         <span class="text-primary">add new replacement service component</span>
         <.shuttle_input field={@form[:new_shuttle_id]} shuttle={input_value(@form, :new_shuttle)} />
+        <div class="row">
+          <.input
+            field={@form[:new_shuttle_start_date]}
+            type="date"
+            label="Start date"
+            class="col-lg-3"
+          />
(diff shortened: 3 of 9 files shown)


lib/arrow/shuttles.ex
```diff
@@ -493,4 +492,12 @@ defmodule Arrow.Shuttles do
 
     Enum.concat(matching_stops, matching_gtfs_stops)
   end
+
+  def shuttles_by_search_string(string) do
+    sanitized_string = Arrow.Util.sanitized_string_for_sql_like(string)
+
+    shuttles_query = from(s in Shuttle, where: ilike(s.shuttle_name, ^sanitized_string))
```
rudiejd, line 499: off-topic: why are we using `LIKE` here? postgres has great support for text search via `tsvector` and `tsquery`: https://www.postgresql.org/docs/current/datatype-textsearch.html

plus I'm not sure that a `btree`-type index is the best choice for a text search https://www.postgresql.org/docs/current/textsearch-indexes.html

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
```diff
@@ -17,6 +19,12 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
       "Silver Line": :silver_line
     }
 
+  attr :id, :string
+  attr :form, :any, required: true
+  attr :disruption_v2, :any
+  attr :icon_paths, :any
```
rudiejd, line 25: thank you for adding these 🙏

lib/arrow/shuttles.ex
lemald: So I considered `tsvector` / `tsquery` for the stop ID autocomplete but ended up not using it there because of a couple of issues that I'm trying to remember. I know one is that the tokenizing in the default setup seemed to happen at the word rather than character level, which felt a bit "off" from how most autocomplete functions work, though now that I see the part of the documentation about marking lexemes as prefixes maybe that would be a suitable solution. I'd definitely be happy to revisit it at some point, I know `LIKE` isn't really the best, plus I think `tsvector` and `tsquery` would probably let us avoid having to do any of our own escaping logic.

lib/arrow_web/components/core_components.ex
```diff
@@ -481,7 +482,35 @@ defmodule ArrowWeb.CoreComponents do
...
+  attr :id, :string
+  attr :field, :any, required: true, doc: "Field for `shuttle_id` value"
+
+  attr :shuttle, :any, required: true, doc: "Currently selected shuttle, if any"
+
+  attr :label, :string, default: "Stop ID"
+  attr :class, :string, default: nil
+
+  def shuttle_input(assigns) do
+    assigns =
```
rudiejd, line 502: I like abstracting away the `LiveComponent` initialization like this 👍

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
```diff
@@ -153,11 +218,22 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
     {:noreply, assign(socket, form: form)}
   end
 
-  @impl true
   def handle_event("save", %{"disruption_v2" => disruption_v2_params}, socket) do
     save_disruption_v2(socket, socket.assigns.form_action, disruption_v2_params)
   end
 
+  def handle_event("add_new_replacement_service", _params, socket) do
```
rudiejd, line 225: question for my own understanding: why is this event handled by the parent live view instead of the live select component?

lemald: The LiveSelect component is just for the one input for shuttle. The `add_new_replacement_service` event represents the act of opening the section to add all of the field of a new replacement service, including start and end dates, which aren't handled by the LiveSelect.

lib/arrow/shuttles.ex
rudiejd: worth noting that this probably matters way more for the `gtfs_stops` / `stops` autocomplete than for the shuttles... unless we suddenly get a huge number of shuttles 🙃

lemald: > worth noting that this probably matters way more for the `gtfs_stops` / `stops` autocomplete than for the shuttles... unless we suddenly get a huge number of shuttles 🙃

Right. I still think it's worth thinking through the correct way of doing this, though, even if it's not worth changing for this one PR.

As an aside, I also considered further abstracting out the autocomplete functionality so there's some abstract "autocompletable" behaviour or something like that, but I didn't immediately see a clean way to do that and it felt like overkill for the time being.

rudiejd:
Deploying to dev-green to play around

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
```diff
@@ -103,6 +113,59 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
...
+        <.shuttle_input field={@form[:new_shuttle_id]} shuttle={input_value(@form, :new_shuttle)} />
+        <div class="row">
+          <.input
+            field={@form[:new_shuttle_start_date]}
+            type="date"
+            label="Start date"
+            class="col-lg-3"
+          />
+          <.input field={@form[:new_shuttle_end_date]} type="date" label="End date" class="col-lg-3" />
+          <.input field={@form[:new_shuttle_activation_reason]} type="text" label="Activation reason" />
```
rudiejd, line 145: why are the input fields not a part of `render/1` in the `shuttle_input` component?

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
```diff
@@ -103,6 +113,59 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
...
+        </div>
+        <div class="row">
+          <div class="col-lg-3">
+            <.button disabled={not Enum.empty?(@form.source.errors)} class="btn btn-primary w-100">
```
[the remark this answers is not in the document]
lemald: I thought the idea was that it had to, because until the disruption is saved we would get a foreign key error if we tried to insert a replacement service with no disruption. Of course, this leaves the situation where there already _is_ a disruption... I'm tempted to say that if we do want to handle those differently, then when implementing the actual save, we can add a `phx-click` that explicitly fires some action that does the relevant checking and saves or doesn't, if that makes sense.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
lemald: I wanted to have the `shuttle_input` component be strictly about choosing a shuttle, so that other places in the application that might involve choosing a shuttle but not (for instance) a date could reuse it.

rudiejd: ahh gotcha, I was not sure whether that component was supposed to be the entire form or just a form input

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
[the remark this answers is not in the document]
rudiejd: yep, I was thinking about the situation where there is already a disruption and you might not want to save the changes to said disruption. having said that, I think this is probably a marginal issue and can be handled in Christian's ticket which actually saves the disruption limit.

rudiejd (approved):

PR #1086, opened 2025-01-13
fix: validate time_to_next stop for activating shuttle

lemald:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹🐛 Shuttles can be saved as active without time to next stop](https://app.asana.com/0/584764604969369/1209094056922520/f)

I wonder if it would also make sense to add validation that the final stop in each direction does _not_ have a `time_to_next_stop`, since it isn't meaningful. That might be a bit over-zealous, though.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/shuttles/shuttle.ex
+++ b/lib/arrow/shuttles/shuttle.ex
@@ -47,11 +47,7 @@ defmodule Arrow.Shuttles.Shuttle do
 
           routes
           |> Enum.map(&get_assoc(&1, :route_stops))
-          |> Enum.any?(fn route_stops ->
-            route_stops
-            |> Enum.slice(0..-2//1)
-            |> Enum.any?(&(&1 |> get_field(:time_to_next_stop) |> is_nil()))
-          end) ->
+          |> Enum.any?(&route_stops_missing_time_to_next_stop?/1) ->
             add_error(
               changeset,
               :status,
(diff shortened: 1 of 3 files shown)


cmaddox5:
Would it be useful/easy to disable the input for the last stop in each direction? Then would could enforce that validation by not allowing it in the first place.

lib/arrow/shuttles/shuttle.ex
```diff
@@ -41,13 +41,25 @@ defmodule Arrow.Shuttles.Shuttle do
...
-          changeset
-        else
-          add_error(changeset, :status, "must have at least two stops in each direction")
+          routes
+          |> Enum.map(&get_assoc(&1, :route_stops))
+          |> Enum.any?(fn route_stops ->
+            route_stops
+            |> Enum.slice(0..-2//1)
+            |> Enum.any?(&(&1 |> get_field(:time_to_next_stop) |> is_nil()))
+          end) ->
```
cmaddox5, line 54: suggestion: I find this hard to follow. Could you pull one or both of these pipelines into named functions?

lemald: Good call - what do you think of 57abfcc ?

cmaddox5: Love it!

lemald:
> Would it be useful/easy to disable the input for the last stop in each direction? Then would could enforce that validation by not allowing it in the first place.

Hmm, that's an interesting idea but it would also prevent adding a time to next stop to the final stop on a not-yet-active shuttle, which you might want to do at an intermediate stage of working on it. It sounds like this is actually a more complicated idea than I thought, so product / UX should be involved in deciding if we want to do it and it's probably out of scope for this particular ticket, now that I think more about it.

cmaddox5 (approved):

PR #1090, opened 2025-01-14
fix(shuttles): Fix indexes changing to 0-based and displayed stop names changing when stops are reordered

rudiejd:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹🐛 Map view - stops index changing (from first stop as 0 or as 1](https://app.asana.com/0/584764604969369/1209034382499890/f)

Problem:
When re-ordering shuttle stops for a new shuttle, the indexing of stops would change from 1-based to 0-based after the first reordering. 

Solution:
Start using 1-based indexing for stops in the `stop_input` component. 

Problem:
When re-ordering shuttle stops for a new shuttle, stops with auto-completed names would switch from the auto completed value (e.g.  1 - Washington St opp Ruggles St) to just the value of the stop ID e.g. just 1

Solution:
On every re-render of stops, ensure that the value of the `stop_input` component is the human readable value. I fetch from the stop from the changeset if it's already there, otherwise I fetch the stop from the DB. 

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/components/stop_input.ex
+++ b/lib/arrow_web/components/stop_input.ex
@@ -32,8 +32,6 @@ defmodule ArrowWeb.StopInput do
         end
       )
 
-    IO.inspect(assigns.field)
-
     ~H"""
     <div id={@id} class={@class}>
       <.live_select
--- a/lib/arrow_web/components/stop_input.ex
+++ b/lib/arrow_web/components/stop_input.ex
@@ -38,9 +38,17 @@ defmodule ArrowWeb.StopInput do
         allow_clear={true}
         target={@myself}
         options={@options}
+        value_mapper={&stop_value_mapper/1}
       />
     </div>
     """
+  end
+
+  defp stop_value_mapper(text) do
+    case Shuttles.stop_or_gtfs_stop_for_stop_id(text) do
+      nil -> {text, text}
+      stop -> option_for_stop(stop)
+    end
   end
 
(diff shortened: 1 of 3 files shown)


lib/arrow_web/components/stop_input.ex
```diff
@@ -38,11 +38,27 @@ defmodule ArrowWeb.StopInput do
...
+      case Map.get(field.form.source.changes, :display_stop) do
+        %Arrow.Gtfs.Stop{} = stop -> stop
+        %Arrow.Shuttles.Stop{} = stop -> stop
+        _ -> Shuttles.stop_or_gtfs_stop_for_stop_id(text)
+      end
+
+    if stop == nil do
+      {text, text}
+    else
+      option_for_stop(stop)
```
meagharty, line 58: I'd love to have `:display_stop` / `stop_or_gtfs_stop_for_stop_id` consolidated in some way but I think this function hints at the (existing) issue where we were deal with the `changes` and the committed values across new and edit.

meagharty (approved):
thanks for fixing up both of these bugs!

lib/arrow_web/components/stop_input.ex
rudiejd: I agree - I feel like `display_stop` is probably unnecessary since the stop is probably loaded in the changeset by the time the user has entered it. In the interest of keeping this PR small and making iterative improvements, I'm going to hold off on that for now.

PR #1092, opened 2025-01-15
feat: Disruption Limits

cmaddox5:
#### Summary of changes
**Asana Ticket:** [🏹 Implement "Create New Disruption Limit" page](https://app.asana.com/0/584764604969369/1208885496650092/f)

Adds the ability to create/edit limits on a disruption.

#### Reviewer Checklist


lib/arrow/disruptions/disruption_v2.ex
```diff
@@ -7,20 +7,40 @@ defmodule Arrow.Disruptions.DisruptionV2 do
...
   import Ecto.Changeset
 
+  @type t :: %__MODULE__{
+          title: String.t() | nil,
+          mode: atom() | nil,
+          is_active: boolean(),
+          description: String.t() | nil,
+          inserted_at: DateTime.t() | nil,
+          updated_at: DateTime.t() | nil,
+          limits: [Arrow.Disruptions.Limit.t() | Ecto.Association.NotLoaded.t()]
```
lemald, line 17: **question:** should this be
```suggestion
          limits: [Arrow.Disruptions.Limit.t()] | Ecto.Association.NotLoaded.t()
```

instead, or am I just misunderstanding how types for Ecto stuff work?

priv/repo/migrations/20250114140020_create_limit_day_of_weeks.exs
```diff
@@ -0,0 +1,20 @@
+defmodule Arrow.Repo.Migrations.CreateLimitDayOfWeeks do
+  use Ecto.Migration
+
+  def change do
+    create table(:limit_day_of_weeks) do
+      add :is_active, :boolean
+      add :day_name, :string, null: false
```
lemald, line 7: **suggestion:** could we make this an enum or something to restrict it to only valid days of the week at the database level?

lib/arrow/disruptions/limit.ex
```diff
@@ -0,0 +1,89 @@
...
+  ]
+
+  @type t :: %__MODULE__{
+          start_date: Date.t() | nil,
+          end_date: Date.t() | nil,
+          disruption: DisruptionV2.t() | Ecto.Association.NotLoaded.t(),
+          route: Route.t() | Ecto.Association.NotLoaded.t(),
+          start_stop: Stop.t() | Ecto.Association.NotLoaded.t(),
+          end_stop: Stop.t() | Ecto.Association.NotLoaded.t(),
+          limit_day_of_weeks: [LimitDayOfWeek.t() | Ecto.Association.NotLoaded.t()]
```
lemald, line 28: **question:** see my other question in `disruption_v2.ex`:
```suggestion
          limit_day_of_weeks: [LimitDayOfWeek.t()] | Ecto.Association.NotLoaded.t()
```

lib/arrow/disruptions/disruption_v2.ex
cmaddox5: O wow, thank you. Will correct that.

lib/arrow/limits/limit_day_of_week.ex
```diff
@@ -0,0 +1,78 @@
+defmodule Arrow.Limits.LimitDayOfWeek do
+  @moduledoc "schema for a limit day of week for the db"
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Arrow.Disruptions.Limit
+
+  @type t :: %__MODULE__{
+          day_name: String.t(),
```
lemald, line 10: **suggestion:** see [my suggestion](https://github.com/mbta/arrow/pull/1092/files#r1918692942) about making this an enum (there on the DB end, but if you make it a DB enum, this could become an enum type, too).

lib/arrow/limits.ex
```diff
@@ -0,0 +1,202 @@
+defmodule Arrow.Limits do
+  @moduledoc """
+  The Limits context.
+  """
+
+  import Ecto.Query, warn: false
```
lemald, line 6: **question:** do you plan on using `Ecto.Query` in here eventually and that's why you're importing it with `warn: false`?

cmaddox5: This may have been leftover from the generator and I didn't notice it. Will remove.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.html.heex
```diff
@@ -11,7 +11,10 @@
 <.disruption_form
   id="disruption_v2_form"
   form={@form}
+  action={@form_action}
   disruption_v2={@disruption_v2}
   icon_paths={@icon_paths}
+  errors={@errors}
   adding_new_service?={@adding_new_service?}
+  show_limit_form?={@show_limit_form?}
```
lemald, line 19: **suggestion (non-blocking):** might I suggest either renaming `show_limit_form?` to `adding_new_limit?`, or updating `adding_new_service?` to `show_service_form?`? Just to make the naming of the two consistent, since I see them as fairly similar.

cmaddox5: Good idea. I think `show_*` might be more specific because the same form will be used for edits.

lib/arrow/limits/limit_day_of_week.ex
cmaddox5: I had some trouble making this schema work with `Ecto.Enum`. From some reading online, it seems like they clash with each other without some code to convert back to a string before inserting into the DB. I think the `validate_required` + `validate_inclusion` in `changeset/2` should mimic the same functionality.

rudiejd: personally my preferred approach would be to use `Ecto.Enum` in the code and avoid DB enums, which allows the code to be the source of truth for allowed values. This article is pretty old and is about MySQL but I think most of the concerns are still valid for DB enums in general: https://komlenic.com/244/8-reasons-why-mysqls-enum-data-type-is-evil/

cmaddox5: I think that article provides good tips for arbitrary enum values, but with it always containing day names I don't think we would encounter most of the issues. That said, I don't have a strong opinion so am down to implement what the team prefers.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
```diff
@@ -179,7 +193,204 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
...
+  attr :form, :any, required: true
+  attr :existing_limits, :any
+  attr :show_limit_form?, :boolean, required: true
+  attr :limit_id_in_form, :integer
+  attr :icon_paths, :map, required: true
+
+  defp limit_form_section(assigns) do
+    ~H"""
+    <h3>Limits</h3>
+    <%= if Ecto.assoc_loaded?(@existing_limits) do %>
```
lemald, line 205: **question:** what's the case where this isn't loaded? Is that for new disruptions or something? Or is it more for error handling?

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
```diff
@@ -179,7 +193,204 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
...
+      from s in Arrow.Gtfs.Stop,
+        where: s.location_type == :stop_platform,
+        join: st in Arrow.Gtfs.StopTime,
+        on: s.id == st.stop_id,
+        join: t in Arrow.Gtfs.Trip,
+        on: st.trip_id == t.id,
+        where: t.route_id == ^route_id,
+        select: s,
+        distinct: s.parent_station_id
+    )
```
lemald, line 390: **question:** have you looked at the performance of this query at all? I feel like it's likely to be reasonably quick with all of the right indexes (which I think we have), but since stop times is going to be a fairly big table I feel like it's worth at least measuring.

test/support/fixtures/gtfs_fixtures.ex
```diff
@@ -0,0 +1,98 @@
+defmodule Arrow.GtfsFixtures do
+  @moduledoc """
+  This module defines test helpers for creating
+  GTFS entities.
+  """
+
+  use ExMachina.Ecto, repo: Arrow.Repo
```
lemald, line 7: **question:** I'm a little confused - this `uses` `ExMachina` but as far as I can tell doesn't actually take advantage of any of the ExMachina functionality. Usually with ExMachina you can just define a `foo_factory` function and it creates clauses of the `build` and `insert` functions to allow you to generate either in-memory or in-database instances of `foo`, among other things.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
cmaddox5: It's for new disruptions. This doesn't feel like the most elegant way to handle that case, but I'm unsure what a better way might be.

lemald: Gotcha - not sure that I have any suggestions on a more elegant way but I'll think about it.

test/support/fixtures/gtfs_fixtures.ex
cmaddox5: Wow I was way off on my usage here. Given that there was already a factory for `Arrow.Gtfs.Stop`, I added these there instead.

lemald: Optional, but if you want to I think you could also move the limit-related fixtures to ExMachina.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
cmaddox5: This was something I meant to revisit and forgot all about. I played around with it a bit trying to find a useful index and our existing ones seem to be as good as it gets. Good news is that the largest execution time I found while running it was 60ms. One thing I did discover is that I am currently running this query twice each time the route value changes... I need to figure out a way to allow the two stop inputs to share the same options.

lemald: That sounds pretty decent performance-wise! I'll keep my eyes open for the change to do a single query as opposed to two.

test/support/fixtures/gtfs_fixtures.ex
lemald: FYI, I think something may have gone awry with a rebase, since this file had appeared deleted but is now back.

cmaddox5: Woops, thanks for the heads up. Should be fixed now.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
```diff
@@ -231,40 +446,104 @@ defmodule ArrowWeb.DisruptionV2ViewLive do
...
+  end
 
-    case save_result do
-      {:ok, _} ->
+  def handle_event(
+        "add_limit",
+        _,
+        %{assigns: %{disruption_v2: %DisruptionV2{id: nil}, form: form}} = socket
+      ) do
+    case Disruptions.create_disruption_v2(form.params) do
```
Whoops, line 533: On the subject of splitting things: I see that the `add_limit` event is saving the disruption. I think this means we can rely on the disruption existing anytime the add limit for is displayed. This seems like there's an opportunity to make the add limit form it's own [LiveComponent](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveComponent.html) with it's own events so that the concerns of creating a limit can be fully isolated. I'm worried about the complexity of this form when we have HASTUS limits + replacement service.

Whoops:
With the exception of the bug around cancelling all my comments are non-blocking :-)

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
cmaddox5: Interesting. This is a great idea and I wish I had thought about it initially. Was trying too hard to keep all this in the same file. Going to play around with this idea and see if I can improve things.

Whoops:
Another interesting bug: when editing a disruption, click add limit, then add replacement service. Cancel the replacement service, then click add again and the LiveView crashes.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.ex
Whoops: In case you don't already know, the main footgun with nesting LiveComponents is to make your events go to the child component instead of the parent is you need to add a [phx-target={@myself} attribute](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveComponent.html#module-events). Both components share a websocket, so without that to disambiguate the event, the parent gets it by default.

cmaddox5: Tried my hand at this and I think I like the way it turned out. Both of the bugs you mentioned seem to have been resolved in the change.

lib/arrow_web/components/limit_section.ex
```diff
@@ -0,0 +1,340 @@
...
+      <%= if Ecto.assoc_loaded?(@disruption.limits) and Enum.any?(@disruption.limits) do %>
+        <div class="mb-3">
+          <.table id="limits" rows={@disruption.limits}>
+            <:col :let={limit} label="route">
+              <span
+                class="m-icon m-icon-sm mr-1"
+                style={"background-image: url('#{get_limit_route_icon_url(limit, @icon_paths)}');"}
+              />
+            </:col>
+            <:col :let={limit} label="start stop">{limit.start_stop.name}</:col>
```
Whoops, line 33: Very nitty nitpick, I found it briefly confusing that this component takes a limit parameter, and this is a different limit. I think it's compounded by the magic iteration that `.table` is doing for us where it's not as obvious it's being shadowed. I do not actually have a better name.

cmaddox5: That's fair. Might be lazy, but would super short-hand `f` be alright here? Either that or `limit_row` maybe?

lib/arrow_web/components/limit_section.ex
```diff
@@ -0,0 +1,340 @@
...
+                data-confirm="Are you sure you want to delete this limit?"
+              >
+                <.icon name="hero-trash-solid" class="bg-primary" />
+              </.button>
+            </:action>
+          </.table>
+        </div>
+      <% end %>
+
+      <.link_button :if={!@show_form?} class="btn-link" phx-click="add_limit">
```
Whoops, line 66: nitpick: could these checked be based on `limit_form` being nil or not removing a bit of state?

lib/arrow_web/components/limit_section.ex
Whoops: I'd suggest `limit_row` to make it clear what it is. I'm also fine with `limit`. The magic just took me a minute to work through, but not enough I'd actually suggest messing with the implementation of `.table`.

Whoops:
Looks like a small bug, this limit was saved with the "all day" boxes checked, but when I edit it they are no longer checked. I also can't seem to save it in this state, but also don't get a validation message anywhere.
Screenshot 2025-01-22 at 12 31 08 PM

lib/arrow_web/components/limit_section.ex
```diff
@@ -0,0 +1,335 @@
...
+
+  def handle_event("create", %{"limit" => limit_params}, socket) do
+    case Limits.create_limit(limit_params) do
+      {:ok, _} ->
+        send(self(), :update_disruption)
+        send(self(), {:put_flash, :info, "Limit created successfully"})
+
+        {:noreply,
+         socket
+         |> assign(limit: nil)
```
Whoops, line 236: I'm not sure why, this sure seems like it should close it, but when I save a new limit, the form stays open for me (this was true before getting rid of `show_form?` too.

cmaddox5: Ugh, that's no good. The edit/duplicate/delete buttons are still disabled too. I wonder if this has something to do with awkward timing between the sent messages to the parent plus assign changes here?

As I was typing this I confirmed it. So the `:update_disruption` message is causing issues for some reason. When I comment it out, it correctly closes the form. Will need to investigate.

lib/arrow_web/components/limit_section.ex
```diff
@@ -0,0 +1,348 @@
...
+                    {format_day_name(input_value(f_day_of_week, :day_name))}
+                  </div>
+                </div>
+                <div class="col col-lg-3">
+                  <input
+                    :if={normalize_value("checkbox", input_value(f_day_of_week, :active?))}
+                    class="form-control"
+                    value={input_value(f_day_of_week, :start_time)}
+                    name={input_name(f_day_of_week, :start_time)}
+                    type="time"
```
Whoops, line 150: I think using time inputs here are going to be problematic. As far as GTFS time is concerned 01:00 is different from 25:00 even though both represent 1:00 AM (and most of our disruptions end at 24+). We could probably find a way to disambiguate on the backend, but I think making it an input maybe validated with a regex would be easier. It also makes it possible to validate the end time is after the start time.

priv/repo/structure.sql
```diff
@@ -713,6 +713,78 @@ CREATE TABLE public.gtfs_trips (
...
 
+--
+-- Name: limit_day_of_weeks; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.limit_day_of_weeks (
+    id bigint NOT NULL,
+    is_active boolean,
+    day_name integer NOT NULL,
+    start_time time(0) without time zone,
```
Whoops, line 724: Same time issue here. I'm going out on a limb and guessing that Postgre's `time` type isn't going to let me but in 25:00.

Whoops:
I think this looks really good! Outside of the time issue and the form not closing, this looks good to me!

lib/arrow_web/components/limit_section.ex
Whoops: I think the call to `:update_disruption` could be left out entirely. The purpose of the `save` button also saving the disruption in the ticket was to ensure that the disruption existed before saving any Limits or ReplacementService. Now that's being accomplished by saving when the "Add Limit" button is clicked.

cmaddox5: The main function of the call is actually to update the disruption saved in the parent's state for the sake of updating the table with the change to its limits. That said, there's got to be a better way to accomplish that so I will look into it.

cmaddox5: Oof, just fixed it and that is the problem 🤦. Form closes as expected as long as I set values to the correct keys. Thank you.

Whoops: I think you could query for the limits inside this component instead of relying on the parent to pass them in couldn't you? (Non-blocking commentary: I have no issue with the current approach)

lib/arrow_web/components/limit_section.ex
```diff
@@ -0,0 +1,348 @@
...
+                data-confirm="Are you sure you want to delete this limit?"
+              >
+                <.icon name="hero-trash-solid" class="bg-primary" />
+              </.button>
+            </:action>
+          </.table>
+        </div>
+      <% end %>
+
+      <.link_button :if={is_nil(@limit_form)} class="btn-link" phx-click="add_limit">
```
Whoops, line 66: I was about to ask a dumb question about why this targets the outer component instead of itself, but I think I answered it: so it can trigger a save of the disruption.

lib/arrow_web/components/limit_section.ex
cmaddox5: Hmm, didn't think about that. That shouldn't be a huge change so I can work on that now.

lib/arrow_web/live/disruption_v2_live/disruption_v2_view_live.html.heex
```diff
@@ -11,7 +11,10 @@
 <.disruption_form
   id="disruption_v2_form"
   form={@form}
+  action={@form_action}
   disruption_v2={@disruption_v2}
   icon_paths={@icon_paths}
-  adding_new_service?={@adding_new_service?}
+  errors={@errors}
+  show_service_form?={@show_service_form?}
```
meagharty, line 18: Noting this because I expect it'll cause a merge conflict with my work in progress:

This change makes sense, but why is this `show_service_form?` but the other is `limit_in_form` (instead of `show_limit_form?` or at least `limit_in_form?`) What's the difference behavior wise?

cmaddox5: I had trouble with the name so it may be a little misleading. `limit_in_form` actually holds a `Limit` struct instead of a boolean. It's really there just for `add_limit` to be able to communicate to the component that a new form should be started. `update/2` in `LimitSection` listens for changes and creates a new form when a non-nil limit comes in.

lib/arrow_web/components/limit_section.ex
cmaddox5: That would be easy enough for new limits, but edited limits is the main reason I took this approach. I felt it was easier to reload the disruption than to find and replace an updated limit in the disruption.

lib/arrow_web/components/limit_section.ex
cmaddox5: Got that change pushed up. Also changed how we compare times in the changeset.

Whoops (approved):
:shipit:

PR #1093, opened 2025-01-15
feat(disruptionsv2): show shuttle map when adding replacement service…

rudiejd:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹 Implement "Create Replacement Service Component" map view](https://app.asana.com/0/584764604969369/1209053528462959/f)

Problem:
You can't easily visualize the shuttle that you're choosing for replacement service in a disruption without navigating to the shuttles page

Solution:
Shows the shuttle map view from the "create new shuttle" map page inline in the new "create disruption" page.
image

This corresponds to the mockup for the new disruption page on [miro](https://miro.com/app/board/uXjVKVLTOI0=/):
image

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/shape_html.ex
+++ b/lib/arrow_web/controllers/shape_html.ex
@@ -1,6 +1,10 @@
 defmodule ArrowWeb.ShapeView do
   use ArrowWeb, :html
+  alias Arrow.Shuttles
   alias Arrow.Shuttles.ShapesUpload
+  alias Arrow.Shuttles.Route
+  alias Arrow.Shuttles.RouteStop
+  alias Arrow.Shuttles.Shape
 
   embed_templates "shape_html/*"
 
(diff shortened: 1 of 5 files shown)


lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -320,13 +320,13 @@ defmodule ArrowWeb.ShuttleViewLive do
     }
   end
 
-  defp routes_to_layers(routes, existing_props) do
+  def routes_to_layers(routes, existing_props) do
```
cmaddox5, line 323: I'm wondering if it would be better to put these functions in a separate file instead of one view references functions of another. What do you think?

rudiejd: yep, I was of the same mindset. but where do you think they should be? maybe a new module within `ArrowWeb`? I was going to put them in `Arrow.Shuttles` before, but that seemed improper since this function is so coupled to the display layer

cmaddox5: My first reaction was something like `ArrowWeb.ShapeView`. On line 311, we are already calling a function from there so it could make sense to put these there as well.

rudiejd: that works for me 👍

lib/arrow_web/controllers/shape_html.ex
```diff
@@ -1,5 +1,9 @@
 defmodule ArrowWeb.ShapeView do
   use ArrowWeb, :html
+  alias Arrow.Shuttles
+  alias Arrow.Shuttles.Route
+  alias Arrow.Shuttles.RouteStop
+  alias Arrow.Shuttles.Shape
   alias Arrow.Shuttles.ShapesUpload
```
cmaddox5, line 7: suggestion: These could be grouped
```suggestion
  alias Arrow.Shuttles.{Route, RouteStop, Shape, ShapesUpload}
```

cmaddox5 (approved):
Looks good! Just one small comment.

PR #1095, opened 2025-01-16
feat(shuttles): Make the direction description a drop-down menu instead of free text

rudiejd:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹 Change direction_desc to drop-down (in Shuttle Definitions)](https://app.asana.com/0/584764604969369/1209163082064165/f)

Problem:
Direction description on the shuttle form was a free text field, but in reality there are a fixed number of potential descriptions [for a direction.](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#directionstxt)

Solution:
Make direction description a drop-down menu instead of a free-text box. 

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/shuttles/route.ex
+++ b/lib/arrow/shuttles/route.ex
@@ -7,7 +7,7 @@ defmodule Arrow.Shuttles.Route do
     field :suffix, :string
     field :destination, :string
     field :direction_id, Ecto.Enum, values: [:"0", :"1"]
-    field :direction_desc, :string
+    field :direction_desc, Ecto.Enum, values: [:Inbound, :Outbound, :North, :South, :East, :West]
     field :waypoint, :string
     belongs_to :shuttle, Arrow.Shuttles.Shuttle
     belongs_to :shape, Arrow.Shuttles.Shape
--- a/lib/arrow/shuttles/route.ex
+++ b/lib/arrow/shuttles/route.ex
@@ -3,11 +3,15 @@ defmodule Arrow.Shuttles.Route do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @direction_desc_values [:Inbound, :Outbound, :North, :South, :East, :West]
+
+  def direction_desc_values, do: @direction_desc_values
+
   schema "shuttle_routes" do
     field :suffix, :string
     field :destination, :string
     field :direction_id, Ecto.Enum, values: [:"0", :"1"]
-    field :direction_desc, Ecto.Enum, values: [:Inbound, :Outbound, :North, :South, :East, :West]
+    field :direction_desc, Ecto.Enum, values: @direction_desc_values
     field :waypoint, :string
     belongs_to :shuttle, Arrow.Shuttles.Shuttle
     belongs_to :shape, Arrow.Shuttles.Shape
--- a/lib/arrow_web/live/shuttle_live/shuttle_live.ex
+++ b/lib/arrow_web/live/shuttle_live/shuttle_live.ex
@@ -87,6 +87,7 @@ defmodule ArrowWeb.ShuttleViewLive do
                   type="select"
                   label="Direction Description"
                   prompt="Choose a value"
+                  phx-change="direction_desc_changed"
                   options={Ecto.Enum.values(Arrow.Shuttles.Route, :direction_desc)}
                 />
               </div>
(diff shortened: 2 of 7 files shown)


lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -82,7 +82,14 @@ defmodule ArrowWeb.ShuttleViewLive do
                 <.input field={f_route[:direction_id]} type="text" label="Direction id" />
               </div>
               <div class="col">
-                <.input field={f_route[:direction_desc]} type="text" label="Direction desc" />
+                <.input
+                  field={f_route[:direction_desc]}
+                  type="select"
+                  label="Direction Description"
+                  prompt="Choose a value"
+                  phx-change="direction_desc_changed"
```
meagharty, line 90:
```suggestion
```

You should be able to remove this and the related `handle_event` calls functions. You don't need to have this input handle it's own change -- I double-checked that this worked locally (just in case there was an issue with the `@form` level handling of the `f_route`). I don't see another reason or edge case as to why it would need to be handled here.

lib/arrow/shuttles/route.ex
```diff
@@ -7,7 +7,7 @@ defmodule Arrow.Shuttles.Route do
     field :suffix, :string
     field :destination, :string
     field :direction_id, Ecto.Enum, values: [:"0", :"1"]
-    field :direction_desc, :string
+    field :direction_desc, Ecto.Enum, values: [:Inbound, :Outbound, :North, :South, :East, :West]
```
meagharty, line 10: Do we need to add a migration or some other mitigation to drop any existing routes? I think this will be a breaking change in dev/prod when trying to render test data.

meagharty (commented):
@jzimbel-mbta I jumped in here because I was curious about inputs_for/`f_route` change propagation.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
```diff
@@ -451,6 +475,31 @@ defmodule ArrowWeb.ShuttleViewLive do
     end
   end
 
+  defp update_second_direction(socket, first_direction) do
```
meagharty, line 478: Ah okay I see why there's custom change handling here.

Not a designer but this side effect feels surprising to me -- generally I'm not a fan of side effects in UIs. It doesn't reverse itself either (as in, if I change the second, the first doesn't change). 

I could see us validating somewhere (like the Shuttle.changeset fn) that the direction pairs line up though. And/or adding to `def mount(%() ...)` (~line 299) to set default Inbound/Outbound.

lib/arrow/shuttles/route.ex
rudiejd: These are actually the only values in use in prod/dev. And switching this to an Ecto.Enum should still have it serialized as a varchar. I can make a migration just to be extra safe though.

lib/arrow_web/live/shuttle_live/shuttle_live.ex
rudiejd: yeah, definitely not a designer either 😅

My idea was that I still wanted to allow people to change the second direction if they don't match up, because I think technically it is valid to have `Inbound` / `North` for example. I can get rid of this if that is more annoying than helpful though

lib/arrow/shuttles/route.ex
meagharty: "update the values in prod/dev" was a valid mitigation to me so if that's already the case then great 👍

lib/arrow_web/live/shuttle_live/shuttle_live.ex
meagharty: I'm not sure it's worth the extra complexity and possible user confusion on the edit page especially, like if the user doesn't notice/realize the other input changed.

We could ask Benji in the arrow slack channel

rudiejd: I can just remove it for now since it's not explicitly part of the AC, and we can always add it back in if it's desired later

test/support/fixtures/shuttles_fixtures.ex
```diff
@@ -107,7 +107,7 @@ defmodule Arrow.ShuttlesFixtures do
         shape: shape1,
         destination: "Harvard",
         direction_id: :"0",
-        direction_desc: "Southbound",
+        direction_desc: :South,
```
meagharty, line 110: I'm not sure we're using it now that the form is complete, but this should be updated in `priv/repo/seeds.exs` as well. We could instead remove that file since we no longer need it for development reasons.

meagharty (changes requested):

priv/repo/migrations/20250121165457_drop_shuttles_invalid_direction_descriptions.exs
```diff
@@ -0,0 +1,20 @@
...
+
+  defp get_valid_direction_sql_strings() do
+    Arrow.Shuttles.Route.direction_desc_values() |> Enum.map(&"'#{&1}'") |> Enum.join(",")
+  end
+
+  def change do
+    execute("""
+      DELETE FROM shuttle_route_stops 
+      WHERE shuttle_route_id IN 
+        (SELECT shuttle_route_id FROM shuttle_routes WHERE direction_desc NOT IN (#{get_valid_direction_sql_strings()}));
```
meagharty, line 12: This works. A downside to using SQL is the migration might fail in the future if we make incompatible changes to the schema and it's not changed here as well.

This seems easiest (especially since we don't have `on_delete: :delete_all` for the parent `shuttles` tables), but we should have an `up` and a `down` to support rollback and prevent Ecto.Migration errors. I don't expect Ecto to be able to reverse this on it's own (as it's raw SQL). Could also pass two commands to `execute` (one for up and one for down): https://hexdocs.pm/ecto_sql/Ecto.Migration.html#module-change

To @jzimbel-mbta's point in Slack, an empty block for the down function should work for this

meagharty (approved):

PR #1097, opened 2025-01-22
tweak: Show edit page after saving disruption

cmaddox5:
#### Summary of changes
**Asana Ticket:** ad-hoc

After saving a disruption, you remain at `/new` which means you can't edit a disruption after immediately after creating it. If we move these routes into a `live_session`, we can navigate the user without reloading the page. 

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/router.ex
+++ b/lib/arrow_web/router.ex
@@ -51,8 +51,10 @@ defmodule ArrowWeb.Router do
     put("/disruptions/:id/row_status", DisruptionController, :update_row_status)
     post("/disruptions/:id/notes", NoteController, :create)
 
-    live("/disruptionsv2/new", DisruptionV2ViewLive, :new)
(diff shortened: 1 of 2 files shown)


lib/arrow_web/router.ex
```diff
@@ -51,8 +51,10 @@ defmodule ArrowWeb.Router do
     put("/disruptions/:id/row_status", DisruptionController, :update_row_status)
     post("/disruptions/:id/notes", NoteController, :create)
 
-    live("/disruptionsv2/new", DisruptionV2ViewLive, :new)
-    live("/disruptionsv2/:id/edit", DisruptionV2ViewLive, :edit)
+    live_session(:default) do
```
meagharty, line 54: If we're using `push_patch` to update the current liveview, not "navigate", do we need this live_session setup? not sure we do [according to docs](https://hexdocs.pm/phoenix_live_view/live-navigation.html)

meagharty (approved):

PR #1101, opened 2025-01-23
feat: add replacement_service schema

meagharty:
#### Summary of changes
**Asana Ticket:** Part of [🏹 Implement "Create Replacement Service Component" - Shuttle Activation Upload/Parsing + Save Component Button](https://app.asana.com/0/584764604969369/1208991517812079)

Adds the database schema and related modules for ReplacementService

I may wait to merge this until the form is complete but opening this up for  👀 anyway

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/disruptions.ex
+++ b/lib/arrow/disruptions.ex
@@ -106,4 +106,100 @@ defmodule Arrow.Disruptions do
   def change_disruption_v2(%DisruptionV2{} = disruption_v2, attrs \\ %{}) do
     DisruptionV2.changeset(disruption_v2, attrs)
   end
+
+  alias Arrow.Disruptions.ReplacementService
+
+  @doc """
+  Returns the list of replacement_services.
+
+  ## Examples
+
+      iex> list_replacement_services()
+      [%ReplacementService{}, ...]
+
+  """
+  def list_replacement_services do
+    Repo.all(ReplacementService)
+  end
+
+  @doc """
+  Gets a single replacement_service.
+
+  Raises `Ecto.NoResultsError` if the Replacement service does not exist.
+
+  ## Examples
+
+      iex> get_replacement_service!(123)
+      %ReplacementService{}
+
+      iex> get_replacement_service!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_replacement_service!(id), do: Repo.get!(ReplacementService, id)
+
+  @doc """
+  Creates a replacement_service.
+
+  ## Examples
+
+      iex> create_replacement_service(%{field: value})
+      {:ok, %ReplacementService{}}
+
+      iex> create_replacement_service(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def create_replacement_service(attrs \\ %{}) do
+    %ReplacementService{}
+    |> ReplacementService.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  @doc """
+  Updates a replacement_service.
+
+  ## Examples
+
+      iex> update_replacement_service(replacement_service, %{field: new_value})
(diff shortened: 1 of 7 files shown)


lib/arrow/disruptions/replacement_service.ex
```diff
@@ -0,0 +1,72 @@
...
+  import Ecto.Changeset
+
+  alias Arrow.Disruptions.DisruptionV2
+  alias Arrow.Shuttles.Shuttle
+
+  @type t :: %__MODULE__{
+          reason: String.t() | nil,
+          start_date: Date.t() | nil,
+          end_date: Date.t() | nil,
+          source_workbook_data: map(),
```
cmaddox5, line 17: I had an idea that I'm not sure is any good but want to bring it up anyway. Instead of saving the data to the DB as JSON, would it be better to parse that data and add them as rows to a new `Timetable` table? I know the parsing needs to happen regardless, but with this we wouldn't have to parse it every time we go to view it. What do you think?

lib/arrow/disruptions.ex
```diff
@@ -101,4 +103,100 @@ defmodule Arrow.Disruptions do
   def change_disruption_v2(%DisruptionV2{} = disruption_v2, attrs \\ %{}) do
     DisruptionV2.changeset(disruption_v2, attrs)
   end
+
+  alias Arrow.Disruptions.ReplacementService
```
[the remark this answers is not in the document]
meagharty: I didn't think that ReplacementService desired its own context since I don't think it exists outside of a Disruption.

The warning message from Phoenix generators always makes me second-guess this piece though.

> As everything else, contexts in Phoenix are modules, but with the distinct reponsibility of drawing boundaries and grouping functionality. In other words, they allow us to reason and discuss about application design.

>Sometimes it may be tricky to determine if two resources belong to the same context or not. In those cases, prefer distinct contexts per resource and refactor later if necessary. Otherwise you can easily end up with large contexts of loosely related entities. Also keep in mind that the fact two resources are related does not necessarily mean they belong to the same context, otherwise you would quickly end up with one large context, as the majority of resources in an application are connected to each other. To sum it up: if you are unsure, you should prefer separate modules (contexts).
https://hexdocs.pm/phoenix/contexts.html#in-context-relationships

I suppose I am generally less concerned about file size than others (not to say that other Contexts in the repo aren't candidates for trimming 🙃 ).

I can move this out. It feels like we'll always end up with separate Contexts though if we follow this pattern?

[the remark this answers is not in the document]
cmaddox5: You make a good point and if anything, I should work on moving the limits functions into this file as well. Makes sense for limits and replacement service to follow the same pattern. So I like this and will work on moving limits.

lib/arrow/disruptions/replacement_service.ex
meagharty: We talked about this in slack the other day, where there doesn't seem to be a lot of value in normalizing the data at this point. The data required for a timetable is currently separate (and can be separately updated!), like the shuttle stop data.

We could always iterate on this later if we see the need, depending on when the computation of the trips/timetable for GTFS occurs.

I'd guess we are better off having a sheet representation and the computed timetable representation. There could be changes in the sheet representation from changes made to the spreadsheet. It's not that hard to perform updates or mutations on json objects in postgres (or in elixir if that's easier, since realistically we won't have that many sheets/rows in the database).

cmaddox5: Makes sense. I don't feel super strongly about it. Just wanted to throw the idea out there just in case.

cmaddox5 (approved):
This looks good to me! I see there are some merge conflicts, but I'll go ahead and approve now. Happy to take another look if the conflicts are gnarly.

PR #1110, opened 2025-02-04
fix(disruptionsv2): disruption limit form is closed after limit is created

rudiejd:
#### Summary of changes
**Asana Ticket:** [🏹🐛 Disruption limit component create form doesn't close after save](https://app.asana.com/0/584764604969369/1209292019824563/f)

Problem:
The limit form was not closing after a new limit was created.

Solution:
Set the correct value (`@limit_in_form`) to `nil` such that the limit form is no longer displayed after a new disruption limit gets created. Add unit test covering this case

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/components/limit_section.ex
+++ b/lib/arrow_web/components/limit_section.ex
@@ -70,8 +70,12 @@ defmodule ArrowWeb.LimitSection do
         </div>
       <% end %>
 
-
-      <.link_button :if={is_nil(@limit_form)} class="btn-link" phx-click="add_limit" id="add-limit-component">
+      <.link_button
+        :if={is_nil(@limit_form)}
+        class="btn-link"
+        phx-click="add_limit"
+        id="add-limit-component"
+      >
         <.icon name="hero-plus" /> <span>add limit component</span>
       </.link_button>
 
--- a/lib/arrow_web/components/limit_section.ex
+++ b/lib/arrow_web/components/limit_section.ex
@@ -96,6 +96,7 @@ defmodule ArrowWeb.LimitSection do
             <div class="col-lg-3">
               <.input
                 class="h-100"
+                id="select-route-id"
                 field={@limit_form[:route_id]}
                 type="select"
                 label="route"
--- a/lib/arrow_web/components/limit_section.ex
+++ b/lib/arrow_web/components/limit_section.ex
@@ -70,7 +70,8 @@ defmodule ArrowWeb.LimitSection do
         </div>
       <% end %>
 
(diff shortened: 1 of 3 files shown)


test/arrow_web/live/disruption_v2_live_test.exs
```diff
@@ -133,5 +135,23 @@ defmodule ArrowWeb.DisruptionV2LiveTest do
...
+    } do
+      {:ok, live, _html} = live(conn, ~p"/disruptionsv2/#{disruption.id}/edit")
+
+      update_response =
+        DisruptionV2ViewLive.handle_info(:update_disruption, %{
+          __changed__: %{},
+          assigns: %{disruption_v2: disruption, limit_in_form: Arrow.Disruptions.Limit.new()}
+        })
+
+      {:noreply, %{limit_in_form: nil}} = update_response
```
lemald, line 154: **suggestion (non-blocking):** I know testing philosophies will differ, but it might also be good to think about how we could test this by actually simulating the relevant form events via clicks / typing and then make an assertion about the resulting HTML. That said, I know that LiveView doesn't always make that as easy as (say) Testing Library would, so no pressure to rewrite this test if it would be overly onerous.

lemald: Oh hah you already had a whole discussion about how to test in Slack that I didn't read, ignore this then.

rudiejd: right, I tried my damnedest to write this test as a series of interactions with the live view. However, using `ex_machina` to build and insert mock GTFS entities in postgres proved to be difficult if not impossible given the bidirectional foreign key constraint between `gtfs_trips` and `gtfs_route_patterns`. I looked into disabling the foreign key constraints, but that seems difficult to do within a single test and would most likely have to go into `seeds.exs`. For future tests, I'm leaning towards fully importing  all or part of GTFS as a setup step and simply referencing real routes.

lib/arrow_web/components/limit_section.ex
```diff
@@ -70,7 +70,12 @@ defmodule ArrowWeb.LimitSection do
         </div>
       <% end %>
 
-      <.link_button :if={is_nil(@limit_form)} class="btn-link" phx-click="add_limit">
+      <.link_button
+        :if={is_nil(@limit_form)}
+        class="btn-link"
+        phx-click="add_limit"
+        id="add-limit-component"
```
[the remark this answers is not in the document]
rudiejd: This was part of my effort to test this end-to-end by interacting with the LiveView. I was formerly using these IDs in element selectors in the test.

rudiejd:
Confirmed the bug is no longer present on `dev-green`. Going to merge this

PR #1114, opened 2025-02-06
Function to generate stop times for replacement service

lemald:
#### Summary of changes
**Asana Ticket:** [🏹 Implement Shuttle Activation Timetable View](https://app.asana.com/0/584764604969369/1208992890474694)

This doesn't yet implement the UI, but since I approached it in a vaguely TDD manner and it therefore has tests I could put it up for review. This doesn't attempt to be a bug-for-bug reimplementation of the gtfs_creator algorithm, but it does follow the same basic approach for computing stop times. Worth keeping in mind that this will also have to support the API endpoint, but that doesn't mean that we can't make minor tweaks or adjustments down the line as long as the broad strokes are correct.

#### Reviewer Checklist


test/arrow/disruptions/replacement_service_test.exs
```diff
@@ -0,0 +1,91 @@
...
+      assert Enum.each(direction_1_trips, fn %{stop_times: stop_times} ->
+               length(stop_times) == 4
+             end)
+
+      assert direction_0_trips
+             |> Enum.filter(
+               &match?(
+                 %{
+                   stop_times: [
+                     %{stop_id: _, stop_time: "05:10"},
```
meagharty, line 31: I saw you mentioned it's not a bug-for-bug re-implementation of the gtfs_creator algorithm, which makes sense. Did you verify these stop time values are about what gtfs_creator produces from the shuttle spreadsheets now?

Algorithm seems to be good to me but I didn't verify these values.

meagharty (approved):
a few non-blocking comments

test/arrow/disruptions/replacement_service_test.exs
lemald: The values aren't exactly copied over, but I did reference [the tests for the algorithm in gtfs_creator](https://github.com/mbta/gtfs_creator/blob/ab5aac52561027aa13888e4c4067a8de177659f6/tests/disruptions/test_activated_shuttles.py) when designing my test data and deciding what the desired outcome should be.

test/arrow/disruptions/replacement_service_test.exs
```diff
@@ -0,0 +1,91 @@
+defmodule Arrow.Disruptions.ReplacementServiceTest do
+  use Arrow.DataCase
+  import Arrow.Factory
+
+  alias Arrow.Disruptions.ReplacementService
+
+  describe "trips_with_times/2" do
```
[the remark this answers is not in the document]
lemald: Actually, this is probably going to end up called from both a controller or LiveView for the timetable view, plus an API controller, so it would probably make sense to move over to the context.

PR #1116, opened 2025-02-07
feat(api): Add API for getting active replacement services

rudiejd:
#### Summary of changes
**Asana Ticket:** [🏹 Implement /replacement_service API endpoint](https://app.asana.com/0/584764604969369/1208578587556358/f)

Problem:
gtfs_creator needs to be able to know what alternative service options are available for all active disruptions so that it can create new trips and other GTFS entities

Solution:
Add `replacement_service` API that gtfs_creator can read to get all service that provided as a result of disruptions

#### Reviewer Checklist


lib/arrow/disruptions/replacement_service.ex
```diff
@@ -21,6 +21,12 @@ defmodule Arrow.Disruptions.ReplacementService do
           shuttle: Shuttle.t() | Ecto.Association.NotLoaded.t()
         }
 
+  @service_type_to_workbook_abbreviation %{
+    :weekday => "WKDY",
```
rudiejd, line 25: these felt like magic values to me, but I didn't want to mess with the parsing code too much so I just made `service_types` externally exposed as atoms.

lib/arrow_web/controllers/api/replacement_service_controller.ex
```diff
@@ -0,0 +1,53 @@
...
+         :ok <- Util.validate_date_order(start_date, end_date) do
+      data =
+        from(r in ReplacementService,
+          # where: r.start_date >= ^start_date and r.end_date <= ^end_date,
+          join: s in assoc(r, :shuttle),
+          join: d in assoc(r, :disruption),
+          join: sr in assoc(s, :routes),
+          join: rs in assoc(sr, :route_stops),
+          left_join: gs in assoc(rs, :gtfs_stop),
+          left_join: st in assoc(rs, :stop),
```
Whoops, line 23: Human whining about an interface meant for computers: I don't think we need the GTFS and Arrow stops themselves. We really only need the `stop_id`. GTFS stops are already in GTFS and GTFS Creator will already have any arrow managed stops by virtue of calling `/api/shuttle-stops`.

I also sort of hate how difficult JSON:API makes it for me as a human to make sense of the output, but I suppose that ship has sailed.

Whoops:
> This doesn't appear to meet the requirements in the ticket. Unless I'm missing something this is encoding the database structures rather than generating the timetables as described.

Or I take that back, I think it is trying to. But my test shuttle is returning null timetables for all days.

lib/arrow/disruptions/replacement_service.ex
```diff
@@ -74,4 +80,154 @@ defmodule Arrow.Disruptions.ReplacementService do
...
+  @spec schedule_service_types :: list(atom())
+  def schedule_service_types, do: [:weekday, :saturday, :sunday]
+
+  def trips_with_times(
+        %__MODULE__{source_workbook_data: workbook_data} = replacement_service,
+        service_type_atom
+      ) do
+    service_type_abbreviation = Map.get(@service_type_to_workbook_abbreviation, service_type_atom)
+
+    if Map.has_key?(workbook_data, service_type_abbreviation) do
```
Whoops, line 93: Chasing my lack of timetables backwards, it looks like this `if` will never be true. The keys on the workbook are `["SAT headways and runtimes", "SUN headways and runtimes",
 "WKDY headways and runtimes", "version"]`, but your trying to look up `SAT`/`SUN`/`WKDY`.

rudiejd: ahh good point! I missed adding the `headways and runtimes` part in

Whoops:
Attempting to test this locally, I'm unable to call this endpoint:
```
Request: GET /api/replacement-service?start_date=2025-01-01&end_date=2025-01-31
** (exit) an exception was raised:
    ** (BadMapError) expected a map, got: nil
        (elixir 1.18.1) lib/map.ex:535: Map.get(nil, "headway", nil)
        (arrow 0.1.0) lib/arrow/disruptions/replacement_service.ex:207: Arrow.Disruptions.ReplacementService.do_make_trip_start_times/4
        (arrow 0.1.0) lib/arrow/disruptions/replacement_service.ex:125: anonymous fn/5 in Arrow.Disruptions.ReplacementService.do_trips_with_times/2
        (elixir 1.18.1) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
        (elixir 1.18.1) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
        (arrow 0.1.0) lib/arrow/disruptions/replacement_service.ex:123: Arrow.Disruptions.ReplacementService.do_trips_with_times/2
        (arrow 0.1.0) lib/arrow_web/controllers/api/replacement_service_controller.ex:34: anonymous fn/2 in ArrowWeb.API.ReplacementServiceController.index/2
        (elixir 1.18.1) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
        (arrow 0.1.0) lib/arrow_web/controllers/api/replacement_service_controller.ex:33: anonymous fn/1 in ArrowWeb.API.ReplacementServiceController.index/2
        (elixir 1.18.1) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
        (arrow 0.1.0) lib/arrow_web/controllers/api/replacement_service_controller.ex:28: ArrowWeb.API.ReplacementServiceController.index/2
        (arrow 0.1.0) lib/arrow_web/controllers/api/replacement_service_controller.ex:1: ArrowWeb.API.ReplacementServiceController.action/2
        (arrow 0.1.0) lib/arrow_web/controllers/api/replacement_service_controller.ex:1: ArrowWeb.API.ReplacementServiceController.phoenix_controller_pipeline/2
        (phoenix 1.7.18) lib/phoenix/router.ex:484: Phoenix.Router.__call__/5
        (arrow 0.1.0) lib/arrow_web/endpoint.ex:1: ArrowWeb.Endpoint.plug_builder_call/2
        (arrow 0.1.0) lib/arrow_web/endpoint.ex:1: ArrowWeb.Endpoint."call (overridable 3)"/2
        (arrow 0.1.0) deps/plug/lib/plug/debugger.ex:136: ArrowWeb.Endpoint."call (overridable 4)"/2
        (arrow 0.1.0) lib/arrow_web/endpoint.ex:1: ArrowWeb.Endpoint.call/2
        (phoenix 1.7.18) lib/phoenix/endpoint/sync_code_reload_plug.ex:22: Phoenix.Endpoint.SyncCodeReloadPlug.do_call/4
        (plug_cowboy 2.7.2) lib/plug/cowboy/handler.ex:11: Plug.Cowboy.Handler.init/2
 ```

**Comment by Whoops** [comment]
Tracked this one down. My test sheet has a last trip time of `25:10` but only headways to `25:00`. So it's a bug that the parser accepted this. It's a common issue and it should have rejected it. I'll put a ticket in.

**Comment by Whoops** [comment]
I will however note that none of my timetables seem to have stop_ids:
```
"stop_times": [
                  {
                    "stop_id": null,
                    "stop_time": "05:01"
                  },
                  {
                    "stop_id": null,
                    "stop_time": "05:05"
                  },
                  {
                    "stop_id": null,
                    "stop_time": "05:09"
                  },
                  {
                    "stop_id": null,
                    "stop_time": "05:13"
                  },
                  {
                    "stop_id": null,
                    "stop_time": "05:17"
                  },
                  {
                    "stop_id": null,
                    "stop_time": "05:21"
                  },
                  {
                    "stop_id": null,
                    "stop_time": "05:25"
                  },
                  {
                    "stop_id": null,
                    "stop_time": "05:29"
                  }
                ]
              },
 ```

Whoops (changes requested):
When using my test shuttle & spreadsheet, Saturday/Sunday don't have timetables in the 1 direction. Only weekdays do.

Screenshot 2025-02-18 at 1 36 41 PM

Whoops:
I think we get to blame the parser again. On those sheets the last trip time is represented as `1:xx:xx` instead of `25:xx:xx`. I'll write up another bug.

Whoops:
Not blocking nitpick: the only stop information gtfs_creator really needs is the `stop_id` which is already present in the timetables, so including the `route_stops` and `stops` in the result is unnecessary.

config/dev.exs
```diff
@@ -76,7 +76,7 @@ config :arrow, dev_routes: true
 
 # Set prefix env for s3 uploads
 config :arrow,
-  shape_storage_enabled?: true,
+  shape_storage_enabled?: false,
```
Whoops, line 79: Why? With this disabled, most shape functionality won't work without re-enabling this.

lib/arrow/disruptions/replacement_service.ex
```diff
@@ -74,4 +81,162 @@ defmodule Arrow.Disruptions.ReplacementService do
...
+         %__MODULE__{source_workbook_data: source_workbook_data, shuttle: shuttle},
+         day_of_week
+       ) do
+    # to do: find a way to ensure that display_stop_id is always populate on every shuttle route stop
+    # regardless of from where the shuttle comes
+    # (e.g. if a shuttle comes from a join, it should still have display_stop_id populated)
+    shuttle = Shuttles.populate_display_stop_ids(shuttle)
+
+    day_of_week_data =
+      Map.get(source_workbook_data, workbook_column_from_day_of_week(day_of_week))
```
Whoops, line 111: nitpick: instead of checking if the map has the key up on line 94, then getting the data here, you could just go ahead and fetch the data in your `if` and pass it into the function

```elixir
if day_of_week_data = Map.get(source_workbook_data, workbook_column_from_day_of_week(day_of_week))
          do_trips_with_times(replacement_service, service_type_abbreviation, day_of_week_data)
```

Whoops: It looks like the `day_of_week` parameter is only used for this get, so it could replace that.

Whoops: As long as I'm offering up such trivialities, since you aren't providing a default value, you could also use `source_workbook_data[workbook_column_from_day_of_week(day_of_week)]` to shorten the code a touch.

config/dev.exs
rudiejd: I just disabled this for dev. I was unable to re-create a lot of shapes in my local environment with this feature enabled, since the shapes already existed on S3. I don't need to check in the change, though, so I can keep it set to `true`

Whoops: Yeah I'd prefer it to stay `true` only because otherwise I'll keep having to manually change it.

We should probably figure out a way to default all of us to using our username as the bucket prefix or something to avoid that problem though. I've run into it to. For now I've just been prefixing my test shapes with `WH-`

rudiejd:
> Not blocking nitpick: the only stop information gtfs_creator really needs is the `stop_id` which is already present in the timetables, so including the `route_stops` and `stops` in the result is unnecessary.

that makes sense. Those were getting included indirectly because I was including `Shuttle`, which has a `ShuttleRoute`, which has `RouteStop(s)`. I can remove them

lib/arrow/disruptions/replacement_service.ex
```diff
@@ -74,4 +81,162 @@ defmodule Arrow.Disruptions.ReplacementService do
...
+
+    {first_trips, last_trips, headway_periods} =
+      Enum.reduce(day_of_week_data, {%{}, %{}, %{}}, fn data,
```
Whoops, line 114: Again just nitpicking style, if you extracted this anonymous function to a private function I think it'd be a little easier to grok.

lib/arrow_web/controllers/api/replacement_service_controller.ex
```diff
@@ -0,0 +1,54 @@
...
+  alias ArrowWeb.API.Util
+  alias Plug.Conn
+
+  @spec index(Conn.t(), map()) :: Conn.t()
+  def index(conn, params) do
+    with {:ok, start_date} <- Util.parse_date(params["start_date"]),
+         {:ok, end_date} <- Util.parse_date(params["end_date"]),
+         :ok <- Util.validate_date_order(start_date, end_date) do
+      data =
+        from(r in ReplacementService,
```
Whoops, line 16: Nitpick: I can't find an explicit rule of thumb documented anywhere, so maybe it's only mine, but I do think this query should probably go in the context rather than the controller.

lib/arrow_web/controllers/api/replacement_service_controller.ex
```diff
@@ -0,0 +1,54 @@
...
+          join: sr in assoc(s, :routes),
+          join: rs in assoc(sr, :route_stops),
+          left_join: gs in assoc(rs, :gtfs_stop),
+          left_join: st in assoc(rs, :stop),
+          where: r.start_date <= ^end_date and r.end_date >= ^start_date,
+          preload: [:disruption, shuttle: {s, routes: {sr, route_stops: [:gtfs_stop, :stop]}}]
+        )
+        |> Repo.all()
+        |> Enum.map(fn replacement_service ->
+          Map.from_struct(replacement_service)
```
Whoops, line 29: More nitpicks: I'd again extract at least the outer, and probably both functions here into named helpers. My brain has a hard time with the 3 layers of nesting + the pipeline.

lib/arrow/disruptions/replacement_service.ex
rudiejd: I actually didn't write this part, I just moved it to a new function. But I agree, I'll move it to a private function

Whoops (approved):
All nitpicks are non-blocking style commentary, feel free to take or leave at your preference!

lib/arrow_web/controllers/api/replacement_service_controller.ex
rudiejd: I see this was also done for the `limits_controller`, so I can go ahead and move it for consistency

lib/arrow/disruptions/replacement_service.ex
rudiejd: I forget that `Access` returns nil if the key doesn't exist in this language sometimes

Whoops: I definitely didn't have to confirm on the REPL before suggesting that ;-)

PR #1121, opened 2025-02-11
feat: Scaffold v2 homepage

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [🏹 Implement Arrow v2 home/index page scaffolding and navigation](https://app.asana.com/0/584764604969369/1209170968037047)

This adds the skeleton of the new page at /disruptionsv2, including header, footer, navbar, and search input.

I'm finishing up tests but wanted to get this review posted in the meantime.
There's no especially complex or stateful logic in these changes, so it's pretty straightforward to check visually.

#### Reviewer Checklist


lib/arrow_web/components/core_components.ex
```diff
@@ -580,6 +585,47 @@ defmodule ArrowWeb.CoreComponents do
...
+  """
+  attr :page, :string, required: true
+  attr :create_disruption_permission?, :boolean, default: false
+
+  def navbar(assigns) do
+    pages = [
+      {~p"/disruptionsv2", "Disruptions"},
+      {~p"/shuttles", "Shuttle Definitions"},
+      {~p"/shapes", "Shuttle Shapes"},
+      {~p"/stops", "Shuttle Stops"}
```
jzimbel-mbta, line 602: One minor thing I'm unsure of and will check with the team and/or Benji on: what's our standard for button label capitalization?

Firas's mockups use all lowercase, but I noticed Meag going back and updating the labels to "Title Case" in some recent cleanup work. And on Screens team I was used to "Title case" where only the first word was capitalized.

lib/arrow_web/components/core_components.ex
```diff
@@ -580,6 +585,47 @@ defmodule ArrowWeb.CoreComponents do
...
+          <a class="btn btn-primary mr-1" href={~p"/disruptionsv2/new"}>+ Create New</a>
+        <% {"/disruptionsv2", false} -> %>
+          <a class="btn btn-primary mr-1" href={~p"/disruptionsv2"}>Disruptions</a>
+        <% _ -> %>
+      <% end %>
+      <a
+        :for={{page, label} <- @pages}
+        class={"btn mr-1 btn-" <> if(page == @page, do: "primary", else: "outline-secondary")}
+        href={page}
+      >
```
jzimbel-mbta, line 621: Annoyingly, tags generated with the `:for` construct have no whitespace between them, and whitespace matters between flow content elements. So if you look closely, there's a larger gap between any individually rendered links vs between the ones rendered iteratively.

I tried switching the links to use flexbox, where whitespace doesn't matter, but couldn't find the right bootstrap classes (if they exist?) to compress them to the start of the row. `justify-content-start` didn't seem to have the desired effect. I'd absolutely be open to any suggestions on this.

lib/arrow_web/components/core_components.ex
meagharty: Do Firas's mockups all use lowercase? Last I checked it was inconsistent...we should decide.

I think we were originally using all lowercase because that seemed to be the existing precedence in the codebase, similar to how Walton added bootstrap styling to the generated core_components.

lib/arrow_web/controllers/disruption_v2_html/index.html.heex
```diff
@@ -0,0 +1,15 @@
...
+  />
+  <%= form_tag(Controller.current_path(@conn), method: "get", class: "col-3") do %>
+    <div class="input-group">
+      <input type="text" name="search" class="form-control" placeholder="search" />
+
+      <div class="input-group-append">
+        <button type="submit" class="btn btn-outline-secondary" title="search">🔎</button>
+      </div>
+    </div>
+  <% end %>
```
jzimbel-mbta, line 14: The search input isn't hooked up to anything for now, since it's part of the suite of filter controls for the not-yet-implemented disruptions table.

lib/arrow_web/components/core_components.ex
jzimbel-mbta: Ah ok. And yeah I see now that it's a mix of cases in the mockups.

I'll start a little slack thread.

lib/arrow_web/components/core_components.ex
rudiejd: does it work differently if you do the `for` loop as part of a separate template block e.g. 
```heex
<%= for {page, label} <- @pages do %>
    ...
<% end %>
```

rudiejd: no problem! this might be worth raising on the phoenix repo - as far as I understand, `:for` comprehensions should behave the same way as `<% for ... %>`

rudiejd:
deploying this to `dev-green` to play around

rudiejd:
a couple of small UI bugs:

- if you double click on the disruption button quickly, you aren't taken to `disruptionsv2/new` - the `Create new` button seems to do nothing when you first navigate disruptions
- in general, repeatedly clicking on a nav button reloads the tab even if you're only navigated to the tab which you clicked. e.g. if you have `Shuttle stops` selected and you click on `Shuttle stops` the page reloads

jzimbel-mbta:
> * in general, repeatedly clicking on a nav button reloads the tab even if you're only navigated to the tab which you clicked. e.g. if you have `Shuttle stops` selected and you click on `Shuttle stops` the page reloads

I fixed this by conditionally removing the `href` attribute from the current page's link-button. I think this change makes sense but I do feel slight hesitation about the difference in behavior of the `.btn-primary` link when you're on /disruptionsv2 vs any other page with the navbar.

I wonder if I should also change the button style to make it clearer that it's non-interactive?

rudiejd:
redeploying to mess around again 🥼

rudiejd:
looks way better! one last thing: is there a way to prevent the page header/nav bar from re-rendering when you switch tabs? maybe I'm too used to react-router based navigation, but my assumption would be that only the content section should re-render



https://github.com/user-attachments/assets/4f422e73-785c-483e-a56f-1db43aba8fdc

jzimbel-mbta:
> is there a way to prevent the page header/nav bar from re-rendering when you switch tabs?

I don't think this is possible without forgoing simple HTML links (`<a>` elements) and separate pages for each "tab".

"Switching tabs" is really loading a whole new page, and I don't know of a way to persist part of a page's state to another page without a more complex JS-based approach, like react-router as you mentioned. Basically this is a new page load, not a re-render.

Since the T logo image seems to be the main source of flickering, I checked caching for it and it is indeed cached. There's still a very short delay before it appears on page load though, because the browser still needs to do a request to the server to make sure its cached version is still current.


If you feel strongly about this, we could discuss doing more of a SPA approach where switching tabs actually updates page state without actually having the browser navigate to a new document, but that would be a significant change from the current agreed-upon approach.

rudiejd (approved):
your explanation of why the flickering happens makes sense, and I don't have a super strong opposition to it, especially considering it seems non-trivial to fix and is only a UI bug not a semantic one. We could probably imitate react router by making the entire application a live view and intercepting the URL changes to change the live view state, similar to how react router works. I'm honestly surprised there's no react router analog for live view yet

PR #1128, opened 2025-02-21
Add navigation to edit button.

cmaddox5:
#### Summary of changes
**Asana Ticket:** ad-hoc

The original table task didn't mention adding navigation to the edit button, but I should have added it. Super quick change to allow editing a disruption from the row in the table.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/disruption_v2_html/_table.html.heex
+++ b/lib/arrow_web/controllers/disruption_v2_html/_table.html.heex
@@ -17,7 +17,7 @@
       <% {start_date, end_date} = get_dates(disruption) %>
       <tr>
         <td>
-          <a href={~p"/disruptionsv2/#{disruption.id}/edit"} class="btn-link btn-sm" type="button">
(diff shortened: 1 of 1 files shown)


lib/arrow_web/controllers/disruption_v2_html/_table.html.heex
```diff
@@ -17,9 +17,9 @@
       <% {start_date, end_date} = get_dates(disruption) %>
       <tr>
         <td>
-          <.button class="btn-link btn-sm" type="button">
+          <a href={~p"/disruptionsv2/#{disruption.id}/edit"} class="btn-link btn-sm" type="button">
```
lemald, line 20: **question:** does `type="button"` still make sense if this is an `<a>`? [Per MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#type) it looks like that's supposed to signify MIME type of the linked resource.

cmaddox5: Good catch. Got that removed.

PR #1132, opened 2025-02-26
tweak: Expose Shuttle shapes in API response

cmaddox5:
#### Summary of changes
**Asana Ticket:** ad-hoc

While working on the GTFS work for replacement services, I realized I didn't have access to route shapes for shuttles. Added an attribute to the API response to expose this. `gtfs_creator` turns kml file names into shape_ids so I did some formatting of the shape `path` with that in mind.

A couple of other tweaks in here:
1. Added back an auth plug that I think was unintentionally removed.
2. Reduced some nesting in the replacement service API response.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/router.ex
+++ b/lib/arrow_web/router.ex
@@ -86,7 +86,7 @@ defmodule ArrowWeb.Router do
   end
 
   scope "/api", as: :api, alias: ArrowWeb.API do
-    pipe_through([:redirect_prod_http, :json_api])
+    pipe_through([:redirect_prod_http, :json_api, :authenticate_api])
 
     get("/replacement-service", ReplacementServiceController, :index)
     get("/shuttles", ShuttleController, :index)
--- a/test/arrow_web/controllers/api/replacement_service_controller_test.exs
+++ b/test/arrow_web/controllers/api/replacement_service_controller_test.exs
@@ -68,15 +68,13 @@ defmodule ArrowWeb.API.ReplacementServiceControllerTest do
                        "sunday" => nil,
                        "weekday" => %{
                          "0" => [
-                           %{
-                             "stop_times" => [
-                               %{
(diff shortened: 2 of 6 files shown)


lib/arrow_web/router.ex
```diff
@@ -86,7 +86,7 @@ defmodule ArrowWeb.Router do
   end
 
   scope "/api", as: :api, alias: ArrowWeb.API do
-    pipe_through([:redirect_prod_http, :json_api])
```
rudiejd, line 89: i think i probably removed this for local testing 🤦

test/arrow_web/controllers/api/replacement_service_controller_test.exs
```diff
@@ -68,15 +68,13 @@ defmodule ArrowWeb.API.ReplacementServiceControllerTest do
                        "sunday" => nil,
                        "weekday" => %{
                          "0" => [
-                           %{
-                             "stop_times" => [
```
rudiejd, line 72: should this include shape ID now?

lib/arrow_web/router.ex
cmaddox5: I figured it was an accident. No biggie!

test/arrow_web/controllers/api/replacement_service_controller_test.exs
cmaddox5: Those will come from the shuttles endpoint. I can update that test to reflect that.

cmaddox5: Updated the test. Unfortunately, the value returned in the response is static in tests, but we can at least make sure the value is returned. Added a comment explaining the assertion.

rudiejd (approved):

PR #1135, opened 2025-02-26
🏹 Implement search box

Whoops:
#### Summary of changes
**Asana Ticket:** [🏹 Implement search box](https://app.asana.com/0/1204473309455397/1209170968037053)

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/disruption_v2_controller/filters.ex
+++ b/lib/arrow_web/controllers/disruption_v2_controller/filters.ex
@@ -41,10 +41,16 @@ defmodule ArrowWeb.DisruptionV2Controller.Filters do
   def from_params(params) when is_map(params) do
     view_mod = if(params["view"] == "calendar", do: Calendar, else: Table)
 
+    search =
+      if params["search"] in [nil, ""] or String.trim(params["search"]) == "",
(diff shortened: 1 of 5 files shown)


Whoops:
Putting this back in draft for now. It takes a pretty different approach than the filters work that was just merged, so I'll need to rework this.

Whoops:
Added @lemald as an additional reviewer since this interacts with his changes.

lemald:
Nitpick I noticed while playing around with this locally: if you enter "foo" into the search box and hit enter, then clear the text in the search box and hit enter again, the resulting path you get taken to is `/disruptionsv2?search=foo&search=`, which ultimately works and shows all disruptions, but is a little weird.

Whoops:
> `/disruptionsv2?search=foo&search=`

This should be fixed. The "search" parameter was being included in the hidden fields. Clearing it will still result in an empty `search=` parameter in the URL since the form is being submitted with a blank search, but I don't think that's fixable without javascript,  and we don't have the duplicate parameter anymore.

lemald (approved):
Tested locally and confirmed that the weird URL argument behavior with the search string is gone. 👍

PR #1146, opened 2025-03-07
feat: Support for uploading HASTUS Exports

cmaddox5:
#### Summary of changes
**Asana Ticket:** [🏹 Implement HASTUS import section + .zip parsing](https://app.asana.com/0/584764604969369/1209459244072058)

This PR adds the ability to upload HASTUS exports for an Arrow disruption. The upload is validated and parsed before showing either an error or a form with details of the export. Schemas have been set up, but there is nothing being saved to the DB yet. Once an upload is successfully parsed, the section on the page will appear blank. The form showing the details of the export will come in a later task.

#### Reviewer Checklist


lib/arrow/hastus/export_upload.ex
```diff
@@ -0,0 +1,176 @@
...
+  @doc """
+  Parses a HASTUS export and returns a list of data
+  Includes a rescue clause to catch errors while parsing user-provided data
+  """
+  @spec extract_data_from_upload(%{:path => binary()}) ::
+          {:ok, {:error, error_message} | {:ok, list(map())}} | rescued_exception_error()
+  def extract_data_from_upload(%{path: zip_path}) do
+    with {:ok, zip_file_data} <- File.read(zip_path),
+         {:ok, unzipped_file_list} <-
+           :zip.unzip(zip_file_data, [{:file_list, @filenames}, {:cwd, @tmp_dir}]),
```
lemald, line 38: **question:** if you always unzip to the same target directory, is there a risk of a race condition if two different users upload HASTUS exports at the same time on the same instance? I'm not sure how much of a real life risk that is, but it seems like something we could safeguard against by just putting a unique ID of some kind in the temp directory name.

cmaddox5: Wow, great point. I didn't think about that. That's a great idea.

lib/arrow/hastus/export_upload.ex
```diff
@@ -0,0 +1,176 @@
...
+    exported_stop_ids =
+      exported_stop_times
+      |> Enum.filter(&(&1["trip_id"] in revenue_trip_ids))
+      |> Enum.map(& &1["stop_id"])
+      |> Enum.uniq()
+
+    routes =
+      Arrow.Repo.all(
+        from st in Arrow.Gtfs.StopTime,
+          where: st.stop_id in ^exported_stop_ids,
```
lemald, line 106: **question:** since stop times is a big table, do we have an index on `st.stop_id`? I see a foreign key constraint but not an index, and I don't _think_ Postgres automatically creates an index when you have a FK constraint.

lib/arrow/hastus/export_upload.ex
```diff
@@ -0,0 +1,176 @@
...
+          join: t in Arrow.Gtfs.Trip,
+          on: t.id == st.trip_id,
+          select: t.route_id,
+          distinct: t.route_id
+      )
+
+    case routes do
+      [route] -> {:ok, route}
+      ["Green-B", "Green-C", "Green-D", "Green-E"] -> {:ok, "Green"}
```
lemald, line 115: **question:** if we try to insert an `export` record with a route of `"Green"` won't that cause a foreign key constraint error, since there's no GTFS route simply called `"Green"`? Or is there some further transformation on the route ID that I just haven't seen yet?

lib/arrow/hastus/export_upload.ex
cmaddox5: Yeah I think you're right. I'll get one added.

lib/arrow/hastus/export_upload.ex
cmaddox5: Ugh, I think you're right. This will be tricky then. My understanding is that GL will always have all branches in each export. Given that, a workaround could be populating the column with any of the 4 GL route IDs. But that seems hacky.

Whoops: While most exports will contain all lines, I wonder if we could ever have an export where one of the branches doesn't appear (e.x. we shut down Green-E)?

cmaddox5: I could probably get around that by checking that all items in the list start with `Green-`. Then we wouldn't need to worry about which branches are actually included at this point.

lemald:
**question:** the zip file ultimately needs to make it into S3, right? Is there a separate ticket for that part?

cmaddox5:
> **question:** the zip file ultimately needs to make it into S3, right? Is there a separate ticket for that part?

I thought that made more sense as part of the form submission task. At this point in the workflow, the user hasn't decided exactly what to do with the upload.

lemald:
> > **question:** the zip file ultimately needs to make it into S3, right? Is there a separate ticket for that part?
> 
> I thought that made more sense as part of the form submission task. At this point in the workflow, the user hasn't decided exactly what to do with the upload.

Fair enough! I just wanted to make sure I understood where it was captured.

test/support/factory.ex
```diff
@@ -147,6 +147,79 @@ defmodule Arrow.Factory do
...
+      tuesday: true,
+      wednesday: true,
+      thursday: true,
+      friday: true,
+      saturday: false,
+      sunday: false,
+      start_date: ~D[2025-01-01],
+      end_date: ~D[2026-01-01]
+    }
+    |> merge_attributes(attrs)
```
lemald, line 162: **nitpick:** if you don't need any specific logic around the attributes before merging them in, I think you can just define the function as taking no arguments instead of taking one argument and then skip the call to `merge_attributes/2`. See [this section of the ex_machina docs](https://hexdocs.pm/ex_machina/readme.html#full-control-of-factory).

lemald:
Okay, I think I've done a full pass and gotten all of my feedback in, so I'll just follow along with the open threads. until those are resolved.

lib/arrow/hastus/export_upload.ex
cmaddox5: So I actually really like the idea of using `line_id` as a FK instead of `route_id`. I feel it much better represents what we're going for in the GL scenario. And also makes the check for more than one route much more straightforward. If no one has any objections, I'm going to go that direction.

lib/arrow/hastus/export_upload.ex
cmaddox5: Made a decent improvement. GL export db time was `346.0ms`. Is now `97.7ms`.

lib/arrow/hastus/export_upload.ex
cmaddox5: I decided to add the `user_id` from the session to the tmp directory path. It took a little more work than I thought it would so maybe it wasn't the best solution.

lemald (approved):

PR #1143, opened 2025-03-05
feat(v2): Sort disruptions by start or end date

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [🏹 Implement sort by date functionality](https://app.asana.com/0/584764604969369/1209170968037055)

Based on the ticket's acceptance criteria, I needed to maintain the state of each sortable column. Otherwise, one of these behaviors would be necessary:
- toggling the sort direction for one column toggles it for all of them, or
- changing the active sorted column resets the previous sorted column to a default
  **(this is what the v1 disruptions table does)**

The new sorting filters look like:
```ex
%{
  ...
  sort: %{field_name => :asc | :desc},
  active_sort: field_name
}
```

which allows us to remember the sort direction of all columns, even when they are not active.

#### Reviewer Checklist
- [ ] Meets ticket's acceptance criteria
- [ ] Any new or changed functions have typespecs
- [ ] Tests were added for any new functionality (don't just rely on Codecov)
- [ ] This branch was deployed to the staging environment and is currently running with no unexpected increase in warnings, and no errors or crashes.

PR #1147, opened 2025-03-12
fix: Improve flaky integration tests

cmaddox5:
#### Summary of changes
**Asana Ticket:** ad-hoc

Our v1 disruption integration tests are pretty flaky and make it hard to know if CI is failing for a valid reason. I _think_ this is due to the test modules running in parallel. I removed those flags in hopes that this will improve. The runtime of the tests should not be impacted significantly by this due to the small number of tests.

#### Reviewer Checklist


rudiejd:
is there a way to verify that this change actually improves the flakiness?

cmaddox5:
> is there a way to verify that this change actually improves the flakiness? I can try repeatedly running until I get a failure, but curious if you had any other ideas

Unfortunately, I think that might be the only way. I have yet to reproduce this failure locally so the CI runner seems to be the only place a failure is likely to occur.

rudiejd:
Jon cherry-picked this as part of his PR, so I think we should be good to close

cmaddox5:
Merged in with #1143.

PR #1149, opened 2025-03-13
feat: HASTUS export form submission

cmaddox5:
#### Summary of changes
**Asana Ticket:** [🏹 Implement HASTUS import forms (post-parse)](https://app.asana.com/1/15492006741476/project/584764604969369/task/1209538099020504)

This PRs finishes the form submission that saves HASTUS exports to the DB and S3. Follows the same S3 uploading pattern as Shapes. There is not currently a way to view existing exports so easiest way to test is locally. After saving, `hastus_exports`, `hastus_services`, and `hastus_service_dates` rows should be created and S3 should contain the uploaded file.

#### Reviewer Checklist

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -81,7 +81,8 @@ config :arrow,
   gtfs_archive_storage_enabled?: true,
   gtfs_archive_storage_prefix_env: "dev/local/",
   hastus_export_storage_enabled?: true,
-  hastus_export_storage_prefix_env: "dev/local/"
+  hastus_export_storage_prefix_env: "dev/local/",
+  use_username_prefix?: true
 
 # Do not include metadata nor timestamps in development logs
 config :logger, :console, format: "[$level] $message\n"
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -79,7 +79,9 @@ config :arrow,
   shape_storage_enabled?: true,
   shape_storage_prefix_env: "dev/local/",
   gtfs_archive_storage_enabled?: true,
-  gtfs_archive_storage_prefix_env: "dev/local/"
+  gtfs_archive_storage_prefix_env: "dev/local/",
+  hastus_export_storage_enabled?: true,
+  hastus_export_storage_prefix_env: "dev/local/"
 
 # Do not include metadata nor timestamps in development logs
 config :logger, :console, format: "[$level] $message\n"
--- a/lib/arrow/hastus/export_upload.ex
+++ b/lib/arrow/hastus/export_upload.ex
@@ -68,11 +68,7 @@ defmodule Arrow.Hastus.ExportUpload do
     s3_bucket = Application.fetch_env!(:arrow, :hastus_export_storage_bucket)
     path = get_upload_path(filename)
 
-    upload_op =
-      file_data
-      |> List.wrap()
-      |> Stream.map(&IO.iodata_to_binary/1)
-      |> ExAws.S3.upload(s3_bucket, path, content_type: "application/zip")
+    upload_op = ExAws.S3.put_object(s3_bucket, path, file_data, content_type: "application/zip")
 
     {mod, fun} = Application.fetch_env!(:arrow, :hastus_export_storage_request_fn)
 
@@ -86,7 +82,13 @@ defmodule Arrow.Hastus.ExportUpload do
     prefix_env = Application.get_env(:arrow, :hastus_export_storage_prefix_env)
     s3_prefix = Application.fetch_env!(:arrow, :hastus_export_storage_prefix)
 
-    [prefix_env, s3_prefix, filename]
+    usename_prefix =
+      if Application.fetch_env!(:arrow, :use_username_prefix?) do
+        {username, _} = System.cmd("whoami", [])
+        String.trim(username)
+      end
+
+    [prefix_env, usename_prefix, s3_prefix, filename]
     |> Enum.reject(&is_nil/1)
     |> Path.join()
   end
--- a/lib/arrow/hastus/export_upload.ex
+++ b/lib/arrow/hastus/export_upload.ex
@@ -163,14 +163,14 @@ defmodule Arrow.Hastus.ExportUpload do
     do: Regex.run(~r/^(\d{4})(\d{2})(\d{2})/, date_string, capture: :all_but_first)
 
   defp add_or_update_list([], new_service_id, new_date),
-    do: [%{name: new_service_id, service_dates: [new_date]}]
+    do: [%{name: new_service_id, service_dates: [new_date], import?: true}]
 
   defp add_or_update_list(
          [h = %{service_id: service_id, service_dates: existing_dates} | t],
          service_id,
          new_date
        ),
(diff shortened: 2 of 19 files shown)


config/dev.exs
```diff
@@ -79,7 +79,9 @@ config :arrow,
   shape_storage_enabled?: true,
   shape_storage_prefix_env: "dev/local/",
   gtfs_archive_storage_enabled?: true,
-  gtfs_archive_storage_prefix_env: "dev/local/"
+  gtfs_archive_storage_prefix_env: "dev/local/",
```
Whoops, line 82: ❓ With shapes we've ran into issues where one user has uploaded a shape of the same name another has already used. How hard would it be to instead make this prefix be based off the users local username on their local machine (e.x. I get whoops)

cmaddox5: Shouldn't be difficult in deployed environments, but the username would come from Keycloak which would be bypassed locally. This leads to all of us likely sharing the same usernames (`fake_uid` I think). Do you have any problems with that in regards to the `dev/local` path?

lib/arrow/hastus/export_upload.ex
```diff
@@ -56,6 +54,43 @@ defmodule Arrow.Hastus.ExportUpload do
...
+    end
+  end
+
+  defp do_upload(file_data, filename) do
+    s3_bucket = Application.fetch_env!(:arrow, :hastus_export_storage_bucket)
+    path = get_upload_path(filename)
+
+    upload_op =
+      file_data
+      |> List.wrap()
```
Whoops, line 73: Just my own curiosity, why is it necessary to wrap this in a list first?

lib/arrow/hastus/service_date.ex
```diff
@@ -12,16 +12,37 @@ defmodule Arrow.Hastus.ServiceDate do
...
+  @spec validate_start_date_before_end_date(Ecto.Changeset.t(t())) :: Ecto.Changeset.t(t())
+  defp validate_start_date_before_end_date(changeset) do
+    start_date = get_field(changeset, :start_date)
+    end_date = get_field(changeset, :end_date)
+
+    cond do
+      is_nil(start_date) or is_nil(end_date) ->
+        changeset
+
+      Date.compare(start_date, end_date) not in [:lt, :eq] ->
```
Whoops, line 41: I think `:eq` is actually OK here. I may have a service that I only want activated on 2025-03-30.

lib/arrow_web/components/hastus_export_section.ex
```diff
@@ -4,15 +4,29 @@ defmodule ArrowWeb.HastusExportSection do
...
   attr :user_id, :string
+  attr :uploaded_file_name, :string
+  attr :uploaded_file_data, :any
+  attr :disruption, DisruptionV2, required: true
+
+  @line_icon_names %{
+    "line-Blue" => :blue_line,
+    "line-Green" => :green_line,
+    "line-Orange" => :orange_line,
+    "line-Red" => :red_line
```
Whoops, line 28: Mattapan?

config/dev.exs
Whoops: Basically I'm thinking each dev gets their own prefix. So when I run Arrow it goes to `/dev/local/whoops` and you get `/dev/local/cdmaddox`, that way we don't conflict if we try to upload the same export when testing locally.

priv/repo/migrations/20250312170355_create_hastus_export_tables.exs
```diff
@@ -0,0 +1,35 @@
+defmodule Arrow.Repo.Migrations.CreateHastusExportTables do
+  use Ecto.Migration
+
+  def change do
+    create table(:hastus_exports) do
+      add :s3_path, :string
+      add :line_id, references(:gtfs_lines, type: :varchar, on_delete: :nothing)
```
Whoops, line 7: I'd test a gtfs import to make sure this foreign key doesn't cause an issue. If it does I think @jzimbel-mbta has a solution.

lib/arrow/hastus/export_upload.ex
cmaddox5: I think this is just a quirk of `ExAws.S3.upload/4` specifically. It uploads parts of a stream concurrently so it expects an `Enum` for the source param. To get around that, it's wrapped in a list, added to a `Stream`, and uploaded.

lib/arrow/hastus/service_date.ex
cmaddox5: I think my conditional could be improved. That was what I was going for, so it would be easier to just check `== :gt` to make it more readable.

lib/arrow_web/components/hastus_export_section.ex
cmaddox5: Woops, thank you.

lib/arrow/hastus/export_upload.ex
Whoops: Looking at the documentation for [upload/4](https://hexdocs.pm/ex_aws_s3/ExAws.S3.html#upload/4) vs [put_object/4](https://hexdocs.pm/ex_aws_s3/ExAws.S3.html#put_object/4) would `put_object` make more sense here? It looks like `upload` is meant for streaming, but since the whole thing is getting rendered in memory anyway, I don't think there's any advantage there.

config/dev.exs
cmaddox5: That's an interesting idea. It looks like RTR does something [similar](https://github.com/mbta/rtr/blob/204a4e4e65237fa23fc970eebebe761986fb4e59/lib/rtr/s3/uploader.ex#L146) already so I can borrow from that.

lib/arrow/hastus/export_upload.ex
cmaddox5: That's a good point, and an easy change. Only issue is I get a 403 when trying to upload to `dev/local`. Will look at our IAM policies and see what's going on.

cmaddox5: Actually, nevermind! I think I just had a typo that made the request invalid. Was able to upload just fine. We'll see if that stays true with the username paths.

priv/repo/migrations/20250312170355_create_hastus_export_tables.exs
cmaddox5: Just tried it out and all is good. All of the removing/re-add [foreign key logic](https://github.com/mbta/arrow/pull/1030) was written in a way that includes all new constraints as they are added. (Nice work @jzimbel-mbta!)

lib/arrow_web/components/hastus_export_section.ex
```diff
@@ -447,4 +471,7 @@ defmodule ArrowWeb.HastusExportSection do
       "*The selected dates are not #{service_part}. Are you sure?"
     end
   end
+
+  defp error_to_string(:too_large), do: "File is too large. Maximum size is 8MB"
```
Whoops, line 475: Maybe we should have a catch-all for errors we don't recognize?

Whoops (approved):

PR #1152, opened 2025-03-19
feat: Reject replacement service sheets with missing runtime rows 

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹🐛 Arrow accepts sheets where the first/last trip times are outside the headway table](https://app.asana.com/0/584764604969369/1209434092170016/f)

#### Reviewer Checklist


lib/arrow/disruptions/replacement_service_upload.ex
```diff
@@ -158,22 +158,79 @@ defmodule Arrow.Disruptions.ReplacementServiceUpload do
...
+
+        end_hour =
+          if String.slice(end_time, 2..-1//1) == ":00" do
+            time_str_to_hour(end_time) - 1
+          else
+            time_str_to_hour(end_time)
+          end
+
+        start_hour..end_hour//1
+      end)
```
jzimbel-mbta, line 206: [example.xlsx](https://github.com/mbta/arrow/blob/jz-warn-on-missing-headway-table-rows/test/support/fixtures/xlsx/disruption_v2_live/example.xlsx) has a runtime row that covers 24:00 - 26:00, so I abandoned the assumption that each row must cover only 1 hour, which made this code a bit more complicated. (basically, `Enum.map` turned into `Enum.flat_map`.)

lib/arrow/disruptions/replacement_service_upload.ex
```diff
@@ -221,23 +278,42 @@ defmodule Arrow.Disruptions.ReplacementServiceUpload do
...
+
+        {:error, ["Duplicate row(s) for #{Enum.join(dups, " and ")} trip times"]}
+
+      not (first? and last?) ->
+        values = [{first?, "First"}, {last?, "Last"}] |> Enum.reject(&elem(&1, 0))
+
+        {:error, ["Missing row for #{values |> Enum.map_join(" and ", &elem(&1, 1))} trip times"]}
 
-      {:error, ["Missing row for #{values |> Enum.map_join(" and ", &elem(&1, 1))} trip times"]}
+      :else ->
```
cmaddox5, line 315: I did not know you could do this. Thought you always needed default to be `true`. Interesting.

cmaddox5 (approved):
Nice, looks good!

lib/arrow/disruptions/replacement_service_upload.ex
jzimbel-mbta: Yup you can use any truthy value as a catch-all `cond`ition! So any atom besides `nil` or `false` would work.

PR #1179, opened 2025-03-31
fix: Last trip between start & end of last headway

cmaddox5:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹🐛 Timetable view internal server error](https://app.asana.com/0/584764604969369/1209814602688449/f)

In these spreadsheets, the last headway row is for hours 24-26. The current logic tries to grab the relevant headway row by looking at the key (Start time column). The problem is that there is no key for hour 25. This causes a server error when building the timetables. Tweaked the logic a bit so it still tries to do a map lookup, but will look for headway by map value if the key doesn't exist. 

Spreadsheet that produced the error:
[RM1 Ashmont to Mattapan (Weekday Only) Operations Summary.xlsx](https://github.com/user-attachments/files/19535036/RM1.Ashmont.to.Mattapan.Weekday.Only.Operations.Summary.xlsx)

#### Reviewer Checklist


rudiejd:
Deploying this to dev green to poke around

rudiejd:
image

Still getting an error when trying to view the timetable for the problematic replacement service

cmaddox5:
So I believe that particular error is happening because the shuttle has no stops. I know in Slack we talked about the selected shuttle having no impact on the timetables, but a shuttle with no stops might be the one case that it causes an error (not a realistic scenario so not a bug to fix imo). That said, there is another error happening for this in [dev](https://mbta.splunkcloud.com/en-US/app/search/search?sid=1743599604.20540). I also think that is unrelated to this change, but I need to find a workaround so that way you can actually test this.

cmaddox5:
@rudiejd I think I figured out a workaround. There seems to be some shuttle stops that cause issues in deployed environments that I can't reproduce locally. Removing those problematic stops from the shuttle will let you at least view the timetable. Can find it [here](https://arrow-dev.mbtace.com/replacement_services/11/timetable) but feel free to make a new one just in case. I'll need to create a task for us to figure out what the deal is with the stops I deleted (they come from a shuttle definition we have used before and they also are present in the current GTFS feed 🤷 )

rudiejd:
@cmaddox5  I'm still getting a 500 when I view that page 🤔  though now it seems to be related to the `selection_recovery` event:  
image

rudiejd:
maybe you shared the wrong link though? I think the `Firas-OrangeLine-Cursed` shuttle doesn't have stops, but I'm able to see a timetable for this shuttle that has stops https://arrow-dev.mbtace.com/replacement_services/12/timetable

cmaddox5:
Ugh, I didn't think about how merging a dependabot PR would kick this branch off dev. Will do a redeploy just in case that fixes it.

rudiejd:
@cmaddox5 Still getting an error for that problematic activation spreadsheet you linked. Here's the disruption on which I tried to activate replacement service with that spreadsheet: https://arrow-dev.mbtace.com/replacement_services/13/timetable

rudiejd:
can confirm that I can see the timetable when there are trips in a `replacement_service` after hour 24. seems good to me (pending a rebase).

image

PR #1192, opened 2025-04-01
fix: handle empty all_calendar.txt in HASTUS import

sloanelybutsurely:
If a HASTUS export had an empty `all_calendar.txt` file the import process would not produce any services because the export parsing relied on iterating over the services as listed in that `all_calendar.txt` file.

To fix this we read the list of unique `service_id`s from the `all_trips.txt` file instead and then "join" that with the values from `all_calendar.txt`. If there is no entry in the `all_calendar.txt` file for that service_id then we still include the service but without any service dates. The user can then enter the service dates in the UI.

best reviewed [hiding whitespace](https://github.com/mbta/arrow/pull/1192/files?diff=unified&w=1)

[🏹🐛 Empty all_calendar.txt leads to weirdness in HASTUS Import](https://app.asana.com/0/584764604969369/1209816734759554/f)

#### Reviewer Checklist


test/arrow/hastus/export_upload_test.exs
```diff
@@ -40,6 +40,18 @@ defmodule Arrow.Hastus.ExportUploadTest do
...
+                   service_dates: [%{start_date: ~D[2025-03-22], end_date: ~D[2025-03-22]}]
+                 },
+                 %{
+                   name: "RTL12025-hmb15017-Sunday-01",
+                   service_dates: [%{start_date: ~D[2025-03-23], end_date: ~D[2025-03-23]}]
+                 },
+                 %{
+                   name: "RTL12025-hmb15mo1-Weekday-01",
+                   service_dates: []
+                 },
```
sloanelybutsurely, line 54: This change in logic surfaced a `service_id` in the `all_trips.txt` test data that is not referenced in the `all_calendar.txt` file. The only way we can show services if the `all_calendar.txt` file is empty is to parse them from `all_trips.txt` so we have to be okay with this.

lib/arrow/hastus/export_upload.ex
```diff
@@ -247,50 +247,60 @@ defmodule Arrow.Hastus.ExportUpload do
...
+           "all_calendar_dates.txt" => calendar_dates,
+           "all_trips.txt" => trips
+         },
          tmp_dir
        ) do
+    service_ids =
+      trips
+      |> Stream.map(& &1["service_id"])
+      |> Stream.reject(&is_nil/1)
+      |> Enum.uniq()
```
rudiejd, line 261: non-blocking: just curious, why switch from `Stream` to `Enum` and materialize the enumerable at this step?

test/arrow/hastus/export_upload_test.exs
```diff
@@ -40,6 +40,18 @@ defmodule Arrow.Hastus.ExportUploadTest do
       assert {:ok,
               {:ok,
```
rudiejd, line 41: could you add a test simulating an empty `all_calendar.txt` here?

sloanelybutsurely:
@rudiejd test added. ptal!

lib/arrow/hastus/export_upload.ex
sloanelybutsurely: There isn't much point to keeping things as `Stream` when using `*.uniq/1` because the process of deduplicating the stream (or enum) requires maintaining a record of all previously seen values.

From the [`Stream.uniq/1`](https://hexdocs.pm/elixir/Stream.html#uniq/1) docs:

> Keep in mind that, in order to know if an element is unique or not, this function needs to store all unique values emitted by the stream. Therefore, if the stream is infinite, the number of elements stored will grow infinitely, never being garbage-collected.

rudiejd: makes sense to me. I was thinking that if the cardinality of unique elements is expected to be less than the cardinality of duplicate, then holding off on materializing the enum could make sense

rudiejd (approved):

PR #1195, opened 2025-04-02
chore: Pin asdf to 0.15.0 to avoid notice about 0.16 Go rewrite

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** ad hoc
(analogue to [this ticket for gtfs_creator](https://app.asana.com/0/584764604969369/1209642274390141/f))

This prevents asdf from printing notices about the v0.16 Go rewrite every time it runs (i.e., between every single shell command issued during CI)

#### Reviewer Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
       # only run `asdf install` if we didn't hit the cache
       - uses: asdf-vm/actions/install@v1
         if: steps.asdf-cache.outputs.cache-hit != 'true'
+        with:
+          asdf_branch: v0.15.0
 
   build:
     name: Build Elixir


.github/workflows/ci.yml
```diff
@@ -23,6 +23,8 @@ jobs:
       # only run `asdf install` if we didn't hit the cache
       - uses: asdf-vm/actions/install@v1
         if: steps.asdf-cache.outputs.cache-hit != 'true'
+        with:
+          asdf_branch: v0.15.0
 
   build:
     name: Build Elixir
```
rudiejd, line 30: I think you would also have to pin asdf for this job

rudiejd:
is it worth just following the migration guide here so we can run the latest and greatest `asdf`? what is blocking us from migrating to 0.16?

cmaddox5:
> is it worth just following the migration guide here so we can run the latest and greatest `asdf`? what is blocking us from migrating to 0.16?

Because this is a 3rd party action, we don't have control of the default version used. There is apparently a recent fix that has been [merged](https://github.com/asdf-vm/actions/issues/587#issuecomment-2723408197), but there doesn't seem to be a release that contains the change. We could use the commit hash instead of the major version (we're a couple behind anyway). Then we wouldn't need to pin.

jzimbel-mbta:
Whoops, I lost track of this PR!

But it looks like we're going to try and switch to `mise` instead of pinning the `asdf` version (first in gtfs_creator, then probably here and other projects)

So I'll close this for now. If `mise` ends up not working out for some reason, I'll reopen.

PR #1210, opened 2025-04-07
polish: Show a warning if limit dates are in the past

cmaddox5:
#### Summary of changes
**Asana Ticket:** [[extra] 🏹💅🏻🇵🇱 Add non-blocking warning for limits created in the past](https://app.asana.com/0/584764604969369/1209715072281159/f)

When a user attempts to create a limit entirely in the past, we want to warn them. 

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/components/limit_section.ex
+++ b/lib/arrow_web/components/limit_section.ex
@@ -151,6 +151,12 @@ defmodule ArrowWeb.LimitSection do
           <div class="row mb-3">
             <.input class="col-lg-3" field={@limit_form[:start_date]} type="date" label="start date" />
             <.input class="col-lg-3" field={@limit_form[:end_date]} type="date" label="end date" />
+            <div class="col text-sm text-danger align-self-center">
+              {get_limit_date_range_warning(
+                input_value(@limit_form, :start_date),
+                input_value(@limit_form, :end_date)
+              )}
+            </div>
           </div>
           <div class="container justify-content-around mb-3">
             <.inputs_for :let={f_day_of_week} field={@limit_form[:limit_day_of_weeks]}>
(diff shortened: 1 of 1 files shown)


lib/arrow_web/components/limit_section.ex
```diff
@@ -359,6 +365,21 @@ defmodule ArrowWeb.LimitSection do
     Map.get(icon_paths, kind)
   end
 
+  defp get_limit_date_range_warning(start_date, end_date)
+       when start_date in [nil, ""] or end_date in ["", nil] do
```
jzimbel-mbta, line 369: I think a limit that's entirely in the past could still have a nil `start_date`, for example: `%{start_date: nil, end_date: yesterday}`

(In what cases do these variables get empty string values? Not sure if that has a different meaning from nil here)

lib/arrow_web/components/limit_section.ex
```diff
@@ -359,6 +365,21 @@ defmodule ArrowWeb.LimitSection do
...
+    ""
+  end
+
+  defp get_limit_date_range_warning(start_date, end_date) do
+    today = Date.utc_today()
```
jzimbel-mbta, line 374: Are `start_date` and `end_date` in UTC timezone?

If they're local you will want to get `today` in local time as well. Otherwise, this could produce the wrong result if Arrow is used after ~6pm Eastern.
```suggestion
    today =
      DateTime.utc_now()
      |> DateTime.shift_zone!("America/New_York")
      |> DateTime.to_date()
```

jzimbel-mbta (commented):
Couple of small things but looks good overall!

lib/arrow_web/components/limit_section.ex
cmaddox5: The guard against empty string is me being overly cautious due to [this](https://mbta.slack.com/archives/C06R4UC5A0Y/p1742487227604269) convo.

You make a good point though. We only really care about the end date. If end date is in the past, start date is either invalid or also in the past. I should be able to simplify this.

lib/arrow_web/components/limit_section.ex
cmaddox5: Good call, will need to shift.

jzimbel-mbta (approved):
👍

PR #1212, opened 2025-04-07
ci: save wallaby screenshots after integration test failures

sloanelybutsurely:
Uses actions/upload-artifact to make the screenshots captured by wallaby for failed integration tests available for download


sloanelybutsurely:
> Can we document how to get the screenshots?

Where would you like this documented, @dks-mbta? The link to download the screenshots is shown in the output of the upload-artifact step.

PR #1223, opened 2025-04-10
revert: HASTUS limit derivations

cmaddox5:
#### Summary of changes
**Asana Ticket:** ad-hoc

Pulling this out since it isn't 100% right for all lines. The solution would cause parsing to timeout so we will need to consider moving the logic to a background process before adding back.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/components/core_components.ex
+++ b/lib/arrow_web/components/core_components.ex
@@ -284,7 +284,7 @@ defmodule ArrowWeb.CoreComponents do
   attr :type, :string,
     default: "text",
     values: ~w(checkbox color date datetime-local email file month number password
-               range search select tel text textarea time url week)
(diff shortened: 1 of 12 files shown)


lib/arrow_web/components/core_components.ex
```diff
@@ -284,7 +284,7 @@ defmodule ArrowWeb.CoreComponents do
   attr :type, :string,
     default: "text",
     values: ~w(checkbox color date datetime-local email file month number password
-               range search select tel text textarea time url week hidden)
+               range search select tel text textarea time url week)
```
sloanelybutsurely, line 287: might as well keep this imo

sloanelybutsurely (approved):

PR #1224, opened 2025-04-11
feat: group limits by origin and destination

sloanelybutsurely:
- groups limits by their route, start, and end stops
- updates the limits section to use a grid layout for consistent column placement even with some rows not having all columns (route, start, end)
- includes days of week in the limits section
- falls back to a simple flexbox based layout at small breakpoints to prevent the element from overflowing its container

Capture 2025-04-11 at 11 42 00@2x

https://github.com/user-attachments/assets/0e3d9041-f73d-42ff-97ec-4444e4836ec7

[🏹💅🏻🇵🇱  Group Limits in Create/Edit Disruption Pages][asana]

[asana]: https://app.asana.com/1/15492006741476/project/584764604969369/task/1209887002402273?focus=true

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/components/limit_section.ex
+++ b/lib/arrow_web/components/limit_section.ex
@@ -24,62 +24,95 @@ defmodule ArrowWeb.LimitSection do
     <section id={@id} class="py-4 my-4">
       <h3>Limits</h3>
       <%= if Ecto.assoc_loaded?(@disruption.limits) and Enum.any?(@disruption.limits) do %>
-        <div class="mb-3 border-2 border-dashed border-secondary border-mb-3 p-3">
-          <table class="w-[40rem] sm:w-full">
-            <thead>
-              <tr>
-                <th>route</th>
-                <th>start stop</th>
-                <th>end stop</th>
-                <th>start date</th>
-                <th>end date</th>
-                <th></th>
-              </tr>
-            </thead>
-            <tbody class="relative divide-y divide-zinc-100 border-t border-zinc-200 text-sm leading-6 text-zinc-700">
-              <tr :for={limit <- @disruption.limits}>
-                <td>
-                  <span
-                    class="m-icon m-icon-sm mr-1 align-middle"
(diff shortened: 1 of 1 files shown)


lib/arrow_web/components/limit_section.ex
```diff
@@ -24,62 +24,95 @@ defmodule ArrowWeb.LimitSection do
...
-                </td>
-              </tr>
-            </tbody>
-          </table>
+        <div class={
+          [
+            "flex flex-col",
+            "md:grid gap-y-4 gap-x-2",
+            # buckle up...
+            "grid-cols-[[route]_minmax(60px,auto)_[start]_1.5fr_[to]_minmax(40px,auto)_[end]_1.5fr_[startdate]_auto_[enddate]_0.8fr_[days]_1fr_[actions]_minmax(90px,auto)]"
```
jzimbel-mbta, line 32: I braced myself but still was not in any way prepared for the line that follows. 😨

Is this a bootstrap thing? I'm struggling to find docs on it.
How does it even translate that to basic CSS?

Whatever it is, it strikes fear and awe into my heart.

Do you think it would be worth splitting the string up into parts with comments, for those less familiar with the special syntax here?
Sort of like how you can split a complicated regex over multiple lines with the `x` flag and describe each component of the pattern with a short comment.

jzimbel-mbta: Got it, thank you for the explanation! I think this is more on me for not being as up-to-speed with Tailwind, and `grid-template-columns`, and CSS grid in general, as I should be.

The code can be left as-is, and if someone gets curious about it in the future they can pretty easily track down the PR and this comment thread to learn more.

(aside: I do feel like these advanced utility classes start to overlap with just setting the `style` attribute directly on the element at some point, though...)

jzimbel-mbta (approved):
Good to go, thanks again for walking me through that special Tailwind syntax.

PR #1243, opened 2025-04-30
v1 Arrow API endpoint for active disruptions

cmaddox5:
#### Summary of changes
**Asana Ticket:** ad-hoc

In the current v1 Arrow disruptions endpoint, we return every disruptions whether it is active or not. This creates a lot of extra work for gtfs_creator because it applies all of these disruptions even though they will be truncated out for being outside the current rating. Adding this endpoint for gtfs_creator to use in an upcoming PR. 

#### Reviewer Checklist

### DIFF
--- a/lib/arrow_web/controllers/api/disruption_controller.ex
+++ b/lib/arrow_web/controllers/api/disruption_controller.ex
@@ -1,37 +1,46 @@
 defmodule ArrowWeb.API.DisruptionController do
   use ArrowWeb, :controller
-  alias Arrow.{Disruption, DisruptionRevision, Repo}
   import Ecto.Query
+  alias Arrow.{Disruption, DisruptionRevision, Repo}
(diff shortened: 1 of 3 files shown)


Whoops:
Yeah, recaps are where this would break down. We do a recap post-rating with diversions so they can see how the actual service diverged from the intended service, and this would miss any v1 disruptions. Filtering GTFS creator is an option, or if you wanted to make the cutoff a parameter there could be a GTFS PR to start adding a cuttoff date. You can look [here](https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/arrow/update_service_schedules.py#L9) for an example of inferring the dates the feed is being built for in a recap friendly way.

cmaddox5:
Good call, got that work started here. Will open a gtfs_creator PR shortly to add the params to the API call.

Whoops (approved):

PR #1244, opened 2025-04-30
chore: update save shape page/form

sloanelybutsurely:
- Adds new fields for "From", "To" and "Via" which are then combined to create the shape `name`
- Validates PascalCase for the generated names
- Removes the dump inspected coords from success flash message

[Asana](https://app.asana.com/1/15492006741476/project/584764604969369/task/1209989320086485?focus=true)

### DIFF
--- a/lib/arrow/shuttles/shape.ex
+++ b/lib/arrow/shuttles/shape.ex
@@ -45,6 +45,12 @@ defmodule Arrow.Shuttles.Shape do
       %__MODULE__{}
       |> cast(attrs, [:name])
       |> validate_required([:name])
+      |> validate_format(
+        :name,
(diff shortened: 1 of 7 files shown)


Whoops:
Screenshot 2025-04-30 at 1 25 42 PM

My only complaint is that the shape name gets cut off, limiting the usefulness of displaying it. I'm not sure if we should consider this a blocker. Users should probably be relying on the map instead of the shape name anyway, since Remix sometimes lies. Thoughts @fsaid90, @shantigonzales?

Whoops (approved):

fsaid90:
> Screenshot 2025-04-30 at 1 25 42 PM
> 
> My only complaint is that the shape name gets cut off, limiting the usefulness of displaying it. I'm not sure if we should consider this a blocker. Users should probably be relying on the map instead of the shape name anyway, since Remix sometimes lies. Thoughts @fsaid90, @shantigonzales?

Wouldn't consider this a blocker, but agreed!

PR #1245, opened 2025-05-01
Move suffix from shuttle_routes to shuttles.

cmaddox5:
#### Summary of changes
**Asana Ticket:** ad-hoc

We are adding Local and Express shuttles to Arrow for the first time and found that suffixes aren't being applied to the shuttle routes correctly. This lead me to realize that this field is associated with the wrong model. There should be a single `Suffix` input for each shuttle instead of one for each shuttle direction.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/shuttles/route.ex
+++ b/lib/arrow/shuttles/route.ex
@@ -17,7 +17,6 @@ defmodule Arrow.Shuttles.Route do
     do: @direction_1_desc_values
 
   @type t :: %__MODULE__{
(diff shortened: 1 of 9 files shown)


cmaddox5:
@Whoops Would love to hear your thoughts on this. This page predates me so want to make sure there isn't a piece that I'm missing here.

Whoops (approved):

Whoops (approved):

PR #1248, opened 2025-05-02
fix: handle errors when a shape file is missing

sloanelybutsurely:
Updates some codepaths to prevent 500 errors on pages that reference a broken shape.

Viewing a broken shape will flag that the shape is broken. Viewing a shuttle that references a broken shape will simply not have any shapes to render (the failure is silent).

[Asana task](https://app.asana.com/1/15492006741476/project/584764604969369/task/1210046524302364?focus=true)

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -55,28 +55,43 @@
       }
     },
     "../deps/live_select": {
-      "version": "1.5.2",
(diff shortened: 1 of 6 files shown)


jzimbel-mbta (approved):
Good to go 👍

PR #1249, opened 2025-05-05
feat: add download of list of trips/services with missing shapes

sloanelybutsurely:
Updates the edit hastus export form to special case when a hastus export has trips with invalid or missing shapes.

Also adds a new `send_download` function in `CoreComponents` which uses a new client-side event handler to send data over the existing LiveSocket connection and trigger a client-side download of that data. This makes sending one-off files from a LiveView much easier than the alternative of creating a separate controller to send the file.

[Asana task](https://app.asana.com/1/15492006741476/project/584764604969369/task/1210125477521799?focus=true)

### DIFF
--- a/assets/src/app.tsx
+++ b/assets/src/app.tsx
@@ -23,6 +23,37 @@ declare global {
   }
 }
 
+interface DownloadFileEventDetails {
+  base64?: boolean
+  filename: string
+  content_type: string
+  contents: string
+}
+
+window.addEventListener("phx:download-file", (event: unknown) => {
+  if (event == null || typeof event != "object" || !("detail" in event)) {
+    throw new Error(`download-file event missing 'detail'`)
+  }
+
+  const {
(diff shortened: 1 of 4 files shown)


rudiejd:
after a lot of fiddling around with `zip` and `unzip`, I was able to successfully test on dev:
image

However, I made the shapes invalid for 5 trips and it's only showing a trip with no shape and ignoring all of my trips with invalid shape `abcd`
image

sloanelybutsurely:
> Should this identify trips with invalid shapes, or only those with no shape?

@rudiejd ah i think this was actually an omission of the original validation logic. i've updated this to also check that the shapes actually exist. can you share the `zip` you created so i can test it?

sloanelybutsurely:
@rudiejd ptal!

rudiejd:
looks great!
image

rudiejd (approved):

PR #1265, opened 2025-06-06
feat: add an Oban worker to sync stops and shapes from prod

Whoops:
#### Summary of changes
**Asana Ticket:** [🏹 Regular sync of Arrow Stops & Shapes from prod to dev](https://app.asana.com/1/15492006741476/project/1204473309455397/task/1210252044451003?focus=true)


lib/arrow_web/try_api_token_auth.ex
```diff
@@ -39,6 +39,10 @@ defmodule ArrowWeb.TryApiTokenAuth do
     ArrowWeb.TryApiTokenAuth.Local
   end
 
+  defp api_login_module_for_token(%Arrow.AuthToken{username: "fake_uid"}) do
```
Whoops, line 42: Fixes an annoyance I've had where API auth with the local fake user tries to use Keycloak. If preferred, I can split this into its own PR.

PR #1290, opened 2025-08-05
feat: Allow shuttle deactivation if not used by an active disruption with a current or upcoming replacement service timeframe

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [🏹🐛 Can't set shuttle definition to inactive when used in a past disruption](https://app.asana.com/1/15492006741476/project/584764604969369/task/1210643570438483?focus=true)

## Summary of new logic

The logic is a bit complicated because we're allowing a child entity to change its parent(s) under certain conditions.

The basic constraints being enforced are:
- If a disruption is _active_, all of the shuttles it uses must also be _active_.
- The converse of the previous constraint: if a shuttle is _inactive_, then all of the disruptions using it must also be _inactive_.
- Lastly, a shuttle deactivation can cause a related disruption to deactivate only if that disruption's limits, replacement services, and HASTUS services all have timeframes only in the past. This prevents users from accidentally turning off a disruption that's currently taking effect, or will take effect in the future.

In terms of changeset validations and DB operations on save, that looks like the following.

_When a shuttle is deactivated:_
- Any associated active disruptions that ended in the past--using the same logic that computes the "End Date" column's values on the homepage--are themselves disabled along with the shuttle.
- If any associated active disruption ends today or in the future, the shuttle deactivation fails.

_When a disruption is activated:_
- If any non-active shuttles are associated with the disruption, which can now happen since shuttles may be deactivated while still associated with disruptions under certain conditions, then the disruption activation fails.

#### Reviewer Checklist


lib/arrow/disruptions/disruption_v2.ex
```diff
@@ -80,4 +89,34 @@ defmodule Arrow.Disruptions.DisruptionV2 do
...
+            where: d.id == ^id,
+            where: s.status != :active,
+            distinct: s.id,
+            select: s
+        )
+
+      if non_active_activated_shuttles == [] do
+        changeset
+      else
+        shuttles = Enum.map_join(non_active_activated_shuttles, ", ", &inspect(&1.shuttle_name))
```
Whoops, line 110: nittiest nit: is the call to `inspect` needed here? `shuttle_name` will always be a string.

lib/arrow/shuttles/shuttle.ex
```diff
@@ -12,23 +13,34 @@ defmodule Arrow.Shuttles.Shuttle do
           id: id,
           status: :draft | :active | :inactive,
           shuttle_name: String.t(),
+          suffix: String.t(),
+          routes: [Arrow.Shuttles.Route.t()] | Ecto.Association.NotLoaded.t(),
           disrupted_route_id: String.t(),
-          suffix: String.t()
+          disrupted_route: Arrow.Gtfs.Route.t() | Ecto.Association.NotLoaded.t()
```
Whoops, line 19: non-blocking nit: I'm not sure the newly added associations (`routes`, `disrupted_route`) are ever used?

lib/arrow/disruptions/disruption_v2.ex
Whoops: Ah! That makes sense. I was wondering if it was left over from printing the whole struct.

jzimbel-mbta:
> I'm imagining a disruption that has a replacement shuttle during the week but not on the weekend. This would allow someone on Saturday to set the Shuttle to pending, cascading to the disruption, and accidentally cancelling the track closure for that weekend.

@Whoops could you elaborate on this? My code judges replacement services as past vs current/future solely based on their `end_date`s. If a disruption has a shuttle running for 2 weeks on weekdays and it's currently the Saturday in the middle of those two weeks, then the disruption must have a replacement service whose end date is still in the future. (and therefore would block deactivation of any associated shuttle)

```
legend: `|*****|` = shuttle run timeframe
        `^` = today

week 1: Mo Tu We Th Fr Sa Su
        |************| ^

week 2: Mo Tu We Th Fr Sa Su
        |************|
```

lib/arrow/disruptions/disruption_v2.ex
jzimbel-mbta: Removed the quotes! I think it still makes sense to use them for disruption titles in the other error message though.

lib/arrow/shuttles.ex
```diff
@@ -360,25 +364,84 @@ defmodule Arrow.Shuttles do
...
-    case updated_shuttle do
-      {:ok, shuttle} -> {:ok, shuttle |> Repo.preload(@preloads) |> populate_display_stop_ids()}
-      err -> err
+  @spec deactivate_past_disruptions(Shuttle.t()) :: {:ok, [DisruptionV2.t()]}
+  defp deactivate_past_disruptions(%Shuttle{id: shuttle_id, status: status})
+       when status != :active do
+    # Since this code will be reached only if the shuttle changeset passed validations,
+    # we know that any active disruptions associated with this shuttle do not have
+    # any current or upcoming replacement service timeframes, and can be safely deactivated.
+    {_, deactivated_disruptions} =
```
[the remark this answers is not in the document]
Whoops: That would be the way to protect against this. Looking at the documentation for [Multi](https://hexdocs.pm/ecto/Ecto.Multi.html#module-changesets), it specifies that changesets are checked before starting the transaction, so I don't think there's a way to get around it using Multi. I think you'd probably have to switch to `Repo.transact`.

Given how hard it would be to trigger and the actual expected usage patterns, I probably wouldn't put too much effort into this.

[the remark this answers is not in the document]
jzimbel-mbta: I actually just made that code change! Previously I had it creating the changeset eagerly, just before the transaction began. Now it should be a properly atomic operation.

jzimbel-mbta:
Pushed a fix--now all of the timeframe-having things that hang off of a disruption are checked.

There was indeed an existing `Disruptions.start_end_dates/1` function that I could use, although I needed to tweak it since it wasn't taking HASTUS service dates into account.

PR #1392, opened 2026-06-03
feat: In GTFS feed validation, only check consistency against disruption data

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [🏹 Prevent feed validation from blocking most application functionality](https://app.asana.com/1/15492006741476/project/1204137169527258/task/1215251730988571?focus=true)

The meat of the change:
- The validation transaction now checks _only_ the "external" FK constraints on the feed, those originating in Arrow disruption data tables. For example, `shuttle_route_stops.gtfs_stop_id` references `gtfs_stops.id`.
  - Feed data is imported only from the GTFS tables referenced by said "external" FKs.
  - While the transaction still locks certain application-critical tables, it runs much, much faster now so it's not really a problem.

In less abstract terms, this means validation only looks at lines.txt, routes.txt, and stops.txt now.

It completes in half a second locally now, should be about as fast in prod as well.

#### Reviewer Checklist

### DIFF
--- a/lib/arrow/gtfs.ex
+++ b/lib/arrow/gtfs.ex
@@ -47,12 +47,14 @@ defmodule Arrow.Gtfs do
   end
 
   defp import_transaction(unzip) do
+    schemas = importable_schemas()
+
     transaction = fn ->
       external_fkeys = get_external_fkeys()
       drop_external_fkeys(external_fkeys)
 
-      truncate_all()
-      import_all(unzip)
+      truncate(schemas)
+      import_feed(unzip, schemas)
 
       add_external_fkeys(external_fkeys)
     end
@@ -67,8 +69,7 @@ defmodule Arrow.Gtfs do
   end
 
   @doc """
-  Validates a GTFS feed for internal consistency, and consistency with Arrow
-  disruption data that depends on it.
+  Validates a GTFS feed for relational consistency with Arrow's disruption data.
 
   `job` is the Oban job running this validation.
   """
@@ -90,17 +91,21 @@ defmodule Arrow.Gtfs do
   end
 
   defp validate_transaction(unzip) do
+    schemas = validation_schemas()
+
     transaction = fn ->
       external_fkeys = get_external_fkeys()
       drop_external_fkeys(external_fkeys)
+      drop_internal_fkeys()
 
-      truncate_all()
-      import_all(unzip)
+      truncate(schemas)
+      import_feed(unzip, schemas)
 
+      # Only re-add external FKs since we're not concerned with validating the internal consistency of the feed.
       add_external_fkeys(external_fkeys)
 
       # Set any deferred constraints to run now, instead of on transaction commit,
-      # since we don't actually commit the transaction in this case.
+      # since we don't actually commit the transaction for validations.
       _ = Repo.query!("SET CONSTRAINTS ALL IMMEDIATE")
       Repo.rollback(:validation_success)
     end
@@ -114,15 +119,15 @@ defmodule Arrow.Gtfs do
     result
   end
 
-  @spec truncate_all() :: :ok
-  defp truncate_all do
-    tables = Enum.map_join(importable_schemas(), ", ", & &1.__schema__(:source))
+  @spec truncate(list(module)) :: :ok
+  defp truncate(schemas) do
+    tables = Enum.map_join(schemas, ", ", & &1.__schema__(:source))
     _ = Repo.query!("TRUNCATE #{tables}")
     :ok
   end
 
-  defp import_all(unzip) do
-    Enum.each(importable_schemas(), &Importable.import(&1, unzip))
+  defp import_feed(unzip, schemas_to_import) do
+    Enum.each(schemas_to_import, &Importable.import(&1, unzip))
   end
 
   defp validate_required_files(unzip) do
@@ -151,6 +156,7 @@ defmodule Arrow.Gtfs do
   defp validate_version_change(_new_version, _current_version), do: :ok
 
   defp importable_schemas do
+    # All the Ecto schemas that represent GTFS feed tables.
     # Listed in the order in which they should be imported.
     [
       Arrow.Gtfs.FeedInfo,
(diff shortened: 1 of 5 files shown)


lib/arrow/gtfs/import_worker.ex
```diff
@@ -26,10 +26,6 @@ defmodule Arrow.Gtfs.ImportWorker do
     end
   end
 
-  # A sane timeout to avoid buildup of stuck jobs. This is especially important
-  # for cases where our RDS instance runs out of credits--it stops the job from
-  # eating up additional credits as they recharge and keeping the server
```
rudiejd, line 31: is this part of the comment no longer relevant?

rudiejd:
noting that this change, if merged, would get rid of the only way we have to validate that each `checkpoint_id` in `stop_times` shows up in `checkpoints.txt`. We should have a follow-up to add this validation to GTFS creator

lib/arrow/gtfs/import_worker.ex
jzimbel-mbta: It was kind of never relevant :\\

My experience is that when the RDS instance runs out of credits, Arrow uses them up as they recharge just by trying over and over to reconnect to the DB. This timeout value doesn't really help with that in any way.

jzimbel-mbta:
> noting that this change, if merged, would get rid of the only way we have to validate that each `checkpoint_id` in `stop_times` shows up in `checkpoints.txt`. We should have a follow-up to add this validation to GTFS creator

Thanks for calling that one out. Yeah, that's the only "feed-internal" consistency check by Arrow that doesn't duplicate what's already validated by other gtfs_creator CI steps.

I'll create a ticket.

jzimbel-mbta:
[Ticket to implement checkpoints<->stop_times validation in gtfs_creator.](https://app.asana.com/1/15492006741476/project/1204137169527258/task/1215341224326412?focus=true)

lib/arrow/repo/foreign_key_constraint.ex
```diff
@@ -45,6 +45,34 @@ defmodule Arrow.Repo.ForeignKeyConstraint do
...
+      internal_constraints(["bar", "baz"])
+
+  Would produce this:
+
+      [
+        %ForeignKeyConstraint{origin_table: "baz", referenced_table: "bar"}
+      ]
+  """
+  @spec internal_constraints(list(String.t() | atom)) :: list(t())
+  def internal_constraints(tables) when is_list(tables) do
```
rudiejd, line 68: could you make a test for this? the logic seems non-trivial. you could even maybe turn the docstring into a doctest

jzimbel-mbta: 7544b50

jzimbel-mbta:
I'm working on splitting the diff into meaningful commits to ease review. Will also try to add some integration tests for validation and import.

lib/arrow/gtfs.ex
```diff
@@ -94,15 +93,14 @@ defmodule Arrow.Gtfs do
     schemas = validation_schemas()
 
     transaction = fn ->
-      external_fkeys = get_external_fkeys()
-      drop_external_fkeys(external_fkeys)
+      re_add_external_fkeys = drop_external_fkeys()
```
rudiejd: the `drop_external_fkeys/0` fn should be named more descriptively since it now does more than just drop foreign keys

lib/arrow/gtfs.ex
```diff
@@ -139,6 +155,7 @@ defmodule Arrow.Gtfs do
   defp validate_version_change(_new_version, _current_version), do: :ok
 
   defp importable_schemas do
```
rudiejd:
```suggestion
  defp gtfs_schemas do
```

lib/arrow/gtfs.ex
```diff
@@ -160,6 +177,19 @@ defmodule Arrow.Gtfs do
     ]
   end
 
+  # For validation, only the feed tables that are referenced by FKs from
+  # Arrow disruption data tables are imported.
+  defp validation_schemas do
```
rudiejd:
```suggestion
  defp validated_schemas do
```

nit: these are the schemas getting validated, not the schemas for validation. 

This could also be named something like `arrow_schemas`?

test/arrow/repo/foreign_key_constraint_test.exs
```diff
@@ -92,6 +92,27 @@ defmodule Arrow.Repo.ForeignKeyConstraintTest do
     end
   end
 
+  describe "constraints_referencing_internal_tables/1" do
```
rudiejd: this should be changed to `internal_constraints/2` now

jzimbel-mbta:
[Related gtfs_creator PR to add stop_times.checkpoint_id <-> checkpoints.checkpoint_id validation](https://github.com/mbta/gtfs_creator/pull/3372)
Now merged.

lib/arrow/gtfs.ex
jzimbel-mbta: I'm struggling to come up with a clearer name for this. `undoable_drop_external_fkeys`?

Maybe returning a closure to undo the action is too "clever", would you prefer that I just revert this particular change?

lib/arrow/gtfs.ex
jzimbel-mbta: 4efe348

lib/arrow/gtfs.ex
jzimbel-mbta: 7ac5dba

test/arrow/repo/foreign_key_constraint_test.exs
jzimbel-mbta: a78931e

lib/arrow/gtfs.ex
rudiejd: I do think that would be easier to read, but I'll leave it up to you!

rudiejd (approved):

